PR #12, opened 2021-10-03
Media button & connect fail

svdztn:
### handle media button:
SeekBackward is often used, so I remap the button of headset.  
- button NEXT to seekBackward
- button PREVIOUS to seekForward
- double click button PLAY/PAUSE to seekBackward

### do not delete ServerUrl:
In the case of poor network, app fails to connect the server, then the `ServerUrl` will be deleted.  
We have to set `ServerUrl` and login again.

### DIFF
--- a/Server.js
+++ b/Server.js
@@ -64,7 +64,7 @@ class Server extends EventEmitter {
     var res = await this.ping(serverUrl)
 
     if (!res || !res.success) {
-      this.setServerUrl(null)
+      //this.setServerUrl(null)
       return false
     }
     var authRes = await this.authorize(serverUrl, token)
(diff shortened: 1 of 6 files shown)


advplyr:
The new way covers are handled on the server was not implemented yet and it was crashing, so I had to fix it to get this to run.
I tested and it worked great, thanks for helping out.

svdztn:
correction：
> button NEXT to seekForward  
> button PREVIOUS to seekBackward

Glad to help, great project!

advplyr:
This update is live on the google play store

svdztn:
> This update is live on the google play store

Great, thanks for notice.

There is another user friendly function about auto seekBackward. The player should jump backward automatically after pause, the lenth of the jump depends on the total pause time. Longer the pause time, longer the seekBackward. (There's a max and min limit.)  
Clients of all platform should have this feature, I think.

PR #20, opened 2021-10-27
Fix seekback with no play & progress bar data loss

svdztn:
## fix seek back with no play
See #19

## progress bar data loss after restore from app destroy when playing is on
There is a long time without metadata while playing.
So add `duration` to `getStreamSyncData()`
and use it set `this.totalDuration` after restore

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/MyNativeAudio.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/MyNativeAudio.kt
@@ -96,11 +96,14 @@ class MyNativeAudio : Plugin() {
       //if (!isPlaying) currentTime -= playerNotificationService.calcPauseSeekBackTime()
       var id = playerNotificationService.getCurrentAudiobookId()
       Log.d(tag, "Get Current id $id")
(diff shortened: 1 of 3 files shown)


advplyr:
I was testing the new sleep timer and noticed the current time was jumping back 2 seconds every time I opened the app.

I just released `v0.9.15-beta` publishing to the play store now, adding a sleep timer option to sleep at the end of chapter. When the sleep timer finishes and pauses the audiobook, I have it save the current time.

After it paused and saved the current time, I close the app and re-open and it jumps back 2 seconds which is good.
Then without pressing play I close the app again, open it and it jumped back another 2 seconds.

Sorry, I haven't done much testing yet, just want to put this in writing now.

svdztn:
@advplyr 
How about re-open the app one minute after pausing.
Because for less than a minute, seekback time grows continuously and linearly.

svdztn:
Have been looking forward to sleep timer for a long time, great job!!

svdztn:
Have been looking forward to sleep timer for a long time, great job!!

PR #21, opened 2021-10-29
Fix playWhenReady

svdztn:
1. `playWhenReady` is ok now.

2. Found another problem: no “seekback” when operating with media notification.
So move the “seekback” to `onEvents() -> Player.EVENT_IS_PLAYING_CHANGED`.
Although there may be some subtle lag, but it works.


advplyr:
Tested and it worked great, nice job!
This is released as `v0.9.13-beta`, in review on the play store.

PR #67, opened 2021-12-23
Add contribution/environment setup docs

tvories:
Added initial environment setup docs for contributing.  Let me know if there's anything I can add!


advplyr:
Wow, this is excellent! The only thing I can think to add right now would be just be mentioning that the app is built with NuxtJS and Capacitor to run on both ios and android with the same codebase.

I haven't tried developing for android on mac, so I cant say I know anything about that process. I know trying to develop ios on windows is more difficult then it is worth, so I just use a macbook when working on ios.

The expand/collapse really helps to keep it clean, I think the same strategy can be used for the website https://github.com/advplyr/audiobookshelf-web
It might be worthwhile to just add a markdown parser on the website and paste in the readme's.

tvories:
I also really like keeping some of the details behind the expand and collapse, but I don't know if I love the default expand icon.  It doesn't really seem obvious enough.  I am sure there are better ways to make the info look good, but I think this is a good start for now!

I just spun up the android app on my mac and it worked fine.  I have worked on android on a mac before, so I can verify that there isn't anything special about it.

I'll add some info about NuxtJS and Capacitor to the readme.

PR #68, opened 2021-12-23
Clarified Android and iOS setup, added more docs

tvories:
I realized that I didn't write up instructions for the iOS setup, so I clarified that what I added were for Android specifically, and I added some details about NuxtJS and Capacitor.


tvories:
Added docs for iOS setup!

PR #69, opened 2021-12-24
Add issue templates

tvories:
Makes it a little easier for for opening issues or feature requests


advplyr:
Good call, thanks for setting that up

PR #106, opened 2022-03-06
Show "now playing" on lock screen & in control center

rasmuslos:
- Fixed a bug that caused the player to always start at 0:00
- Added `UIApplication.shared.beginReceivingRemoteControlEvents()` to show a "now playing" notification
- Fixed a bug that caused the cover to not load correctly
- Added more remote transport commands


advplyr:
Awesome, thanks for getting that working. I got stuck trying to get the now playing to work.

When testing on iphone 10 if I press skip forward 30s the track timestamps and position get hidden until I press pause/play again.

rasmuslos:
Thanks, it was annoying to get it to work.
I spend way too much time doing so, which resulted in working, but nothing more. I massively improved it and will create a PR

PR #107, opened 2022-03-06
Massively improved the MPNowPlayingInfoCenter

rasmuslos:
- Covers don't get fetched every time the metadata is updated
- Seeking/playing/pausing are handled correctly

If you are wondering, the method returned when the player wasn't playing, which resulted in funny things happening when you seek/play/pause. I also added a `seek` method, which updated the MP-state

### DIFF
--- a/ios/App/App/MyNativeAudio.swift
+++ b/ios/App/App/MyNativeAudio.swift
@@ -3,16 +3,6 @@ import Capacitor
 import MediaPlayer
 import AVKit
 
-extension UIImage {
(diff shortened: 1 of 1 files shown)


advplyr:
I'm not sure if this is what you meant in your first bullet point, but when I skip forward/backward the cover art does not get set.

I'm pretty sure this was working before.

rasmuslos:
I noticed this too, and have a idea how i could fix it, but probably not today

PR #118, opened 2022-03-28
Player rework

rasmuslos:
- Changed app icon
- Support for v2

The sleep timer, direct play and some other minor features are missing, but everything important is implemented
https://github.com/advplyr/audiobookshelf/pull/461 should be merged to or some stuff won't work


advplyr:
Great, I have some uncommitted code also that reverted some of the API changes and attempts at cleaning up Android.

PR #126, opened 2022-04-20
Enable swipe gesture to go back

benonymity:
This isn't a good implementation of the idea, but allowing some form of swipe back gesture would be helpful for user experience to not reach for the back button. Maybe we could make this a dedicated plugin to enable and disable the gesture on item pages and disable elsewhere, as swiping back to previous library pages isn't very intuitive, but that makes this improvement exponentially more complex. I'll tinker with making it a plugin, [this](https://github.com/diiiary/capacitor-plugin-ios-swipe-back) already exists but hasn't been updated in 3 years and calls deprecated methods. At this point I just threw a line in the middle of the AudioPlayer plugin to make it work.


advplyr:
I think we can do this cleanly in javascript also. I'm not sure about that setting you used I didn't look into it yet but it may be the only thing we need to set natively. If it is all we need natively then we don't need to do a plugin for it.

benonymity:
Cool, do you want to merge this now and add js limitations yourself or do you have any pointers how the previous webView could be disabled? Yep, that one (undocumented) setting should be all we need to enable gestures natively, though we might want to find a better place for it than smack in the middle of the Audio Player class haha

advplyr:
Yeah actually I just realized this already works for android. So maybe the reason it doesn't work for iOS is because this setting was set before.

advplyr:
Just tested on iOS and this works. I already pushed the release for 0.9.42 so it will need to go into the next release.

PR #127, opened 2022-04-20
Increase server ping timeout

benonymity:
As mentioned in [a discussion](https://github.com/advplyr/audiobookshelf-app/discussions/60#discussioncomment-2601321) there seems to be an issue occasionally with being unable to ping the server on cellular data, and the probable cause is the 1 second timeout. Does 3 seconds sound like a good number instead?


advplyr:
Yeah 3s sounds good. Thanks

PR #132, opened 2022-04-23
New Data Classes

benonymity:
This copies most of the data classes present in the android app to the iOS side, but I can't figure out how to toggle the MediaType struct in libraryItem based on mediaType (line 28 of DataClasses.swift) since there is no JSON parser built-in that can differentiate and I'm not experienced enough in Swift to know some magic solution that doesn't disrupt the inheritability. Any ideas on how to go about implementing that? 
My goal was originally to fix the issue where there is no cover displayed in the native player, and a simple fix works for that by fetching again the cover for the given ID each time. But I figured this needed to be implemented at some point especially with an eventual goal of offline downloads so I took the long route--once we figure out the MediaType switching it should all work. Input appreciated!


benonymity:
Finally got this to work. Might need to be slightly reworked when local downloads are implemented, but should be a decent foundation. This also fixes the blank cover in the native player, and works for both podcasts and books. The solution turns out to have been combining the two classes and making the different values optionals, and then allowing nil values in the native audio player. Should be ready to merge!

advplyr:
Nice work!

PR #134, opened 2022-04-23
Add mark as read button

### LINKED ISSUE #61: REQUEST:  Mark Unread
A way to mark a book as unread would be helpful for those accidental click plays.

benonymity:
Simple addition of button to mark book as read, fixes #61. Though with the download button that area might start to get cramped on smaller devices.


advplyr:
There is additional steps to take here for downloaded items.

There are several cases:
1. Local library item not connected to a server item: Call native update media progress only
2. Local library item connected to a server item: Call native update to media progress, IF currently connected to the server of this library item then also make the server request.
3. Library item on server: Make server request

benonymity:
I unfortunately don't have access to an Android device to test the local progress updates for the read button--does it look like it works? It does in theory but there might be some issue I can't see in practice. I can confirm it does definitely sync to the server at the very least and appears properly in the iOS app. If so should be ready to merge!

advplyr:
Did you run this in the browser using npm run dev? Or were you looking at it on iphone?

benonymity:
No errors with `npm run dev`, and also works when built on iOS. I just downloaded Android Studio, but I can't emulate a device as my CPU doesn't support virtualization, so can't confirm it works there though I imagine so.

advplyr:
Does it look right for you? I did an npm run dev and the buttons were different sizes and overflowing depending on the device.

benonymity:
Seems I tested it on the one device where the natural sizing worked haha, just added proper classes for CSS. 
Read_Btn Should display properly now.

pages/item/_id.vue
```diff
@@ -351,6 +354,52 @@ export default {
         console.log('New local library item', item.id)
         this.$set(this.libraryItem, 'localLibraryItem', item)
       }
+    },
+    async toggleFinished() {
+      this.isProcessingReadUpdate = true
+      if (this.isLocal || this.localEpisode) {
```
advplyr, line 360: This button should not show for podcasts.

advplyr:
Something about those buttons not being the same height is throwing me off. Maybe we make the play button skinnier or meet in the middle somewhere.

benonymity:
Removed podcast reference and made CSS consistent with server UI for the play button, so now it's the same size at the read button. Does this look okay?

shorter_playbtn

pages/item/_id.vue
```diff
@@ -26,17 +26,18 @@
...
-          <ui-btn color="success" :disabled="isPlaying" class="flex items-center justify-center flex-grow mr-2" :padding-x="4" @click="playClick">
+          <ui-btn color="success" :disabled="isPlaying" small class="flex flex-grow justify-center items-center h-9 mr-2" :padding-x="4" @click="playClick">
             <span v-show="!isPlaying" class="material-icons">play_arrow</span>
             <span class="px-1 text-sm">{{ isPlaying ? 'Playing' : 'Play Local' }}</span>
           </ui-btn>
           <ui-btn v-if="showRead && isConnected" color="info" class="flex items-center justify-center mr-2" :class="showPlay ? '' : 'flex-grow'" :padding-x="2" @click="readBook">
             <span class="material-icons">auto_stories</span>
             <span v-if="!showPlay" class="px-2 text-base">Read {{ ebookFormat }}</span>
           </ui-btn>
+          <ui-read-icon-btn :disabled="isProcessingReadUpdate" :is-read="userIsFinished" class="mx-0.5 flex items-center justify-center" @click="toggleFinished" />
```
advplyr, line 37: We can't show this for podcasts

benonymity: I didn't think this page was accessible for a podcast item, they have their own page with independent buttons to mark as read correct? But if this does ever get used for a podcast I added `v-if="!isPodcast"`. Was that what you meant?

advplyr:
image

benonymity:
Okay, this *should* work... I resized the read button slightly to fit alongside the old play button. The resizing doesn't impact the podcast view. I also disabled the download button for podcasts on iOS because that just threw errors. Sorry for all the mistakes and back and forth!

advplyr:
No problem! Are you running this in your browser to test? I'm getting the same layout issues.

image

benonymity:
Woah! That's so much easier, I was unaware of that dev tool to change display size! I've been compiling the app in XCode and sending it to my iPhone each time I made a change to test...hot reloading is incredible! Thanks for showing me the light lol. *Now* it should actually be working, I fixed the padding and tested it at a bunch of different resolutions and it seems to work well. Also fixed the download icon being out of line with the read button icon in the podcast table view.

advplyr:
Is this intentional being on 2 lines? I think we can shorten this to just "Stream" and "Play". The difference in wording there helps to distinguish between when you are playing audio from your server verses playing a locally downloaded audio file.

image

benonymity:
Good idea! Done.

advplyr:
Awesome, great job. I tested on android and it works for local downloads too.

PR #135, opened 2022-04-23
Loading indicator while awaiting library

### LINKED ISSUE #53: [Bug] App always starts with "bookshelf empty" message
Everytime I start the app, it shows the following screen.
Screenshot_20211203-105718_AudioBookshelf

It's only after a few seconds (about 7-10s) that my books load into the app.

benonymity:
Instead of displaying "Bookshelf empty for library ____" while the app is awaiting a response on opening the Home bookshelf, use ui-loading-indicator to indicate data is being fetched. Fixes #53.

### DIFF
--- a/pages/bookshelf/index.vue
+++ b/pages/bookshelf/index.vue
@@ -6,7 +6,7 @@
       </template>
     </div>
 
-    <div v-if="!shelves.length" class="absolute top-0 left-0 w-full h-full flex items-center justify-center">
+    <div v-if="!shelves.length && !loading" class="absolute top-0 left-0 w-full h-full flex items-center justify-center">
       <div>
         <p class="mb-4 text-center text-xl">
           Bookshelf empty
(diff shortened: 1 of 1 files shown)


pages/bookshelf/index.vue
```diff
@@ -27,6 +27,9 @@
         </div>
       </div>
     </div>
+    <div v-if="loading" class="absolute top-0 left-0 w-full h-full flex items-center justify-center">
+      <ui-loading-indicator text="Loading Library...">
```
advplyr, line 31: I just noticed when I ran this that there was an error, it didn't have a closing />
You can run your updates locally first by just doing `npm run dev` and accessing from the browser.

benonymity: Whoops! I missed that! Huh, I built that branch with `npm run generate` and `npx cap sync` to run the app on my iOS device, no errors. But I do see it with `npm run dev`, in the future I'll check that too. Sorry!

Read_Btn

PR #162, opened 2022-05-01
Tweaks for bookshelf list view

benonymity:
Add lines to more easily distinguish between items in the app list view and add duration label.
List View

### DIFF
--- a/components/cards/LazyListBookCard.vue
+++ b/components/cards/LazyListBookCard.vue
@@ -114,6 +114,8 @@ export default {
         } else {
           return this.media.numEpisodes + ' episodes'
         }
+      } else {
+        return null
       }
     },
     placeholderUrl() {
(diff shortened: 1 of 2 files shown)


advplyr:
Nice touch! Podcasts don't have a duration on the media object, so we can hide the duration for podcasts. Maybe we show number of episodes instead.

benonymity:
Good idea! Duration should now be hidden for podcasts and episode numbers shown instead. 

Episode Numbers

components/cards/LazyListBookCard.vue
```diff
@@ -99,9 +101,21 @@ export default {
     mediaType() {
       return this._libraryItem.mediaType
     },
+    duration() {
+      return this.media.duration || null
+    },
     isPodcast() {
       return this.mediaType === 'podcast'
     },
+    episodes() {
```
advplyr, line 110: Computed property should always return some value

benonymity: Whoops! I'll keep that in mind for the future--this is the first time I've used Vue so still learning the ropes.

PR #213, opened 2022-06-01
Added narrator to library item page.

cassieesposito:
This is where it looks best to me. I can see an argument for putting it right below the author by line, but it felt cluttered to me up there, especially on smaller devices or on library items with several narrators. I'll happily move it anywhere if this doesn't seem like the spot.

### DIFF
--- a/pages/item/_id.vue
+++ b/pages/item/_id.vue
@@ -62,6 +62,7 @@
     </div>
 
     <div class="w-full py-4">
+      <p v-if="narrators" class="text-sm text-gray-400">Narrated by {{ narrators }}</p>
       <p class="text-sm">{{ description }}</p>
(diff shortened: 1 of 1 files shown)


advplyr:
image

I think we need to work on this design. I agree it probably doesn't look great at the top either so maybe we need to redesign the entire page.

advplyr:
Here is my attempt at a redesign. I've been meaning to add a "More menu" so we have a place to add more options like editing, adding to collections, etc.
This is all pushed to master already.

image
image

advplyr:
This will be in [v0.9.47-beta](https://github.com/advplyr/audiobookshelf-app/releases/tag/v0.9.47-beta)

PR #272, opened 2022-07-06
iOS downloads

benonymity:
Here is some groundwork I've done so far for iOS downloads, now to get it to talk nice with the frontend and actually play local audio. Lots of work left to be done, and I don't believe my Realm objects are perfect yet especially with lists of optionals, which apparently don't exist in Realm...so I need to find a solution for that. Maybe separating the objects for JSON serialization and local database saving is the solution. Feel free to take what I've done so far to go any other directions with the iOS downloads! I'll try to do a bit of work on it over the next few days, but after that I'm going to be away for a week.


benonymity:
Far better work being done by @ronaldheft in PR #277!

PR #277, opened 2022-07-09
iOS Downloads  / Improved Progress Syncing

### LINKED ISSUE #178: [iOS] Will it be possible to download on the iOS app?
From what I can see there is no option to download an audiobook on iOS. I just switched over from android and am not seeing that option. Perhaps I am missing something or is that feature not implemented?

### LINKED ISSUE #319: iOS Progress Syncing Improvements
This is to track a list of improvements to iOS progress syncing once #178 is complete. The goal is to fix all the blind spot areas that may not accurately be syncing progress

- [ ] Use [iOS background task API](https://developer.apple.com/documentation/backgroundtasks) to sync progress on all playback events, even if the app was removed from memory
- [ ] Refactor the "tick timer" logic for progress tracking to resume when the app is returning from running in the background
- [ ] Support progress syncing entirely in iOS app lifecycle events to lay the groundwork for CarPlay
  - [ ] Opening the app from the background should sync all progress with the server, if online

ronaldheft:
# iOS Downloading / Improved Progress Syncing

Supports iOS downloads and improved progress syncing. Resolve  #178. Fixes #319.

## User Interface

- Adds lifecycle events to the front-end to provide better hooks for the native code
	- Event `abs-ui-ready` when UI is fully mounted
	- `AbsAudioPlayer.onReady()` when the audio player is fully loaded and all settings have been initialized
- Enables the download button for audiobooks on iOS
- Enables the download button for podcasts on iOS
- Disables reordering downloaded tracks on iOS
- Removes the "Folder" location on downloaded files for iOS, as it does not apply
- Removes Scan and Force Re-scan options on iOS
- Send iOS users to the bookshelf view after deleting a file vs. the folders view
- Rephrases the delete confirmation box to make sense for both iOS and Android

## iOS General

- Bumps the minimum iOS version to iOS 14 from iOS 12 to enable use of `async`/`await` Swift functionality
	- [92.7% of iOS users are on iOS 14 or above](https://iosref.com/ios-usage)
	- With the tech-focused audience of ABS, I anticipate that percentage is even higher
	- Supported on all iOS devices from iPhone 6 / iPhone 5s and above
- Adds log statements to iOS background lifecycle hooks, to assist with development and tracking down syncing issues
- Adds helper extensions for working with JSON and serializing to dictionary objects
- Adds decoder logic for JSON where fields may be returned as String or Int/Double
	- This is not an issue for API endpoints on Java with Jackson, as that is transparently handled

## Realm

- Converts the existing DataClasses to Realm objects, to allow for persisting into the database as needed
- Manually implements `Codeable` on Realm objects, as Realm `List` is not `Codeable`
- Refactors object models into a model-per-file structure vs. having all classes in one file
- Adds Realm Extensions like `update()` and `delete()` to make working with Realm objects easier and eliminate the need for DAO pattern, which Realm is not designed around
- Bumped the schema to version `2` from `1` to accommodate the new models required for offline functionality
- Refactors Realm usage to not use a shared Realm instance locked to a specific Thread
	- [This is a bad practice for Realm](https://stackoverflow.com/a/31594548), and the recommended usage is to create a new Realm instance as needed
	- [Realm internally caches instances per thread and handles this logic for us](https://www.mongodb.com/docs/realm-legacy/docs/swift/3.17.1/api/Classes/Realm.html)
	- Addresses multi-threading issues by use `frozen()` objects and `ThreadSafeReferences`

## AbsAudioPlayer

- Adds a new Capacitor method `onReady()` to fire when UI is fully loaded
	- This is the hook to restore a previously active listening session
- Refactores the logic of starting a playback session to find common code paths between online / offline playback
- Persists the `playbackSpeed` in Realm and references it when needed
	- This is to enable proper restoration of the last playback speed when opening the app fresh, as well as support CarPlay restoring in the future, when the JavaScript front-end would not be loaded
- Implements logic to sync local playback sessions to the UI
- Prevents metadata from being sent to the UI when the AudioPlayer is not fully loaded
	- Guards against incorrect progress, playback state, or playback speed from appearing in UI, and in some cases, syncing back the native code

## AbsDatabase

- Implements Capacitor methods present on Android to persist local media progress
- Implements Capacitor methods present on Android to sync local media progress with the server
- Implements local library item methods

## AbsFileSystem

- Marks methods that will never be implemented on iOS as unavailable
- Implements deleting library items / podcast episodes

## ApiClient

- Implements endpoints to sync progress with the server
- Begins the work of using `async`/`await` syntax for some endpoints
- Improves using `Encodable`/`Decodable` objects in API endpoints

## PlayerProgress

- Implements syncing between sessions, `MediaProgress`, and server objects
- Uses the iOS background APIs to ensure sync operations always occur, even if iOS is in the process of removing the app from memory
	- This also lays the groundwork for syncing with CarPlay and improving sync actions when the iOS app is completing running in the background with the front-end unloaded
	

## PlayerHandler

- Ensures bad data is not passed around by verifying the AudioPlayer is initialized before returning data
- Implements syncing progress with the server when the player is paused
- Persists playback sessions to the database for restoration
- Marks old playback sessions as inactive when starting a new session

## Native iOS AudioPlayer

- Loads downloaded iOS assets
	- Will fallback to streaming from server if file is corrupt
- Implements logic for displaying downloaded files in the Now Playing views on iOS
- Address some playback session threading issues, with the playback session persisted in the database
- Fixes a progress syncing issue, where the playback progress could get synced at `currentTime=0` with a race condition when initializing the AudioPlayer

## AbsDownloader

- Handles downloading of library items and tracks
- Sends download progress to the user interface
- Implements downloading in a way that will allow us to leverage iOS background downloading in the future
	- For example, for podcast episodes or letting a slow download run in the background when the app is exited

---

Current task list of remaining items:

- [x] Fix logging out of app
- [x] Fix incorrect progress object sent to local sync endpoint
- [x] Fix download progress never completing
- [x] Cascade deletes and prevent orphaned records in the database
  - [x] Remove Unrealm and merge native Realm branch into current downloads branch
- [x] Test offline artwork display in Now Playing widget
- [x] Test downloading an audiobook without artwork
- [x] Test downloading a podcast without artwork
- [x] Persist the playback session and restore on app start
  - [x] Determine how to start a playback session offline
  - [x] Determine how to sync a playback session to the server if it started offline
  - [x] Will the front-end respond to this through events, or is work required on the front-end?
  - [x] Restore playback session on startup
  - [x] Implement `PlayerNotificationService.checkCurrentSessionProgress` from Android
  - [x] Implement session sync on server connection
  - [x] Fix audio starting to play when restoring session on launch
  - [x] Lifecycle hook for when webapp is started (trigger session restore)


ios/App/App/plugins/AbsAudioPlayer.swift
```diff
@@ -7,11 +7,12 @@
 
 import Foundation
 import Capacitor
+import RealmSwift
 
 @objc(AbsAudioPlayer)
 public class AbsAudioPlayer: CAPPlugin {
     private var initialPlayWhenReady = false
-    private var initialPlaybackRate:Float = 1
```
ronaldheft, line 14: Removed this in favor of persisting the last playback rate in the native DB

advplyr:
This is excellent, thank you for all your effort on this.

PR #299, opened 2022-07-27
Added alternative view setting

samblooms:
This adds a device setting that will act similarly to the 'Alternative Bookshelf View' toggle on the web version. A toggle switch added to the settings page will change the shelf color, shelf background, and text color on the labels to match the alternate style. 

* Changing the label style on the series page ( LazySeriesCard.vue ) is buggy. 
   Using the same 'v-bind:class' and a method to return either 'shinyBlack' or 'altBookshelfLabel' breaks the loading of series.


advplyr:
The alternate bookshelf view in the web app shows the title/author under the LazyBookCard
image

I think we should implement it this way and use the regular bg instead of importing the texture. What to you think?
I'm going to push some commits here

advplyr:
I made everything look like the web app except I kept in your altBookshelfLabel for series and collections pages. I think it looks nice thanks for contributing.

samblooms:
Wow. I totally overlooked that bit with the title/author. It was kind of a spur of the moment "change the look" type deal at first. 
That's definitely a more elegant solution.

PR #308, opened 2022-08-04
Add Statistics

benonymity:
Copies most of the stats from the normal web app, with adjusted spacing for mobile and adds a tab to the hamburger menu for it.


benonymity:
stats
Looks something like this

advplyr:
Great thanks, I made some minor updates.

PR #322, opened 2022-08-18
added manual sleep timer dialog

jakob42:
Not much to say, removed the 1 minute sleep timer preset to make room for a new manual sleep timer entry. It will open a dialog to set the number of minutes for the sleep timer.


advplyr:
I was able to go into negative minutes when testing

jakob42:
🤦 I'll take care of it

Issue #79, opened 2022-01-07
Add skip ahead and chapter control on Android Auto


Lau-ie:
The android app works pretty well on Android auto, but skip 10s ahead and chapter control are disabled in the interface. It's actually pretty annoying if I read the hardcopy and need to skip ahead a couple chapters. Can those controls be made active?

advplyr:
I couldn't find any documentation to implement this.
Overall the Android Auto developer experience has been poor. If you or anyone finds any info on how to do this it would be appreciated, I also use this feature a lot.

Xuereb:
Before I made the comment in discord I tried looking for documentation of implementation, but was disappointed with what little I could find

AwsomeFox:
Added another request for control options https://github.com/advplyr/audiobookshelf-app/issues/333

Also since it was discussed in the pr I think the order of the custom android auto playback options is dependent on the order they're added in code. I can't find any docs to support this but 🤷

advplyr:
Jump forward/backward buttons are added. Chapter control should be working for audiobooks with multiple mp3s but not yet working for single file m4b.
[v0.9.57-beta](https://github.com/advplyr/audiobookshelf-app/releases/tag/v0.9.57-beta)

JvdMaat:
Bumping this up. I just converted my Audible library to m4b so I can listen through Audiobookshelf in my car using Android Auto. It's a slight step down from what my experience with Audible was. (No cover art, no chapters). Also, the audio pauses for other notifications but doesn't do a quick-rewind. So I sometimes end up manually rewinding so I can catch what was said.

stevenffrench:
Good to know. Yes, I'm using almost entirely single file M4Bs with chapters. You might take a look at https://github.com/mattttvaughn/chronicle. The project has been without development for a couple of years, but it works pretty well for detecting/using chapters in M4Bs.

Issue #239, opened 2022-06-12
Notification progress bar does not respect "chapter track" setting


DDriggs00:
### Steps to reproduce

1. Have an audiobook with chapters
2. Play the book, and turn on the "chapter track" setting
3. Exit the app and expand the notification to reveal the progress bar

### Expected behaviour
- The progress bar covers only the current chapter when chapter tracks are enabled

### Actual behaviour
- The progress bar covers the entire book, not just the current chapter

### Environment data

Audiobookshelf Version:

- [X] Android App?
- [ ] iOS App?

#### Android Issue
Android version: Oxygen OS 11.0.7.1.HD65AA

Device model: OnePlus 7T

Stock or customized system: Stock

advplyr:
Are you saying it doesn't work when you close and re-open the app? Or it doesn't work at all?

Are you able to see the chapters in the web app to make sure they are recognized by abs?

Do you have any audiobooks that do work as expected?

DDriggs00:
It doesn't work at all. The in-app progress bar (with both collapsed and full-screen players) works as it should, showing progress for the current chapter. It is only the notification (seen when outside the app) that doesn't respect the setting. The web app shows the chapters. All audiobooks with chapters are affected. [Screenshots](https://imgur.com/a/6lby7R0)

advplyr:
Oh I see what you mean. The notification is getting passed in the duration of the currently playing audio file I believe. Are you using a single audio file with chapters?
I think I have an idea on how this could be fixed. It may also be worthwhile showing the chapter title in place of the book title and the book title in place of the author.

DDriggs00:
Yes, I think all the books I've tested have been single-file m4b with internal chapter marks.

ronaldheft:
I love this idea, and will look to implement this on the iOS side.

advplyr:
I added an explanation for why this happens here https://github.com/advplyr/audiobookshelf-app/issues/729#issuecomment-1575082870 for future reference

advplyr:
This has been added for iOS in [v0.9.71-beta](https://github.com/advplyr/audiobookshelf-app/releases/tag/v0.9.71-beta) so this is Android specific now

TheDogg:
Any updates on this issue? Still not fixed on Android. Would really appreciate it, having the whole file in the notification is kind of annoying.

Thanks

quartertone:
Yes, please implement this on Android!

[...]

PR #328, opened 2022-08-21
Add additional controls to android auto

vangorra:
This PR adds additional media controls to Android Auto (and perhaps Wear and maybe Automotive?) I know little about how Android's MediaBrowserService works and Google's lack of documentation seems motivated to keep things a mystery.

So far, this change includes:
- Addition of jump forward button in Android Auto.
- Addition of jump back button in Android Auto.
- Player, Android notification, and Android Auto all respect the "Jump" times specified in device settings.

Discussion Points:
- Previous button - This button appears on Android Auto and I have no idea how to remove it, so I mapped it's functionality to the media player's previous functionality.
- Next button - Since it seems we have no control over which standard buttons are shown, I experimented with adding a next button as a custom action. This is doable but we'll have to start managing our own prev/next drawables if we want a guarantee of consistent UI.
- Order of custom action buttons on the UI is weird. Some buttons are added before play/pause and some after. I didn't find anything that gives us control of the weight.
- String - A lot of UI strings are defined in the capacitorjs code. Rather than add to strings.xml, we should be pulling from the JS if we can but I'm not sure how.

image

Ref: #79

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/data/DeviceClasses.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/data/DeviceClasses.kt
@@ -37,9 +37,9 @@ data class DeviceSettings(
   }
 
   @get:JsonIgnore
-  val jumpBackwardsTimeMs get() = jumpBackwardsTime * 1000L
+  val jumpBackwardsTimeMs get() = (jumpBackwardsTime ?: default().jumpBackwardsTime) * 1000L
   @get:JsonIgnore
-  val jumpForwardTimeMs get() = jumpForwardTime * 1000L
+  val jumpForwardTimeMs get() = (jumpForwardTime ?: default().jumpBackwardsTime) * 1000L
 }
 
 data class DeviceData(
--- a/android/app/src/main/java/com/audiobookshelf/app/player/MediaSessionCallback.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/player/MediaSessionCallback.kt
@@ -7,10 +7,8 @@ import android.os.Handler
 import android.os.Looper
 import android.os.Message
 import android.support.v4.media.session.MediaSessionCompat
-import android.support.v4.media.session.PlaybackStateCompat
 import android.util.Log
 import android.view.KeyEvent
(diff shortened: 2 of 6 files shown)


advplyr:
Nice thanks, I had just figured out how to add an icon in AA but didn't get around to implementing it.

I'm testing this out and noticing that while the audio is paused it doesn't jump forward/backward. I think this is just a feature of AA since it doesn't look like the code is doing anything to prevent the seek on pause, but curious if you have the same behavior.

vangorra:
I was unable to reproduce what you described by its likely I tested incorrectly.

I applied the following cases with Android Audio, Notification, and in App.
- Start a book.
- Wait for it to get around 40 seconds.
- Pause book.
- Jump forward.
- Play. Confirm it resumes here it should.
- Pause book.
- Jump backward.
- Play. Confirm it resumes here it should.

On another note. I think I understand what gets next/previous buttons to show. I suspect these buttons are added based upon the playback queue. For books with multiple audio files, this isn't an issue. For single files (like m4b), one would expect prev/next to navigate chapters. I' dig a bit further but I feel confident this can be worked out.

All in all, while the documentation for creating apps using MediaBrowserService is terrible, this is a welcomed interface compared to the old way (creating our own media browser, streamer, etc).

advplyr:
It does actually jump forward, I meant to say the timestamp in the UI doesn't change until you press play.

That makes sense that the next/prev button has to do with the track list.

The single file m4b is also problematic for the notification player when we want to show the chapter track and not the full audio track. See https://github.com/advplyr/audiobookshelf-app/issues/239

My first idea for a solution to this is to "trick" the media player by pushing a media item into the queue for each chapter that would be the same audio file. We would probably need to override some Exoplayer functions for this but when loading that track we just jump to the timestamp of the chapter.
That may be too much effort and the alternative would be writing custom next/prev buttons for AA. For the notification player overriding the function sending data to the notification about the current time and duration.

vangorra:
I had the same idea about customizing the media loader and extending exoplayer to handle a virtual queue. I'll see what I can do, it's pretty similar to something I did ages ago before android really had any audio players.

PR #330, opened 2022-08-23
Improve iOS Player Sync Logic

ronaldheft:
This PR refactors the iOS player sync hooks to rely on the native instead `AVPlayer` time observation hooks instead of using timers.

The theory is in low memory situations, iOS is stopping the current tick timer in the background, leading to issues where the player does not have the most accurate `currentTime`.

## Changes

* Persisting a syncing sessions are now done via `addPeriodicTimeObserver()` firing every 0.5 wall-clock seconds
  * The database `currentTime` is updated on every invocation
  * The server receives the latest session and media progress every 10 wall-clock seconds and when the player is paused or stopped
* `timeListening` is now calculated based on `updateAt` times between sync vs. running an incremental counter each second
  * This ensures it's reliable regardless of the interval the value is updated
* Sleep timers are now done via `addBoundaryTimeObserver` scheduled for the time to stop sleep
  * This refactor takes into account playbackSpeed, changes in playbackSpeed, seeks, and locks on a specific chapter time when required, to ensure the scheduled sleep time is always accurate
* Syncing the current progress from the server is still done via a timer, but has been refactored into the Audio Player, and is now aware of pauses using the Now Playing widget or a remote
* When resuming audio with allowed seek back, the last played time is based off the database session, allowing persistence across app restarts
* Fixes an existing Realm memory leak that was made worse by this change
  * Adds lots of error handling logic as part of tracking down this memory leak

### DIFF
--- a/ios/App/Shared/models/local/LocalMediaProgress.swift
+++ b/ios/App/Shared/models/local/LocalMediaProgress.swift
@@ -120,7 +120,7 @@ extension LocalMediaProgress {
     }
     
     func updateIsFinished(_ finished: Bool) {
-        try! Realm().write {
+        try! self.realm?.write {
             if self.isFinished != finished {
                 self.progress = finished ? 1.0 : 0.0
             }
@@ -136,7 +136,7 @@ extension LocalMediaProgress {
     }
     
     func updateFromPlaybackSession(_ playbackSession: PlaybackSession) {
-        try! Realm().write {
+        try! self.realm?.write {
             self.currentTime = playbackSession.currentTime
             self.progress = playbackSession.progress
             self.lastUpdate = Date().timeIntervalSince1970 * 1000
@@ -146,7 +146,7 @@ extension LocalMediaProgress {
     }
     
     func updateFromServerMediaProgress(_ serverMediaProgress: MediaProgress) {
-        try! Realm().write {
+        try! self.realm?.write {
             self.isFinished = serverMediaProgress.isFinished
             self.progress = serverMediaProgress.progress
             self.currentTime = serverMediaProgress.currentTime
--- a/ios/App/Shared/models/local/LocalMediaProgress.swift
+++ b/ios/App/Shared/models/local/LocalMediaProgress.swift
@@ -158,19 +158,24 @@ extension LocalMediaProgress {
     }
     
     static func fetchOrCreateLocalMediaProgress(localMediaProgressId: String?, localLibraryItemId: String?, localEpisodeId: String?) -> LocalMediaProgress? {
-        if let localMediaProgressId = localMediaProgressId {
-            // Check if it existing in the database, if not, we need to create it
-            if let progress = Database.shared.getLocalMediaProgress(localMediaProgressId: localMediaProgressId) {
-                return progress
+        let realm = try! Realm()
+        return try! realm.write { () -> LocalMediaProgress? in
+            if let localMediaProgressId = localMediaProgressId {
+                // Check if it existing in the database, if not, we need to create it
+                if let progress = Database.shared.getLocalMediaProgress(localMediaProgressId: localMediaProgressId) {
+                    return progress
+                }
             }
-        }
-        
-        if let localLibraryItemId = localLibraryItemId {
-            guard let localLibraryItem = Database.shared.getLocalLibraryItem(localLibraryItemId: localLibraryItemId) else { return nil }
-            let episode = localLibraryItem.getPodcastEpisode(episodeId: localEpisodeId)
-            return LocalMediaProgress(localLibraryItem: localLibraryItem, episode: episode)
-        } else {
-            return nil
+            
+            if let localLibraryItemId = localLibraryItemId {
+                guard let localLibraryItem = Database.shared.getLocalLibraryItem(localLibraryItemId: localLibraryItemId) else { return nil }
+                let episode = localLibraryItem.getPodcastEpisode(episodeId: localEpisodeId)
+                let progress = LocalMediaProgress(localLibraryItem: localLibraryItem, episode: episode)
+                realm.add(progress)
+                return progress
+            } else {
+                return nil
+            }
         }
     }
 }
--- a/ios/App/Shared/models/local/LocalMediaProgress.swift
+++ b/ios/App/Shared/models/local/LocalMediaProgress.swift
@@ -119,8 +119,8 @@ extension LocalMediaProgress {
         self.finishedAt = progress.finishedAt
     }
     
-    func updateIsFinished(_ finished: Bool) {
-        try! self.realm?.write {
+    func updateIsFinished(_ finished: Bool) throws {
+        try self.realm?.write {
             if self.isFinished != finished {
                 self.progress = finished ? 1.0 : 0.0
             }
@@ -135,8 +135,8 @@ extension LocalMediaProgress {
         }
     }
     
-    func updateFromPlaybackSession(_ playbackSession: PlaybackSession) {
-        try! self.realm?.write {
+    func updateFromPlaybackSession(_ playbackSession: PlaybackSession) throws {
+        try self.realm?.write {
             self.currentTime = playbackSession.currentTime
             self.progress = playbackSession.progress
             self.lastUpdate = Date().timeIntervalSince1970 * 1000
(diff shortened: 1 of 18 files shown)


ios/App/Shared/models/local/LocalMediaProgress.swift
```diff
@@ -63,8 +63,12 @@ class LocalMediaProgress: Object, Codable {
...
         try container.encode(localEpisodeId, forKey: .localEpisodeId)
         try container.encode(duration, forKey: .duration)
-        try container.encode(progress, forKey: .progress)
-        try container.encode(currentTime, forKey: .currentTime)
+        if progress.isNaN == false {
+            try container.encode(progress, forKey: .progress)
+        }
+        if currentTime.isNaN == false {
+            try container.encode(currentTime, forKey: .currentTime)
+        }
```
ronaldheft, line 71: This is honestly to fix an introduction of bad data into my database during development. I left it in, as it is desirable not to bomb out on encoding if these values are NaN.

ronaldheft:
I want to give a day or so testing on my phone before I mark this as ready, but I think I have addressed all the various scenario with this refactor.

ronaldheft:
Found one issue that needs to be fixed: pausing and playing repetitively with a sleep timer set reduces the amount of sleep time. I will get that fixed before marking this as ready.

advplyr:
If the listening time is going to be based on the session last update timestamp does this take into account if the player is paused?

1. User starts listening for 1 min
2. Pauses for 1 min
3. Pressing play again should still be listening time of 1 min

ronaldheft:
@advplyr Great call out! It does!

First, the sync to method is only called during playback. Specifically:
1. When playback is started
2. Every 0.5 wall-clock seconds
3. When paused or the player is closed

Second, there is a flag if the progress sync included progress during playback: https://github.com/advplyr/audiobookshelf-app/pull/330/files#diff-901b57b98bd64390120848c0802fb3fb81817426667cf6ad461c189fc2ff7d9cR60

Finally, this flag is set on sync scenarios 2 and 3. Sync method 1, does not update the session's `timePlaying`.

So in your scenario, during steps 1 and 2, there are a number of +0.5 seconds added to the `timePlaying` value of the session. This stops when paused at step 2. When played against at step 3, continue adding +0.5 seconds.

ronaldheft:
I'm also happy to report the new progress syncing logic appears to be working well. Progress remained accurate during my commute, which involved:

* Car via CarPlay
* Bluetooth
* Period of time where I did a bunch of things on my phone during playing
* Stopping playing via remote
* Work for a while, using my phone occasionally but not using ABS
* Press play via the remote
* Playback resumed where expected, including when I opened the app after it was in the background

I did have it not reopen the in-app playback controls when coming back to the app after a while, so I suspect I need to add an event when the app returns from the background. However, despite this, I was able to force quit and reopen, the player came back and no progress was lost. So the session was being updated, just a UI glitch which I can fix.

advplyr:
Awesome, great progress. I was just testing out the session sync and caught a bug. I will put it here incase it is something you can adjust easily in this PR.

1. Connect to a server and download a local book.
2. Disconnect from that server and login to another server.
3. Listen to the downloaded book
4. Session sync requests are sent to the current (wrong) server, showing "Library item not found" in the other server.

ronaldheft:
> Awesome, great progress. I was just testing out the session sync and caught a bug. I will put it here incase it is something you can adjust easily in this PR.
> 
>     1. Connect to a server and download a local book.
> 
>     2. Disconnect from that server and login to another server.
> 
>     3. Listen to the downloaded book
> 
>     4. Session sync requests are sent to the current (wrong) server, showing "Library item not found" in that server.

Ah, I bet this is actually related to the active session being restored on a startup. I just added this fix to factor in the server config to the active session: https://github.com/advplyr/audiobookshelf-app/pull/330/commits/10ddc1c9117c6df6a79a274cd63b56fe77a8849e

advplyr:
I noticed that 2 local session sync requests are sent to the server back-to-back when first playing an item.
They come in within a second of each other usually and I only caught it because 2 duplicate sessions got added to the server.

Part of this is a server update I will make which will be locking a session while it is still being updated. What happened was when first inserting the new session the write to the db took a bit longer so when the next request came it it also inserted it instead of updated. On every other test I did the db write was fast enough that the second request updated.

advplyr:
I updated the server with a lock on local sessions syncing to prevent duplicates: https://github.com/advplyr/audiobookshelf/commit/e611d7a8fd01e793c6cc3add28e60f17e1ebe851

ronaldheft:
> I noticed that 2 local session sync requests are sent to the server back-to-back when first playing an item. They come in within a second of each other usually and I only caught it because 2 duplicate sessions got added to the server.
> 
> Part of this is a server update I will make which will be locking a session while it is still being updated. What happened was when first inserting the new session the write to the db took a bit longer so when the next request came it it also inserted it instead of updated. On every other test I did the db write was fast enough that the second request updated.

Sweet! Glad there is a server-side fix as well. I did track down the issue from the app side, and it was also due to write transactions not being respected: https://github.com/advplyr/audiobookshelf-app/pull/330/commits/2a2ebefeb987554a4aae7f30f373777503bc17e1

There should not be two back-to-back sessions updates from the app. I was expecting an update when starting to play, and then an update every 10 seconds. The transaction issue lead to two back-to-back requests.

ronaldheft:
Issues I'm still tracking on this PR:
* Sleep timer losing time when pausing/playing (just haven't looked at it yet)
* An occasional crash related to Realm when the app is the background
  * I think this is thread related, possibly with the await/async and using different run queues

ronaldheft:
I did a bunch of reading on Realm with Swift's async/await logic, and the general sense I'm getting is Realm has not fully been optimized to run in that context. It's a simple fix though, just move any modifications back to a DispatchQueue thread, which I did with the latest commit.

I have a commute again tomorrow, so should get plenty of opportunity to verify my fix. That leaves the following remaining:

* Sleep timer losing time when pausing/playing (just haven't looked at it yet)
* ~~An occasional crash related to Realm when the app is the background~~
* Player UI not appearing if iOS closed or suspended the webview while playing (missed this on my last update)
  * I suspect this is just hooking into the foreground notifications and alerting the UI

ronaldheft:
Still digging on the crash, so I am anticipating this will not be ready for this weekend. I found out how to reproduce the crash, and it appears memory related... possibly a memory leak. I'll keep on digging, but not anticipating I'll have this ready until later in the weekend or early next week.

ronaldheft:
Haha, found the damn memory leak! May be able to have this ready for this weekend after all. https://github.com/advplyr/audiobookshelf-app/pull/330/commits/5c76158729cf2a1e00cc3d17a4f15bcd400bdd73

Turns out calling Realm's `freeze()` does not create a detached Realm object, but disconnect the object from live updates and makes it thread-safe. Realm still tracks changes to all objects in the Realm in memory while the object is frozen so when `thaw()` is called, it can catch the object up.

The issue was the ServerConfiguration was stored as `frozen()` object, and every database change, such as progress updates, were accumulating in memory. Refactoring that logic so a frozen object is not held in memory resolved the slow leak leading to a crash.

Now to fix the other issues, but the issue racking my brain for the past several days is solved!

ronaldheft:
Alright, fixing that memory leak led to an evening of productivity! No more known issue on my end. Going to keep on hammering at it looking for bugs, but I think if everyone else feels comfortable with this change, this should be a nice improvement to sync reliability and sleep timer reliability.

advplyr:
Awesome! I'll give it a thorough testing tomorrow

ronaldheft:
Added a fix for an edge case with the chapter sleep timer. Other than that, so far, so good

advplyr:
I downloaded a book and been listening checking in on the listening session. The listening session updates are either not updating the `currentTime` or not including it.

The column for "Last Time" is showing the `currentTime` on the session and it is not being updated after the initial session open.

image

ronaldheft:
@advplyr I just added a debug statement that shows the currentTime being updated in the app before being sent to the server. Also, if you start a new playback session from the web player, it will start with the correct time. I think something is just wonky with listened history page.

I can dig on it more after figuring out the issue with the servers.

advplyr:
I tried to break it but could not! Very solid

PR #338, opened 2022-08-30
Fix: Extra timeListening when player initializes on iOS

ronaldheft:
There was an edge case where extra `timeListening` would be added to the playback session.

This turned out to be the `PeriodicTimeObserver` firing on initialization of the audio player, which was marking the first update as including playback progress, thus adding the time since player was last opened to the playback session.

This PR adds an extra check that the player is actually playing something instead of marking all progress updates from the `PeriodicTimeObserver` as including playback progress.


advplyr:
Does this PR address https://discord.com/channels/942908292873723984/981321213882282035/1014225891766833264 ?

I'm assuming so just making sure. We may want to push an iOS release soon to patch this one.

ronaldheft:
@advplyr Yes, this is the fix for it.

PR #359, opened 2022-09-02
Rewrite iOS sleep timer to handle multi-track files

### LINKED ISSUE #351: Sleep timer fails to stop playing with multi-file books on iOS
I am actively working on a fix for this. Logging so I don't forget.

---

### Steps to reproduce

1.
2.
3.

### Expected behaviour
- Tell us what should happen

### Actual behaviour
- Tell us what happens

### Environment data

Audiobookshelf Version:

- [ ] Android App?
- [ ] iOS App?

#### Android Issue
Android version:

Device model:

Stock or customized system:

#### iOS Issue
iOS Version:

iPhone model:

ronaldheft:
This took me quite a while to figure out, but I rewrote the iOS sleep timer again to address issues with multi-track books. Fixes #351.

The iOS sleep timer now:
* Uses a `Timer` for countdown-based sleep timers
* Uses a `TimeObserver` for a scheduled chapter end time

This seems to strike the balance of accuracy, reliability, and code readability. I have tested this PR with:
* Single-track books
* Multi-track books
* Podcasts

For each media type, I have verified:
* How the timer reacts to play/pause
* How the timer reacts to changes in playback speed
* How the timer reacts to seeking
* How the timer handles seeking past the end time (in case of a chapter end timer)

Why rewrite the sleep timer again? While the previous strategy worked, it involved recalculating the stop time each time seeking and playback rate changes occurred. Not to mention the edge cases with a chapter sleep time.

In theory, the previous strategy should work, but the way the player reinitialized when tracks are changed out during seeks with multi-file books caused the stop time to be recalculated wrong depending on the order of operations. I tried a number of different strategies to debug it, and race conditions would continue to come up, due to the fact that pause/stopping audio causes the playback rate to change on iOS. This would force the stop time to be recalculated, which would be occurring while the seek was happening. There is probably a way to solve for it, but I opted to simplify the logic.

### DIFF
--- a/ios/App/Shared/player/AudioPlayer.swift
+++ b/ios/App/Shared/player/AudioPlayer.swift
@@ -178,6 +178,11 @@ class AudioPlayer: NSObject {
                 if self.isSleepTimerSet() {
                     // Update the UI
                     NotificationCenter.default.post(name: NSNotification.Name(PlayerEvents.sleepSet.rawValue), object: nil)
+                    
+                    // Handle a sitation where the user skips past the chapter end
+                    if self.isChapterSleepTimerBeforeTime(currentTime) {
+                        self.removeSleepTimer()
+                    }
                 }
             }
         }
--- a/ios/App/Shared/player/AudioPlayer.swift
+++ b/ios/App/Shared/player/AudioPlayer.swift
@@ -100,20 +100,21 @@ class AudioPlayer: NSObject {
 
         NSLog("Audioplayer ready")
     }
+    
     deinit {
         self.stopPausedTimer()
         self.removeSleepTimer()
         self.removeTimeObserver()
         self.queueObserver?.invalidate()
         self.queueItemStatusObserver?.invalidate()
-        destroy()
     }
+    
     public func destroy() {
         // Pause is not synchronous causing this error on below lines:
         // AVAudioSession_iOS.mm:1206  Deactivating an audio session that has running I/O. All I/O should be stopped or paused prior to deactivating the audio session
         // It is related to L79 `AVAudioSession.sharedInstance().setActive(false)`
-        pause()
-        audioPlayer.replaceCurrentItem(with: nil)
+        self.pause()
+        self.audioPlayer.replaceCurrentItem(with: nil)
         
         do {
             try AVAudioSession.sharedInstance().setActive(false)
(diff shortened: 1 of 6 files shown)


ios/App/Shared/player/AudioPlayer.swift
```diff
@@ -378,32 +375,24 @@ class AudioPlayer: NSObject {
...
             self.audioPlayer.seek(to: CMTime(seconds: seekTime, preferredTimescale: 1000)) { [weak self] completed in
-                if !completed {
-                    NSLog("WARNING: seeking not completed (to \(seekTime)")
-                }
+                guard completed else { return NSLog("SEEK: WARNING: seeking not completed (to \(seekTime)") }
+                guard let self = self else { return }
                 
                 if continuePlaying {
-                    self?.play()
+                    self.resumePlayback()
```
ronaldheft, line 387: This change prevents a condition where seeking would occur within seeking, due to `play()` also needing to seek to the proper start time (even outside of `seekBack` conditions).

ios/App/Shared/player/AudioPlayerSleepTimer.swift
```diff
@@ -0,0 +1,168 @@
...
+//  AudioPlayerSleepTimer.swift
+//  App
+//
+//  Created by Ron Heft on 9/2/22.
+//
+
+import Foundation
+import AVFoundation
+
+extension AudioPlayer {
```
ronaldheft, line 11: All the sleep timer logic is in one file now

advplyr:
Found one bug. After starting a sleep timer then changing the audiobook the sleep timer does not reset.

Then you have to cancel the sleep timer and start another to get it going again.

ronaldheft:
Good catch! I’m probably missing a call to cancel the timer on deinit. I’ll reproduce later this morning and verify that fixes the issue.

ronaldheft:
@advplyr Fixed it! Turned out `AudioPlayer` had a memory leak and `deinit` was not being called when switching player items.

Some of the `MPRemoteCommandCenter` targets had strong references to `self`, one of timers I had added had a strong reference, and the `AVAudioPlayer` observers also held strong references to `self`. FYI, I found [this article](https://stackoverflow.com/questions/41660486/knowing-where-retain-cycles-are-and-removing-them) really helpful in tracking down the bits of code holding onto references to the `AudioPlayer`.

Outside of fixing the memory leak, I also took a second look at the weird code I wrote to clear the chapter sleep timer when you skip back, and I rewrote that in a way that was not awkward.

advplyr:
Sleep timer is working perfectly in all my tests. Nice!

PR #361, opened 2022-09-03
Fix: Handle iOS audio events

### LINKED ISSUE #343: Handle iOS system audio events and resume playblack when necessary
Right now ABS on iOS is not aware of system audio events. This means if driving directions or Siri pauses audio from ABS, playback will not resume after the audio is over. Additionally, this also may mean audio continues to play when disconnecting headphones.

ABS for iOS should implement the system audio API events, and pause and resume audio for appropriate scenarios.

None. This is a requirement for a CarPlay app, as iOS audio apps are not approved for CarPlay if they're not a good iOS audio citizen (been reading the CarPlay docs 😄 )

ronaldheft:
This PR addresses several issues (fixes #343):
* The active audio session is now changed to `false` when the player is paused
  * This also ensures the below fixes do not trigger when the player is paused
* When an audio interruption occurs (such as Siri, directions in CarPlay, a phone call, etc), if ABS was playing before the interruption occurred, ABS will resume playback after the interruption
  * The player will use the `skipBack` setting when resuming, so for long interruptions, audio will be allowed to rewind back
* If plugging in headphones, audio will continue in the headphones
* If disconnecting headphones, audio will stop playing
  * There may be other scenarios we want to pause audio, but I'm not sure of them at this moment
  * We now have the framework to add others in the future

Note: This PR also includes #359 as there was a merge conflict that needed to be fixed between them. If you would like to review the code for this without #359, I can trigger an update to this PR after #359 is merged


ronaldheft:
Synced this branch after the merge, so the PR shows only the code changed for this issue.

advplyr:
I tested connecting/disconnecting headphones and it worked as you described.

I also tested with maps and the pause then seek back worked great.

Issue #358, opened 2022-09-02
Play/pause remote button on Android TV not working as expected


galaxyfar:
### Steps to reproduce

1. v0.9.57-beta on Nvidia Shield (also happening on generic Android TV box, so not a Shield issue per se)
2. Play any audiobook, press play/pause remote button
3. Play will either not pause (if already playing) or with resume (if already pause)

### Expected behaviour
- Play/pause remote button should play/pause playback (this was not an issue with older version, sorry I lost track, which was last stable release)

### Actual behaviour
- Play/pause remote button it acting wonky with only this app. Other music app responds to play/pause as expected

### Environment data

Audiobookshelf Version:

- [ v0.9.57-beta] Android App  

#### Android Issue
Android version: 11 

Device model: Nvidia shield TV Pro

Stock or customized system: No

dalins:
### Steps to reproduce

Same as original post.

**Additional information:**
When a Bluetooth headset is used and the app is in the foreground, it works. But when the app is in the background, it does not work. I also tried a wired headset and with that it doesn't work in the foreground or background.

### Expected behaviour

Same as original post.

### Actual behaviour

Same as original post.

### Audiobookshelf Version:

[ v0.9.57-beta] Android App

### Android Issue

Android version: 11

Device model: Hallo Welt Systeme / Volla Phone

Stock or customized system: Stock system (Android without Google Services)

advplyr:
This should be fixed in the next release. Along with #427 #429

advplyr:
Fixed in [v0.9.62-beta](https://github.com/advplyr/audiobookshelf-app/releases/tag/v0.9.62-beta)

Issue #352, opened 2022-09-01
Bluetooth remote control broken since update to v0.9.57-beta


andonevris:
Not sure if something changed recently but previously I could control playback skipping back and forwards from my watch via bluetooth connection to phone, since the last update skipping back still works but skipping forwards does not. Note this is for remote control, skipping still works from phone directly.

I should also add my watch and many other remote control devices only have a "next track" button so other podcast/audiobook players commandeer the "next track" command and translate that to skip forwards as it is far more useful to be able to skip rather than go to the next podcast or audiobook.

### Steps to reproduce

1. Start book or podcast on phone
2. Attempt to skip forwards from bluetooth remote control
3. Nothing happens

### Expected behaviour
- Pressing next track should skip forwards as previously or at least have a setting, as is standard on many audio players, to choose whether user wishes to use that function for "next track" or "skip"

### Actual behaviour
- Nothing happens.

### Environment data

Audiobookshelf Version:
v2.1.4

- [ x] Android App?
- [ ] iOS App?

#### Android Issue
Android version: 10

Device model: Galaxy S10

Stock or customized system: Stock, Rooted

advplyr:
This is from #328 for android auto

vpoluyaktov:
The same problem on IOS devices

advplyr:
Is this still an issue in app version 0.9.61?

There is a fix for the next release relating to play/pause from bluetooth.
Related issues #427 #429 #573

andonevris:
Yes still an issue, it was never fixed.

Forward key does nothing, back key jumps all the way back to the beginning.

advplyr:
What device are you using to press forward/backward?

andonevris:
I'm using a Garmin Fenix smart watch, it used to work fine but then something changed  around September 2022

advplyr:
Are there other bluetooth devices that you use that are also not working?

andonevris:
I don't have any other devices I could test.  Maybe some of the others who have this issue can help?

advplyr:
Those others with issues with iOS are separate and was updated in the last release.

There is one other issue with Android bluetooth control and it is also a watch. Pebble smart watch #423

andonevris:
That Pebble smart watch issue has exactly the same symptoms, it could be the same issue.

advplyr:
Most likely. The best way to address this will be to build from source and watch the logcat while pressing the controls. I'm not sure what events these watches are firing.

[...]

PR #363, opened 2022-09-03
Add debug build to to allow prod and dev apps to work side by side.

vangorra:
Development can be made better by allowing the prod and dev apps to coexist.

Current process to develop with Android auto:
Send Android auto will only work with apps distributed from Google play. Currently, in other to develop the app, one has to uninstall the current (prod) version, and setup the debug version for development. Then when we want to play Android auto on real hardware, one must remove the dev app, install the prod and setup again.

Solution:
Provide an alternative application id for debug build. This allows both apps to coexist and avoids all re-setup. The apps are easily distinguished because the debug app has an upsidedown icon.

Book with multiple tracks:
skip

Book with one track:
no_skip

Note: This is the first of several PRs that fix Android auto buttons and queue. I finally figured it out, it took a lot of research.


advplyr:
Nice I like that. Have you seen this issue? #352

I'm not sure the best way to handle this as some devices would prefer the next/prev buttons to jump forward/backward. Particularly the ones that don't have a display like headphone buttons.

I'm not sure if this one is related to the AA update #358. I don't think so since we didn't touch the play/pause buttons.

vangorra:
I doubt 352 or 358 are caused by my last change but I'm unwilling to rule it out. As you know, Google has provided a hot mess of APIs to use for various devices and it's easy to mess it up. I'll see if there is a small patch that can help.
I'm investigating methods of organizing this better. I suspect there aren't any better ones. So I'll likely end up working a bunch of tests to lock in functionality.

advplyr:
I looked at #352 again and I was mistaken. I thought I remembered the behavior of the next/prev buttons getting changed but I see now it wasn't.

As far as organizing this better I think the PlayerNotificationService is long overdue for a refactor. A lot of this was me hacking things together and learning on the fly with the intention of circling back to clean it up later.
There is one particular issue that has been bugging me for months and it is showing the cover art of downloaded books. Android auto will not display images using a url to the local file system.

PR #365, opened 2022-09-04
Cleanup Android Auto buttons

vangorra:
Control buttons in Android Auto (and probably Android Wear too) are messy. The skip back/forward buttons appears sometimes and the jump forward/back a few seconds don't show at all in minimized modes.

This PR does the following:
- Show skip buttons if multiple items are available. minimized
- Hide skip buttons when multiple queue items are not available. skip_hidden
- Show jump buttons when the app is "minimized" instead of skip buttons. skip_shown


advplyr:
Tested multi-track and single track audiobooks and it works great.

PR #370, opened 2022-09-07
Fix: Pause if clearing the sleep timer under a second

### LINKED ISSUE #367: iOS chapter sleep timer does not always stop at chapter end
### Steps to reproduce

1. Set a sleep timer to the end of the chapter
2. Wait for the end of the end of the chapter (can skip forward to reduce the wait time)
3. Observe that a small percentage of the time, playback does not stop, yet the sleep timer is cleared
4. This doesn't always happen, so it make take several tries to reproduce

### Expected behaviour
- The chapter sleep timer should always stop playback

### Actual behaviour
- The chapter sleep timer only most of the time stops playback

### Proposed fix
- This is related to logic for clearing the chapter sleep timer if you skip past the chapter end
- I will either ensure that the sleep timer logic can be guaranteed to run before the sleep timer is cleared, or I will ensure that the timer clear logic also stops playback if within the time range the sleep timer is supposed to end

### Environment data

Audiobookshelf Version:

- [ ] Android App?
- [x] iOS App?

#### Android Issue
Android version:

Device model:

Stock or customized system:

#### iOS Issue
iOS Version:

iPhone model:

ronaldheft:
Fixes #367

### DIFF
--- a/ios/App/Shared/player/AudioPlayer.swift
+++ b/ios/App/Shared/player/AudioPlayer.swift
@@ -193,10 +193,6 @@ class AudioPlayer: NSObject {
                     
                     // Handle a sitation where the user skips past the chapter end
                     if self.isChapterSleepTimerBeforeTime(currentTime) {
-                        // If we are firing before the chapter sleep timer trigger
-                        if let timeUntilSleep = self.getSleepTimeRemaining(), timeUntilSleep <= 1 {
-                            self.pause()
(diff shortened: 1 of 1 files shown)


ronaldheft:
Set this as a draft, as I want to test more before considering this ready.

ronaldheft:
Closing in favor of #392

PR #371, opened 2022-09-07
Improve iOS Logging

ronaldheft:
* Switches all classes to use the newer [OSLog framework provided by Apple](https://developer.apple.com/documentation/os/logging/generating_log_messages_from_your_code)
  * This allows logs to be captured via a sysdiagnosis even when not actively capturing logs on a development machine for easier issue diagnosis
* Adds a helper Logger class to make interacting with the OSLog framework cleaner
* Renames the build target from the generic `App` to `Audiobookself` to make finding log messages easier among console logs


ronaldheft:
This is now complete, and I should have all classes converted.

PR #386, opened 2022-09-16
Fix various iOS app startup issues

### LINKED ISSUE #374: iOS , book autoplays, when opening the app after the phone is left for an extended period of time, controls in app doesn't show up
### Steps to reproduce

1. Play an audiobook pause it or let it do the sleep function (sleep function is what triggered it last time, I set sleep to 30min)
2. wait some time, seems to be over an hour , last time for me was 8 hours (never manually close the app)
3.click the app icon.

### Expected behaviour
- audiobook app opens, latest book is in a paused state, I can see the player controls in app

### Actual behaviour
- Audibook app opens, last audiobook listened to starts auto playing. There is no controls. There is lockscreen controls so I can pause the book. If I click the same book again, it will pause for a second( like a new instance is opening) and now it will work correctly.

### Environment data
Audiobookshelf server is on docker
Audiobookshelf Version: server: 2.1.5 , IOS 0.9.58

- [ ] Android App?
- [x ] iOS App?

#### iOS Issue
iOS Version: 16

iPhone model: Iphone XS Max

ronaldheft:
This PR addresses a few different issues related to app startup:

- Fixes a warnings related to accessing `window` too early
- Fixes a crash where a race condition can cause the AudioPlayer's rate to be modified off the main queue
- Fixes #374 by ensure if WebKit is reloaded and the native code is not, the two can get back in sync

### DIFF
--- a/ios/App/App.xcodeproj/project.pbxproj
+++ b/ios/App/App.xcodeproj/project.pbxproj
@@ -34,7 +34,7 @@
 		504EC30F1FED79650016851F /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 504EC30E1FED79650016851F /* Assets.xcassets */; };
 		504EC3121FED79650016851F /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 504EC3101FED79650016851F /* LaunchScreen.storyboard */; };
 		50B271D11FEDC1A000F3C39B /* public in Resources */ = {isa = PBXBuildFile; fileRef = 50B271D01FEDC1A000F3C39B /* public */; };
-		A084ECDBA7D38E1E42DFC39D /* Pods_App.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = AF277DCFFFF123FFC6DF26C7 /* Pods_App.framework */; };
+		AC98F95F300E46A27FAE47A4 /* Pods_Audiobookshelf.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = B775656E49492CFD6763B7B6 /* Pods_Audiobookshelf.framework */; };
 		E9D5504628AC1A3900C746DD /* LibraryItem.swift in Sources */ = {isa = PBXBuildFile; fileRef = E9D5504528AC1A3900C746DD /* LibraryItem.swift */; };
 		E9D5504828AC1A7A00C746DD /* MediaType.swift in Sources */ = {isa = PBXBuildFile; fileRef = E9D5504728AC1A7A00C746DD /* MediaType.swift */; };
 		E9D5504A28AC1AA600C746DD /* Metadata.swift in Sources */ = {isa = PBXBuildFile; fileRef = E9D5504928AC1AA600C746DD /* Metadata.swift */; };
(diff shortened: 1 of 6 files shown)


ios/App/App.xcodeproj/project.pbxproj
```diff
@@ -34,7 +34,7 @@
 		504EC30F1FED79650016851F /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 504EC30E1FED79650016851F /* Assets.xcassets */; };
 		504EC3121FED79650016851F /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 504EC3101FED79650016851F /* LaunchScreen.storyboard */; };
 		50B271D11FEDC1A000F3C39B /* public in Resources */ = {isa = PBXBuildFile; fileRef = 50B271D01FEDC1A000F3C39B /* public */; };
-		A084ECDBA7D38E1E42DFC39D /* Pods_App.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = AF277DCFFFF123FFC6DF26C7 /* Pods_App.framework */; };
+		AC98F95F300E46A27FAE47A4 /* Pods_Audiobookshelf.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = B775656E49492CFD6763B7B6 /* Pods_Audiobookshelf.framework */; };
```
ronaldheft, line 37: The target changed in #371, but I missed updating the Pods config for the new target.

ios/App/App/AppDelegate.swift
```diff
@@ -85,10 +85,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
         super.touchesBegan(touches, with: event)
 
-        let statusBarRect = UIApplication.shared.statusBarFrame
+        let statusBarRect = self.window?.windowScene?.statusBarManager?.statusBarFrame
```
ronaldheft, line 88: Fixes a window access deprecation with iOS 13+

ios/App/App/AppDelegate.swift
```diff
@@ -7,7 +7,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     
     private let logger = AppLogger(category: "AppDelegate")
 
-    var window: UIWindow?
+    lazy var window: UIWindow? = UIWindow(frame: UIScreen.main.bounds)
```
ronaldheft, line 10: Fixes a warning about accessing `window` too early

ios/App/App/plugins/AbsAudioPlayer.swift
```diff
@@ -14,7 +14,6 @@ public class AbsAudioPlayer: CAPPlugin {
     private let logger = AppLogger(category: "AbsAudioPlayer")
     
     private var initialPlayWhenReady = false
-    private var isUIReady = false
```
ronaldheft, line 17: Unused variable removed. I added this in a previous PR and never removed it.

ios/App/App/plugins/AbsAudioPlayer.swift
```diff
@@ -35,15 +34,14 @@ public class AbsAudioPlayer: CAPPlugin {
     }
     
     func restorePlaybackSession() async {
-        // We don't need to restore if we have an active session
-        guard PlayerHandler.getPlaybackSession() == nil else { return }
-
```
ronaldheft, line 40: We actually do need to restore, in the case of the WebKit reloading and existing playback session is already in the player.

ios/App/App/plugins/AbsAudioPlayer.swift
```diff
@@ -35,15 +34,14 @@ public class AbsAudioPlayer: CAPPlugin {
...
-        guard PlayerHandler.getPlaybackSession() == nil else { return }
-        
         do {
             // Fetch the most recent active session
             let activeSession = try await Realm().objects(PlaybackSession.self).where({
                 $0.isActiveSession == true && $0.serverConnectionConfigId == Store.serverConfig?.id
             }).last?.freeze()
+            
             if let activeSession = activeSession {
+                PlayerHandler.stopPlayback(currentSessionId: activeSession.id)
```
ronaldheft, line 44: When reloading the session, we should stop playback so audio doesn't play while the next line's network connection is made.

This is an issue because the Vue code will change playback rate during startup, which will unpause the audio. By stopping playback, changing the playback rate ensures audio doesn't start playing.

ios/App/Podfile
```diff
@@ -17,7 +17,7 @@ def capacitor_pods
   pod 'CapacitorStorage', :path => '../../node_modules/@capacitor/storage'
 end
 
-target 'App' do
+target 'Audiobookshelf' do
```
ronaldheft, line 20: The target changed in #371, but I missed updating the Pods config for the new target.

ios/App/Shared/player/AudioPlayer.swift
```diff
@@ -104,9 +104,6 @@ class AudioPlayer: NSObject {
     }
     
     deinit {
-        self.stopPausedTimer()
-        self.removeSleepTimer()
-        self.removeTimeObserver()
```
ronaldheft, line 109: These needed to be moved to `destroy()`, as when setting `player = nil`, `removeSleepTimer` will throw an event that causes `PlayerHandler.player` to be read, which is causes an access crash due to a mutated variable being read.

ios/App/Shared/player/AudioPlayer.swift
```diff
@@ -133,8 +130,13 @@ class AudioPlayer: NSObject {
         // Remove observers
         self.audioPlayer.removeObserver(self, forKeyPath: #keyPath(AVPlayer.rate), context: &playerContext)
         self.audioPlayer.removeObserver(self, forKeyPath: #keyPath(AVPlayer.currentItem), context: &playerContext)
+        self.removeTimeObserver()
         
         NotificationCenter.default.post(name: NSNotification.Name(PlayerEvents.closed.rawValue), object: nil)
+        
+        // Remove timers
+        self.stopPausedTimer()
+        self.removeSleepTimer()
```
ronaldheft, line 139: Move these from `deinit`.

ios/App/Shared/player/AudioPlayer.swift
```diff
@@ -300,9 +302,12 @@ class AudioPlayer: NSObject {
         // Start playback, with a seek, for as smooth a scrub bar start as possible
         let currentTrackStartOffset = session.audioTracks[self.currentTrackIndex].startOffset ?? 0.0
         let seekTime = currentTime - currentTrackStartOffset
-        self.audioPlayer.seek(to: CMTime(seconds: seekTime, preferredTimescale: 1000), toleranceBefore: .zero, toleranceAfter: .zero) { [weak self] completed in
-            guard completed else { return }
-            self?.resumePlayback()
+        
+        DispatchQueue.runOnMainQueue {
```
ronaldheft, line 306: Actions that modify the `audioPlayer` must be on the main queue. This was the case 99% of the time before, but by adding this code, we are being explicit it's happening.

ios/App/Shared/player/PlayerHandler.swift
```diff
@@ -15,10 +15,7 @@ class PlayerHandler {
         guard let session = Database.shared.getPlaybackSession(id: sessionId) else { return }
         
         // Clean up the existing player
-        if player != nil {
-            player?.destroy()
-            player = nil
-        }
+        resetPlayer()
```
ronaldheft, line 18: Minor refactor as I was tracking downing the `player = nil` access error.

ios/App/Shared/player/PlayerHandler.swift
```diff
@@ -31,15 +28,11 @@ class PlayerHandler {
         player = AudioPlayer(sessionId: sessionId, playWhenReady: playWhenReady, playbackRate: playbackRate)
     }
     
-    public static func stopPlayback() {
+    public static func stopPlayback(currentSessionId: String? = nil) {
```
ronaldheft, line 31: New parameter ensures the current playback session does not get invalidated during restore, which would cause it not be restored the next time.

PR #413, opened 2022-10-16
Colored status and navigationbar

wnhrt:
On some Samsung devices, white is used as the default color for the navigation bar in dark mode, see https://github.com/jellyfin/jellyfin-android/issues/581.

I colored the status and navigationbar with the same background color as the header of the website.

image

There is no issue for this PR.


advplyr:
Nice thanks

PR #455, opened 2022-12-07
Don't allow webview to scroll out of bounds vertically

benonymity:
Prevents the app's webview from allowing a scroll of the actual app body, as per issue [#5907](https://github.com/ionic-team/capacitor/issues/5907) in the Capacitor repo. Previously the app could behave like this:
verticalScroll


advplyr:
Seems to work for me. Not sure how you tested it since I messed up the lockOrientation update but that is fixed now

benonymity:
I just stripped out code until it built in XCode haha. Glad that's working now!

PR #456, opened 2022-12-07
Fix Listening Stats

### LINKED ISSUE #404: User stats showing incorrect time (iOS and PC)
### Steps to reproduce

1. Open up AudioBookShelf
2. click the three bars in upper right hand corner
3. Click on user stats
4. Time is not being displayed correctly

### Expected behaviour
- The correct amount of minutes for Week, Daily and Best should be displayed correctly

### Actual behaviour
- It appears that maybe milliseconds are being displayed vs minutes/hours

### Environment data

Audiobookshelf Version: 2.2.0

- [ ] Android App?
- [x] iOS App?

#### Android Issue
Android version:

Device model:

Stock or customized system:

#### iOS Issue
iOS Version: 16.0.2 though it was happening on 15.x.x

iPhone model: 11 11 pro
IMG_2769

benonymity:
Resolves #404. The issue seems to be that the server expects that when syncing an open (streamed) session, the attribute `timeListened` to be time since the last sync to the server (as per the [API docs](https://api.audiobookshelf.org/#sync-an-open-session)). But the app uses that logic to store timeListened offline as well, so updates it cumulatively and sends that to the server where the time is compounded. This explains why the listening sessions seemed only slightly off for a few seconds of listening, but exponentially huge (thousands of hours) for longer streamed listens. This PR just adds a bit of logic to reset that timeListening attribute each time there is a server sync when the book is being streamed. I did some testing both with downloaded and local books and the listening time now seems to be reported accurately.

### DIFF
--- a/ios/App/Shared/player/PlayerProgress.swift
+++ b/ios/App/Shared/player/PlayerProgress.swift
@@ -151,6 +151,14 @@ class PlayerProgress {
         } else {
             let playbackReport = PlaybackReport(currentTime: session.currentTime, duration: session.duration, timeListened: session.timeListening)
             success = await ApiClient.reportPlaybackProgress(report: playbackReport, sessionId: session.id)
+            // Reset time listening becuase server expects that time to be since last sync, not for the whole session
+            if success {
+                if let session = session.thaw() {
+                    try session.update {
+                        session.timeListening = 0
+                    }
+                }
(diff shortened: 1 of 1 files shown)


advplyr:
I did some testing and it is working for me but I never reproduced the original issue.

PR #458, opened 2022-12-07
Prevent selecting of text in the app

### LINKED ISSUE #379: iOS app - when long pressing, app selects and highlights.
On iOS, anywhere in the app, if you long press it selects text, pictures, links, buttons. It will even select text inside cover art.
you can even long press cover art to save to your device. It even will click and drag items around the screen.

Visual bug, but annoying.

benonymity:
Fixes #379 (mostly). Prevents selecting text in the app, but does not prevent long-press to download covers on iOS. Seems like an appropriate "bug" to keep as it exists on the item page?

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="w-full layout-wrapper bg-bg text-white">
+  <div class="select-none w-full layout-wrapper bg-bg text-white">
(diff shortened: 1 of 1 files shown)


advplyr:
I think it works as a partial fix for text for now. Not sure if there is another solution that people implement for this.

PR #459, opened 2022-12-08
Add haptic feedback

benonymity:
Adds haptic feedback to all buttons that perform a tangible action (play/pause, change setting, download book, etc.) I don't have access to a physical Android device, does the haptic feedback respond in the same way on those devices? This makes a ton of functions async but it doesn't seem to have any negative impacts that I noticed. Adds another dimensions in interacting with the app.

### DIFF
--- a/components/home/BookshelfToolbar.vue
+++ b/components/home/BookshelfToolbar.vue
@@ -8,7 +8,7 @@
         <p v-show="!selectedSeriesName" class="font-book pt-1">{{ totalEntities }} {{ entityTitle }}</p>
         <p v-show="selectedSeriesName" class="ml-2 font-book pt-1">{{ selectedSeriesName }} ({{ totalEntities }})</p>
         <div class="flex-grow" />
-        <span v-if="page == 'library' || seriesBookPage" class="material-icons px-2" @click="bookshelfListView = !bookshelfListView">{{ !bookshelfListView ? 'view_list' : 'grid_view' }}</span>
+        <span v-if="page == 'library' || seriesBookPage" class="material-icons px-2" @click="changeView()">{{ !bookshelfListView ? 'view_list' : 'grid_view' }}</span>
         <template v-if="page === 'library'">
           <div class="relative flex items-center px-2">
             <span class="material-icons" @click="showFilterModal = true">filter_alt</span>
@@ -25,6 +25,8 @@
 </template>
 
 <script>
+import { Haptics, ImpactStyle } from '@capacitor/haptics';
+
 export default {
   data() {
     return {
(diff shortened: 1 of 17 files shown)


advplyr:
I made a haptics plugin a while ago but didn't actually put it to use.

You don't have to import the haptics at the top of each page/component if you use the plugin.

So for example instead of
`await Haptics.impact({ style: ImpactStyle.Medium })`
you can do
`await this.$hapticsImpactMedium()`

All those functions you will find in https://github.com/advplyr/audiobookshelf-app/blob/master/plugins/haptics.js

If you can make that update on those and also remove the semicolons. I'm also seeing semicolons in the haptics.js file that I must have missed when I was copying those methods in from the docs.

benonymity:
Oh, didn't realize you already implemented the plugin! Though I did wonder why CapacitorHapics was already in the Podfile... Just removed the capacitor imports using the existing plugin and removed the semicolons from the `haptics.js` file.

pages/settings.vue
```diff
@@ -47,6 +47,7 @@
 
 <script>
 import { Dialog } from '@capacitor/dialog'
+import { Haptics, ImpactStyle } from '@capacitor/haptics';
```
advplyr, line 50:
```suggestion
```

pages/settings.vue
```diff
@@ -149,6 +150,7 @@ export default {
       this.saveSettings()
     },
     async saveSettings() {
+      await Haptics.impact({ style: ImpactStyle.Medium });
```
advplyr, line 153:
```suggestion
      await this.$hapticsImpactMedium()
```

components/home/BookshelfToolbar.vue
```diff
@@ -8,7 +8,7 @@
         <p v-show="!selectedSeriesName" class="font-book pt-1">{{ totalEntities }} {{ entityTitle }}</p>
         <p v-show="selectedSeriesName" class="ml-2 font-book pt-1">{{ selectedSeriesName }} ({{ totalEntities }})</p>
         <div class="flex-grow" />
-        <span v-if="page == 'library' || seriesBookPage" class="material-icons px-2" @click="bookshelfListView = !bookshelfListView">{{ !bookshelfListView ? 'view_list' : 'grid_view' }}</span>
+        <span v-if="page == 'library' || seriesBookPage" class="material-icons px-2" @click="changeView()">{{ !bookshelfListView ? 'view_list' : 'grid_view' }}</span>
```
advplyr, line 11:
```suggestion
        <span v-if="page == 'library' || seriesBookPage" class="material-icons px-2" @click="changeView">{{ !bookshelfListView ? 'view_list' : 'grid_view' }}</span>
```

advplyr:
Tested on Android and it works great. Thanks!

PR #493, opened 2023-01-05
Make chapters identifiable

### LINKED ISSUE #479: Chapters may not be identifiable
When long chapter names are being used, it may easily happen that parts of the name is cut off. If this part identifies the latest chapter name or number, it may make chapters indistinguishable from one another. This makes it near imposible to select a specific chapter if the book has many chapters.

Example (please select chapter 9 ;-):
</img>

It would already be very helpful to simply have a number in front of the chapter title.

### Environment data

- Audiobookshelf Android App, 0.9.60-beta?

lkiesow:
When long chapter names are being used, it may easily happen that parts of the name is cut off. If this part identifies the latest chapter name or number, it may make chapters indistinguishable from one another. This makes it near imposible to select a specific chapter if the book has many chapters.

This patch addresses the issue by starting each item in the list of chapters with an index number.

This fixes #479


advplyr:
Works for me!

PR #499, opened 2023-01-08
Metadata Table Alignment

lkiesow:
This patch updates the metadata table in the book details view, turning it into a CSS grid layout. This fixes a few issues with the current layout:

- Keys and values are not vertically aligned with the value being displayed a little bit above the baseline of the key.
- There is no horizontal alignment of the table columns. In fact, it is technically not a table despite the CSS being tuned to almost look like one. Just turning this into a real grid looks cleaner.
- The code for making multi-line values not wrap is broken. This fixes the problem.

This also renames “Publish Year” to “Published”. Mainly because it is shorter.

Screenshot from 2023-01-08 17-02-43

### DIFF
--- a/pages/item/_id.vue
+++ b/pages/item/_id.vue
@@ -35,33 +35,32 @@
     <p v-if="isLocal && serverLibraryItemId" style="font-size: 10px" class="text-success py-1 uppercase tracking-widest">connected</p>
     <p v-else-if="isLocal && libraryItem.serverAddress" style="font-size: 10px" class="text-gray-400 py-1">{{ libraryItem.serverAddress }}</p>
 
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks. I think we need a new design of this entire page

lkiesow:
> I think we need a new design of this entire page

Do you have anything specific in mind? The current one is very similar to the web client in mobile mode. Just with less options:

Screenshot from 2023-01-09 00-09-57

Do you want them to look alike in the end? Or are you looking for something else entirely?

advplyr:
I think a new design entirely. Something just doesn't look right about spacing and sizes. It doesn't have to do with your PR

PR #514, opened 2023-01-17
Chapter and Total Track in Player

lkiesow:
This patch allows user to not only add a chapter track but also remove the total track. This means that you can have only the total track or only the chapter track.

This patch does not allow user to have no track at all. Disabling one track will automatically enable the other one, if that one is already disabled.

The reasoning behind this is that for long-ish audiobooks, users will not very often look at the overall progress. In fact, I find the progress bar mostly useful only for quickly seeking to a position when jumping back a few seconds is not sufficient.

In the seldom occasion that I want the overall progress, I can easily get it from the book details page.

Screenshot from 2023-01-17 00-47-05


advplyr:
Thanks!

PR #519, opened 2023-01-24
iOS Player Updates

### LINKED ISSUE #471: iOS App keeps crashing (Offline/Not on same Network)
### Steps to reproduce

1. Have a downloaded audiobook
2. Start playing
3. After a few minutes the app stops and crashes. Pausing the audiobook also crashes and won't resume till I launch the app again and keeps saying the app crashed. I have sent several emails from testflight app with stack trace of the issue.
4. This issue seems to be related to app not communicating with server when offline or not on the same network as the server.

### Expected behaviour
- Tell us what should happen
- app shouldn't crash trying to play a downloaded audiobook

### Actual behaviour
- Tell us what happens
- app crashes after playing an audiobook after awhile.

### Environment data

Audiobookshelf Version:

- [ 0.9.60 (17)] iOS App

#### iOS Issue
iOS Version: 16.2

iPhone model: iPhone 13 PM

### LINKED ISSUE #476: Excessive Battery Usage
The iOS app somehow manages to use more power when playing audiobooks in the background than other apps do with the screen on. Some of this is probably due to background sync and progress saving, but maybe if the device is in low power mode the frequency of server syncing could be significantly reduced to preserve battery life?
image

### Environment data

Audiobookshelf Version: 2.2.10

#### iOS Issue
iOS Version: 16.2

iPhone model: iPhone SE (2020)

benonymity:
Changes the Realm DB save time from multiple times per second to once every 15 seconds, and the same is true for the server sync. I have been unable to reproduce the offline crash yet on this branch, but I'll leave this a draft until I can test it more thoroughly. Hopefully ~~fixes~~ #471 though!

If the user enables low power mode, the server sync interval will be changed to 60 seconds from 15. ~~Hopefully helps with~~ Fixes #476.


benonymity:
I haven't been able to reproduce #471 on this branch so far, but I don't want to say it's solved until those who were having it often can test. Definitely reduces battery usage by extreme amounts, so for sure fixes #476! Should be ready to merge.

gh0sti:
Hope this can be pulled in soon to help us all out for the next release.

advplyr:
This has been working great for me, thanks!

PR #528, opened 2023-01-30
Handle skip track events on iOS

blampe:
This adds nextTrack/previousTrack handlers. I've never needed to skip entire chapters from my headphones, so the behavior is similar to skipForward/skipBackward -- only moving 5-30s in either direction.

This resolves an issue I've had where double/triple clicking AirPod Pros wouldn't change playback position.


advplyr:
Thanks. I was testing this and it was crashing due to `deviceSettings` getting accessed on the wrong thread.

By setting the jump forward/backward vars above this is fixed but if you change these values in settings they won't apply until you reset the player. That was already the case before this though with the jump forward/backward.

Fyi this is already how it functions on Android with next/prev chapter from bluetooth/headsets.

PR #533, opened 2023-01-31
Fix Metadata List Separators

lkiesow:
This patch fixes the separators between metadata list entries in the book detail view. Items were separated only by comma. A space was missing.

Screenshot from 2023-01-31 14-44-33

This just adds a space to both the genre as well as to the narrator separators.

Screenshot from 2023-01-31 14-43-10


advplyr:
Oops, thanks!

Issue #333, opened 2022-08-26
Add playback speed control in Android Auto


AwsomeFox:
### Is your feature request related to a problem? Please describe.
I'd like to be able to change playback speed from the android auto interface. When driving sometimes the reader changes or you change books or get to a complicated subject and you would like to speed up or slow down. 
### Describe the solution you'd like
Add a button that would cycle through speeds on the android auto interface. 
### Describe alternatives you've considered
In order to change speeds you would need to pull your phone out and open the app there.
### Additional context
I've implemented this in another app before before so I know it's possible. If I get time to get android studio setup again I may take a look at implementing this myself. [Link to pr in another repo](https://github.com/mattttvaughn/chronicle/pull/15)

advplyr:
Awesome yeah I may get a chance to put this in this weekend before we push the next release. Are those drawables you used for the playback speeds in that PR something you made? I would implement it the same way you did in that PR if I get a chance to do it this weekend.

AwsomeFox:
They were already part of the app for the regular player ui. I'm not sure where they came from.

feerlessleadr:
I'd love this as well since I listen to my books sped up an my podcasts at 1x speed.

advplyr:
Added in [v0.9.62-beta](https://github.com/advplyr/audiobookshelf-app/releases/tag/v0.9.62-beta)

PR #554, opened 2023-02-06
Add playback speed adjustment for Android Auto

KeyboardHammer:
Adds a button to adjust the play speed from the Android Auto media player.  It's not a 1-to-1 match with the current android app play speed settings, but I did copy the existing preset options show in the android dialog.  

Also save the speed setting so it stays set between plays

image

Implements #333


advplyr:
Great, thanks! There are a few things remaining that I couldn't knock out quickly.

1. The app allows for setting custom playback rates anything between 0.5 and 10 incremented by 0.1. We could just round to the nearest icon so if you have playback speed set to `0.7` it would show as `0.5` in Android Auto.
2. Floating points are stored with trailing decimals and I'm not sure the best way to approach that. When it comes to comparisons Kotlin is handling that fine, so this could just be as simple as rounding it before displaying. Not sure yet, need to think on it. (Screenshot below)
3. When the playback speed is updated in the app it should reflect in Android Auto. I don't think this is a requirement for this PR but thought it worth mentioning. We would need to refresh those custom actions in Android Auto which I'm not sure about.

image

KeyboardHammer:
Thanks for adding the event listener, it's been years since I've done android development (pre kotlin) so I didn't realize it was so simple now.

1. I added some ranges to both the icon display and the speed change so it's a little more fluid if a custom range is set.  I hesitate to add any more steps or custom ranges to Android auto so it doesn't become a distraction while driving. Not to mention anything higher than 2 is unusable for me :).
2. I fixed the conversion issue before save so it shouldn't need to be handled anywhere else, although I'm curious why the phone app doesn't have that issue saving values like 1.2f.
3. Yeah that's a nice to have but the idea is you shouldn't have to use the phone at all while driving so it's an edge case that doesn't really need to be solved.

advplyr:
I agree that we shouldn't add a bunch of playback speed options in android auto. This is going to have people posting bug reports and bringing it up in reviews, but I'm not sure of a better way. I think I'll check out how some other apps handle it.

I was able to implement the speed refreshing in android auto when you update it in the app. I also found a previous bug that may have caused other playback speed issues. It was setting the `userSettingsPlaybackRate` when updated in the app.

I updated the rounding for showing the icons. I'm not sure if there is a reason or you did that intentionally but I changed it so that it rounds to the nearest instead of ceiling. e.g. 2.1f was showing as `x3` but I changed it to show `x2`.

PR #559, opened 2023-02-07
Move Discard Progress to Menu

lkiesow:
This patch moves the discard progress functionality from the tiny icon in the top right corner of the progress display to the options menu.

The reasoning for this is that resetting progress is an option which is unlikely to be used very often and therefor does not need to be on the main screen. In addition to that, the menu already holds the related options to mark the book as finished and to discard the progress of a finished book.

Finally, this removes the tiny icon which does not really match the rest of the user interface of the app.

Screenshot from 2023-02-07 22-41-35

### DIFF
--- a/pages/item/_id/index.vue
+++ b/pages/item/_id/index.vue
@@ -292,6 +292,7 @@ export default {
       return this.$store.getters['user/getToken']
     },
     userItemProgress() {
+      if (this.isPodcast) return null
       if (this.isLocal) return this.$store.getters['globals/getLocalMediaProgressById'](this.libraryItemId)
       return this.$store.getters['user/getUserMediaProgress'](this.libraryItemId)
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks, agreed. We may not even need it if you can just toggle the mark as finished button. This was carried over from the component in the website.

lkiesow:
> We may not even need it if you can just toggle the mark as finished button.

If you want to go in that direction, I would probably always keep “Discard Progress” and remove the “Mark as not finished”. That may even be a bit easier to understand, requires less different options and less thinking on the user's part.

advplyr:
In that case we would just be changing the string. "Mark as not finished" is only shown if you have finished the book. If you have not finished the book it would say "Mark as finished"

PR #560, opened 2023-02-07
App info in account settings

lkiesow:
This patch updates the account settings view, trying to make it look a bit cleaner by:

- Making the logout button look more like a button. The full width looked odd on large devices or in landscape mode.
- Removing the version number. It is already listed in the main menu anyway.
- Moving the hint about reporting bugs and the GitHub link to the bottom.

The hint is still a bit weird in this view. Maybe long-term, we want to move all these information about the app (version, issues, …) to a separate view. But that's something for another day.

Screenshot from 2023-02-07 23-54-27

### DIFF
--- a/pages/account.vue
+++ b/pages/account.vue
@@ -4,11 +4,10 @@
 
     <ui-text-input-with-label :value="username" label="Username" disabled class="my-2" />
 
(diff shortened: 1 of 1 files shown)


advplyr:
Looks great, thanks!

hobesman:
Just make sure the text at the bottom isn't covered by the now playing bar.

advplyr:
Checked that, all good!

PR #562, opened 2023-02-08
Icons for book details menu

lkiesow:
This patch adds icons similar to those used in other app menus to the book details menu.
This is based on pull request #561 and needs to be slightly adjusted if that one is not merged.


advplyr:
Thanks!

PR #563, opened 2023-02-08
Increase padding for menus

lkiesow:
This patch slightly increases the padding for menu dialogs. No real reason for this except that I like the look of this better ;-)

Screenshot from 2023-02-08 22-56-35


advplyr:
There is actually a good reason for this. Some iOS devices cutoff the top part of the screen.
Thanks!

PR #564, opened 2023-02-08
Re-generate lock file

lkiesow:
Running `npm ci` you always get a warning about an old lock-file format being used:

```
npm WARN old lockfile
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile
npm WARN old lockfile This is a one-time fix-up, please be patient...
npm WARN old lockfile
```

This patch simply re-generates the file by removing the old one and running`npm install`. That gets rid of the warning.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,18932 @@
 {
   "name": "audiobookshelf-app",
   "version": "0.9.61-beta",
-  "lockfileVersion": 1,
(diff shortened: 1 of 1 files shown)


advplyr:
Works for me!

Issue #538, opened 2023-02-01
[iOS] External connection over HTTP Failure to Connect


skylarrudolph:
### Steps to reproduce
1. Add server (my-custom-domain).lan:13378
2. Failed to ping server

### Expected behaviour
- Should be prompted with Username/Password for Audiobookshelf application 

### Actual behaviour
- Failed to ping server

### Environment data

Audiobookshelf Version:0.9.61

- [ ] Android App?
- [X] iOS App?

#### Android Issue
Android version:

Device model:

Stock or customized system:

#### iOS Issue
iOS Version: 16.2

iPhone model: 12

gh0sti:
So either your firewall is blocking it or you don't have the correct port forwarding if your phone can't ping your server.

Hexaphim:
I have the same issue after upgrading to the latest version of the iOS app. It worked fine before the upgrade, after I can't connect to the server with a "Failed to ping server" error. The server is accessible in a browser on my phone using the same port.

emmertex:
This issue also appeared on 2 iOS devices which previously worked.
Have downgraded both to last version to allow connectivity. 

If you point it directly to an IP, it works, but we can't do this normally as we access through reverse proxy.
It is not a firewall issue as suggested by @gh0sti

gh0sti:
Does it still happen on iOS 16.3?

Hexaphim:
I upgraded to iOS 16.3 just to test this out, and it does not work on iOS 16.3 either.

emmertex:
I can now confirm the issue is also present on Android

emmertex:
[log] - [Axios] Making request to http://audiobook.url.com/ping
⚡️  [error] - Axios error code {"message":"Network Error","name":"Error","stack":"@capacitor://localhost/_nuxt/22a3f3e.js:2:190849\n@capacitor://localhost/_nuxt/22a3f3e.js:2:189807","config":{"url":"http://audiobook.url.com/ping","method":"get","headers":{"Accept":"application/json, text/plain, */*"},"baseURL":"http://localhost:3000/","transformRequest":[null],"transformResponse":[null],"timeout":3000,"xsrfCookieName":"XSRF-TOKEN","xsrfHeaderName":"X-XSRF-TOKEN","maxContentLength":-1,"maxBodyLength":-1,"transitional":{"silentJSONParsing":true,"forcedJSONParsing":true,"clarifyTimeoutError":false}}}
[...]

PR #568, opened 2023-02-09
Disable ATS for iOS

emmertex:
As discussed on Discord, and tested as far as having Apple approve the bundle.
Just disable ATS.
https://testflight.apple.com/join/5Nf8O3mI for the approved build.
https://github.com/advplyr/audiobookshelf-app/issues/538 for related issue.
https://developer.apple.com/forums/thread/3544 for related response from Apple making it sound this is acceptable.


advplyr:
Thanks!

PR #569, opened 2023-02-09
Blur Cover Image Background

lkiesow:
This patch adds a heavily blurred version of the cover image as background to the cover image area in the book details.
This looks especially nice on larger devices like tablets.

Screenshot from 2023-02-09 23-55-15

Screenshot from 2023-02-10 00-04-21


advplyr:
When using a non-square cover the `BookCover.vue` component is adding a blurred background. I added a prop that disabled the background for that page.

Before:
image

After:
image

advplyr:
Looks great, thanks!

PR #571, opened 2023-02-10
Center title on book details

lkiesow:
This patch centers the title, author and series metadata on the book details page below the cover. This makes this view look similar to the places where the metadata are also places in the center.

Screenshot from 2023-02-10 23-10-55


advplyr:
I like this for the title and subtitle but not as much for the series and authors.

Do you think we should move the series and authors below the buttons. So above narrators would be series, then above series would be authors. What do you think?

advplyr:
Maybe just putting the series above narrators actually. Leaving author centered and removing the "by"

yuuzhan:
I would actually say keep the series centered up top, but move the author down above the narrator.

lkiesow:
Let me just play around with a few variants and we can see how they look like. I'll probably have some time tomorrow to play around

lkiesow:
I've tried moving data around and ended up with a mockup for a somewhat larger change. My basic idea behind this was to not have too many components at the top – that's basically what you were complaining about and I agree – and to have as little different or different looking components as possible. This helps to make the overall look a bit more neat.

- I've moved the author and series down to the metadata table
- All the metadata are now in the table instead of some being presented differently
- I dropped the number of tracks and chapters. That information isn't really necessary/helpful in the app, is it?
- I've moved the progress into the metadata table. I've also removed the percentage.

This is how this could look:

Screenshot from 2023-02-13 00-15-35

And here is a quick demo video:


https://user-images.githubusercontent.com/1008395/218343325-3c7160a5-2b54-4502-b24f-9b672e62ee32.mp4

What do you think? If you like it, then I can turn this into a proper patch. I think it looks cleaner this way and I'm not missing any information, but that said, I did drop some information.

lkiesow:
Another idea could be to move the title down below the buttons. Something like this:

Screenshot from 2023-02-13 00-59-50

yuuzhan:
I think the title has to be above the buttons - or if it does go below have it left justified.  I dont mind the author + series being moved into the metadata table.

I personally like the progress dialog, but that could go below the streaming button.  I also would not be upset if it was a simple green line progress bar without any details to give you an idea at a glance how far through you are.

lkiesow:
Summary from the discussion on Matrix and Discord:

- Keep the progress separated
- Table is fine but don't remove additional info. At least keep the number of chapters
  - Maybe put the less often used stuff in something like “Show more…”
- Just the title above the green button seems lost
  - Either put it below…
  - or keep author there after all
- In case we move the title down (last mockup), don't center title
  - Align it left…
  - or align it with the other content in the second column

These are just the suggestions I collected and no hard requirements.
I'll try creating a few mock-ups based on that.
Feel free to add more ideas here.

lkiesow:
Hey everyone, I played around with the book details page a bit more. I think, I got most of the feedback incorporated into the draft. This needs some cleaning up in the code, but this is essentially, what it looks like on a smartphone and a tablet:

Screenshot from 2023-02-26 12-58-57

A quick tour and explanation of what has changed:

https://user-images.githubusercontent.com/1008395/221409842-855c10c6-5a85-4556-b684-677ee13241f7.mp4

If you want to test it yourself, the APK for Android is available at:
https://github.com/advplyr/audiobookshelf-app/suites/11209564342/artifacts/573398508

Please let me know what you think about this.

advplyr:
Thanks for the video. Seeing the tablet version does have the center alignment make a lot more sense.
I like this version but will look for feedback in the chat also.

yuuzhan:
Love it. Looks great.

The only nitpicky thing I might say is on tablet the large width progress bar and huge streaming button look odd to me.  Not sure id would look better smaller and centered though.

The "gold plated" thing I could think of would be to show 3 metadata items dynamically and "more" the rest.  This would solve the podcast issue of less metadata and potentially some books with lacking metadata.

Not worth holding up a release for either of these very minor things I thought of.

Great work!  Excellent video!

advplyr:
I updated this to move chapters and tracks to a table. Move the size details to the more info modal.
I thought this made the details few enough to show them all.
Description is clamped to 4 lines with a read more button.

image

PR #572, opened 2023-02-10
Improve word wrapping

lkiesow:
This patch improves word wrapping for the description text in book details by justifying the text block and allowing for words to be wrapped by automatically inserting hyphens if necessary.

This causes the description box to look far less ragged on the right edge which I think helps the overall cleanliness of the look.

Unfortunately the app's/browser's aren't as good as e.g. the TeX algorithm for hyphening and there are JavaScript libraries which seem to reimplement that. But this is already a significant step up and far less work for something which is not the main focus of this app.

Screenshot from 2023-02-10 23-21-09

### DIFF
--- a/pages/item/_id/index.vue
+++ b/pages/item/_id/index.vue
@@ -127,7 +127,7 @@
       </div>
 
       <div class="w-full py-4">
-        <p class="text-sm">{{ description }}</p>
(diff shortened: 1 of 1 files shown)


advplyr:
Looks much better, thanks!

PR #574, opened 2023-02-11
Easy way to play current audiobook

### LINKED ISSUE #494: [Usability] Easy way to play current audiobook
If you open the app to start listening to the audiobook you are currently listening to, you first need to click on the audiobook, then scroll to the play/stream button to start playback.

It would be much nicer if the Audiobookshelf app would remember the current audiobook and would automatically load the player for that book in the minimized form at the bottom of the screen.

The currently active audiobook should be that last audiobook being listened __if__ the book's progress is less than 100%. I don't think it would be helpful to load the player for a book if the progress is already 100% and I don't think it is helpful to load the player for older books, even if they aren't finished. I would just consider the last book the user listened to.

This would effectively mean that I can launch the app and simply hit play to continue.

lkiesow:
This patch implements a generic way of re-opening the player for the book you last listened to. This is especially handy when first opening the app and you can restart playback with a single click.

The player is opened only if no other book is already open in the player. It will not overwrite or replace playbacks or player already in progress.

The player is opened paused. No automatic playback since that could be really annoying.

This closes #494

https://user-images.githubusercontent.com/1008395/218283320-511981e5-26ea-4ab2-a74b-ed8aaac743a9.mp4

### DIFF
--- a/components/app/AudioPlayerContainer.vue
+++ b/components/app/AudioPlayerContainer.vue
@@ -188,6 +188,7 @@ export default {
       const libraryItemId = payload.libraryItemId
       const episodeId = payload.episodeId
       const startTime = payload.startTime
+      const startWhenReady = !payload.paused
 
       // When playing local library item and can also play this item from the server
       //   then store the server library item id so it can be used if a cast is made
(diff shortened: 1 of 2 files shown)


advplyr:
I updated this to work for podcast episodes. Also, check if there is a local copy of the episode/book and use that. Lastly, I updated this to only work on the first load. Before if you changed libraries after closing the player then it would auto-open again.

Thanks!

PR #576, opened 2023-02-11
Account Setting Information

lkiesow:
The account setting list the connection name which is the server address directly followed by the username. This is then followed by a field repeating the username again.

This duplication of information is nor really helpful, which is why this patch replaces this with two fields containing just the two information: host address and username.

Screenshot from 2023-02-11 23-46-54


advplyr:
Thanks!

PR #577, opened 2023-02-12
Automatically test and build app

lkiesow:
This patch adds a GitHub Actions workflow which will automatically run through all steps to build the Android version of the app. This is an easy way of catching major build errors in pull requests or pushed commits.

Additionally, this uploads the Android APK as build artifact. This allows anyone to easily download any build for a while. That will make it much easier for non-developers to participate in testing a new version or a specific test since the no longer have to wait for a new release.

The build artifacts are a bit tricky to find, but they are easy to point out if anyone asks.


advplyr:
This is nice. Thanks!

PR #582, opened 2023-02-14
Build APK on Workflow Change

lkiesow:
This patch removes a few restriction about when the Android build workflows run. In particular, we want to run workflows…

- If the workflows themselves change because that could mean that the output changes.
- If the deployment page template changes because that would mean that the published page needs to be updated.

We could still exclude a few specific files if we wanted to, but they should rarely change and it's probably not worth the effort of keeping those lists updated.


advplyr:
That makes sense, thanks!

PR #597, opened 2023-02-22
Android Home Screen Widget

KeyboardHammer:
Implements the rest of the existing partial widget implementation, I did rename the existing files from NewAppWidget to MediaPlayerWidget.

The Widget is 4x1 but can be resized to 5x1
Screenshot2


android/app/src/main/java/com/audiobookshelf/app/player/PlayerNotificationService.kt
```diff
@@ -258,7 +258,7 @@ class PlayerNotificationService : MediaBrowserServiceCompat()  {
     playerNotificationManager.setPriority(NotificationCompat.PRIORITY_MAX)
     playerNotificationManager.setUseFastForwardActionInCompactView(true)
     playerNotificationManager.setUseRewindActionInCompactView(true)
-    playerNotificationManager.setSmallIcon(R.drawable.exo_icon_localaudio)
+    playerNotificationManager.setSmallIcon(R.drawable.icon_monochrome)
```
KeyboardHammer, line 261: Also updated the icon in the notification player since I pulled the abs font icon out to a normal icon for the widget

advplyr:
On my Pixel 6
image

advplyr:
There is one major issue that I've been stuck on with the widget. When the app is completely closed I haven't been able to get the service to start playback.

I recently added to the saved DeviceData object the `lastPlaybackSession` so we can restore the last playback if you press play in the widget while the app is completely closed.

My attempt at implementing this was here https://github.com/advplyr/audiobookshelf-app/blob/master/android/app/src/main/java/com/audiobookshelf/app/player/MediaSessionCallback.kt#L169

This just restarts the last playback session. The problem is the service stops immediately after triggering the `onMediaButtonEvent`.

This UI is great thanks for doing that. Before the widget is available the need issue of using the widget when the app is closed needs to be resolved otherwise we will be bombarded with bug reports.

KeyboardHammer:
> On my Pixel 6 image

I don't understand why the buttons are different sizes, they all have the same css applied

advplyr:
> > There is one major issue that I've been stuck on with the widget. When the app is completely closed I haven't been able to get the service to start playback.
> > I recently added to the saved DeviceData object the `lastPlaybackSession` so we can restore the last playback if you press play in the widget while the app is completely closed.
> > My attempt at implementing this was here https://github.com/advplyr/audiobookshelf-app/blob/master/android/app/src/main/java/com/audiobookshelf/app/player/MediaSessionCallback.kt#L169
> > This just restarts the last playback session. The problem is the service stops immediately after triggering the `onMediaButtonEvent`.
> > This UI is great thanks for doing that. Before the widget is available the issue of using the widget when the app is closed needs to be resolved otherwise we will be bombarded with bug reports.
> 
> It would be nice to have, and given that abs is still running as a background service even after i force close it, it should still work. I'll try to take a stab at it in the next few days.
> 
> I'm not sure there would be too many issues opened though, I just checked the spotify widget and it has the same issue where the buttons at the bottom are useless when the app is force closed. Maybe it's a non issue for most people, I know I rarely force close apps nowadays unlike back in the early days of android.

Maybe we could detect when the app is closed and update the widget to hide the playback buttons?
I tested 2 other audiobook player widgets and they both handle it. I just tested Spotify and they grey out the buttons when the app is closed which I think is also a fine solution if it's easier.

KeyboardHammer:
> > > There is one major issue that I've been stuck on with the widget. When the app is completely closed I haven't been able to get the service to start playback.
> > > I recently added to the saved DeviceData object the `lastPlaybackSession` so we can restore the last playback if you press play in the widget while the app is completely closed.
> > > My attempt at implementing this was here https://github.com/advplyr/audiobookshelf-app/blob/master/android/app/src/main/java/com/audiobookshelf/app/player/MediaSessionCallback.kt#L169
> > > This just restarts the last playback session. The problem is the service stops immediately after triggering the `onMediaButtonEvent`.
> > > This UI is great thanks for doing that. Before the widget is available the issue of using the widget when the app is closed needs to be resolved otherwise we will be bombarded with bug reports.
> > 
> > 
> > It would be nice to have, and given that abs is still running as a background service even after i force close it, it should still work. I'll try to take a stab at it in the next few days.
> > I'm not sure there would be too many issues opened though, I just checked the spotify widget and it has the same issue where the buttons at the bottom are useless when the app is force closed. Maybe it's a non issue for most people, I know I rarely force close apps nowadays unlike back in the early days of android.
> 
> Maybe we could detect when the app is closed and update the widget to hide the playback buttons? I tested 2 other audiobook player widgets and they both handle it. I just tested Spotify and they grey out the buttons when the app is closed which I think is also a fine solution if it's easier.

Yeah that was my first thought, as the Intent to open the app still works after a force close.  If the row of buttons can be hidden then clicking anywhere on the widget would open the app allow the use to start playback.

advplyr:
Ah I updated the widget to hide the action buttons when the app is closed and fixed the icon size for the fast forward. Went to push and have merge conflicts but don't have time to resolve them right now.
My fault, I assumed you weren't working on it today.

advplyr:
On my Pixel 6 the widget was taking up 2 grid spaces in height. This section in the docs helped me make the adjustment https://developer.android.com/develop/ui/views/appwidgets/layouts#anatomy_determining_size

Hopefully that still looks right on your device.

I updated the app closed logic to be contained in the MediaPlayerWidget and set it up to initialize with the last playback session when first added.

This seems to be working well right now. The updated layout looks great for me.
We may modify the design in the future to better support 1.6:1 aspect ratio covers.

Thanks for your work on this, I know a lot of users have been asking for a widget.

KeyboardHammer:
> On my Pixel 6 the widget was taking up 2 grid spaces in height. This section in the docs helped me make the adjustment https://developer.android.com/develop/ui/views/appwidgets/layouts#anatomy_determining_size
> 
> Hopefully that still looks right on your device.
>

The original 272x56 came from the bottom of this section, nice to know the android docs are not 100% accurate: https://developer.android.com/guide/practices/ui_guidelines/widget_design.html#anatomy_determining_size  

> I updated the app closed logic to be contained in the MediaPlayerWidget and set it up to initialize with the last playback session when first added.
> 
> This seems to be working well right now. The updated layout looks great for me. We may modify the design in the future to better support 1.6:1 aspect ratio covers.

Yeah it's doing some cropping in most cases but it's not that noticeable on square covers.  I saw that VLC is doing a lot of work custom scale the cover for their widget, but the 1.6:1 covers still look terrible so I didn't even try to handle those.

> 
> Thanks for your work on this, I know a lot of users have been asking for a widget.

No problem!

PR #649, opened 2023-03-27
Add `android:largeHeap="true"` to `AndroidManifest.xml`

### LINKED ISSUE #227: OOM causes player to fall back from direct stream to transcoding
### Steps to reproduce

1. Get a large fsingle M4B file in to ABS (~1.3GB, though might vary depending on phone model)
2. Open abs-app
3. Start logcat
4. Start streaming
5. Notice it goes from direct -> transcode
6. Notice in logcat that it fails to stream direct due to an OOM, and falls back to transcoding

### Expected behaviour
It should handle the direct playback better, and support the larger file without the need to fallback to transcoding

### Actual behaviour
It falls back to transcoding due to OOMing

### Environment data

Audiobookshelf Version:

- [x] Android App?
- [ ] iOS App?

#### Android Issue
Android version: Android 11

Device model: OnePlus 6

maltejur:
Closes #227

This fix was suggested [here](https://github.com/google/ExoPlayer/issues/7354#issuecomment-627811740) in the ExoPlayer repo and completely solves the issue for me. The `android:largeHeap="true"` line also seems to be in the main ExoPlayer demo code.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -21,7 +21,8 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/AppTheme"
-        android:usesCleartextTraffic="true" >
+        android:usesCleartextTraffic="true"
+        android:largeHeap="true" >
 
         <!-- Android app widget -->
         <receiver


advplyr:
Thanks for tracking that down!

PR #692, opened 2023-04-29
Added Auto Rewind on Sleep Timer setting to Android

### LINKED ISSUE #623: An 'Auto-Rewind' feature under the 'Auto-Sleep' settings
I have a sleep mask I used with bluetooth headphones built into it. Since I wake up often in the middle of the night, I have to restart my playback to fall back asleep with. This involves me:

1. Turning the mask on
2. taking it off
3. opening ABS
4. skipping back 10-15 minutes to find where I fell asleep at
5. putting the mask back on
6. falling asleep again

Ive raised this issue before with #311, but I think the new auto-sleep settings give the perfect opportunity to add a new setting to 'auto rewind' for the user when the auto sleep is triggered. It could be as simple as a toggle to enable 'auto rewind', and a numerical input for the amount of time to rewind. 

This would cut out all the steps of taking off the mask as the user could just:
1. Turn the mask on
2. ABS Starts, Rewinds, Starts a sleep timer all on its own
3. Fall back asleep

The only problem I could see coming up is how it may handle books that finish in the last 'period' of sleep. If my auto timer runs for 20 minutes when it starts, and my book only has 17 minutes left, will ABS be able to handle resuming playback in a book that is completed?

HStep20:
Adds a Setting which will allow the user to Automatically Rewind their position in a book when the Auto Sleep Timer is finished. 
Should aid those who wake up during the night and want to 'backtrack' to an approximate location of where they left off, without turning on their phone screen and fiddling with the position in the book and close #623 

If theres a better way to handle 'rewinding x minutes'  rather than using `playerNotificationService.seekBackward`, I can implement that. 

A possible future feature could be 'rewind to start of current chapter', but I plan to use time as my quantifier for a rewind as opposed to chapter position.


itkfilelor:
Using my VERY limited knowledge in android dev.. but my ability to understand code (sorta 😁) I have a suggestion for handling a book change during the sleep timer. What about storing the current book (name,id,etc not sure how its handled in the app) in a variable or a file and check the current book against it when the timer completes. If it is different, load the stored book and apply the bookmark. There is an intent ``` android.media.action.MEDIA_PLAY_FROM_SEARCH ``` that may prove useful. but again, I am an utter novice with java/kt.

advplyr:
This wasn't working because the auto rewind time wasn't being used in `checkShouldResetSleepTimer()`. 

In general, the coding is the easy part of adding a new feature. The planning, design and testing is the difficult and time consuming part. If you get a chance to contribute again it is very helpful to test the code.

I was able to get this working and additionally save the playback session id so if you change books it will reset the `isFirstAutoSleepTimer` to `true`.

Issue #628, opened 2023-03-12
Rename "Logout" to "Switch Account"


nichwall:
### Is your feature request related to a problem? Please describe.
Some friends and I each have our own ABS instance. This has caused some confusion whenever a mutual friend is added to multiple servers since it hides the ability that the app can be configured for multiple servers or multiple users of the same server (such as in https://github.com/advplyr/audiobookshelf/issues/1601).

### Describe the solution you'd like
Rewording of the "Logout" button in the mobile apps. Maybe this button should also be brought up to 
to the normal menu like in the attached screenshot. Thoughts on the new phrasing:
- Change User
- Change Server
- Change User/Server

![Screenshot_20230312-123333](https://user-images.githubusercontent.com/5686638/224565628-9bdae406-cb3e-4657-9c2b-6c16beb079f2.jpg)

nichwall:
Quote from above PR in case that gets closed/stale.

> One hesitation I have on this is maybe you want to logout and use the App in offline mode. If you have a shaky connection and you just want to listen to your downloaded books maybe you don't want to switch account you just want to logout.
>
> I understand the confusion around not knowing you can have multiple server connections
>


What about:
Change the server picker (where you can see all of the saved connections) to have a little checkmark or "current server"? There could be another "use offline" button below the "add server" button, and the server connection is not removed until that button is pressed or a different server is selected. Then, pressing the back button on the server picker doesn't return the user to "offline mode", since I don't find that intuitive.

I think this would help alleviate the confusion about getting into offline mode too (it's come up in Discord since this issue was opened and I wasn't aware of it before that comment).

The text could be "Offline/Switch User" instead.

advplyr:
I put something together for this next release. The "disconnect" option will just switch to offline mode without redirecting to the server connection picker.
The Switch Server/User button redirects to the server connection picker.

I agree about having a dropdown on the account page with the existing server connections. I would like to support having one server connection with a local ip and an external ip so the UI/UX will have to be revised for that also.


![image](https://github.com/advplyr/audiobookshelf-app/assets/67830747/31552f01-bfc1-420b-bca0-4bcd755455cb)

![image](https://github.com/advplyr/audiobookshelf-app/assets/67830747/52eba2b0-b350-4474-bcb8-e5842cbafdcb)

advplyr:
Updated in [v0.9.66-beta](https://github.com/advplyr/audiobookshelf-app/releases/tag/v0.9.66-beta)

PR #755, opened 2023-06-24
Rename "Logout" button to "Switch account"

bLuka:
Address #628

This makes the multiaccount more explicit, while making it less confusing when wanting to switch account from different Audiobookshelf backends.

Though this does not add a "Switch account" button within the side drawer, this renames the "Logout" button from the "Account" submenu to "Switch account".

IMG_4244

### DIFF
--- a/components/app/SideDrawer.vue
+++ b/components/app/SideDrawer.vue
@@ -25,7 +25,7 @@
           <p class="text-xs">{{ $config.version }}</p>
           <div class="flex-grow" />
           <div v-if="user" class="flex items-center" @click="logout">
-            <p class="text-xs pr-2">Logout</p>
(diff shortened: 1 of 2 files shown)


advplyr:
One hesitation I have on this is maybe you want to logout and use the App in offline mode. If you have a shaky connection and you just want to listen to your downloaded books maybe you don't want to switch account you just want to logout.

I understand the confusion around not knowing you can have multiple server connections

nichwall:
I just responded to the above comment in the original issue (not sure if the conversation should continue here or there)
#628

advplyr:
I pushed up a solution for this. See comment https://github.com/advplyr/audiobookshelf-app/issues/628#issuecomment-1722501514

PR #813, opened 2023-07-26
Add missing space for authors as well #777

### LINKED ISSUE #777: [Bug] Incorrect spacing around commas
As seen from the screenshot above, the comma delimiter between narrators and tags uses incorrect spacing.

It's supposed to be item 1, item 2, item 3 and not item 1 ,item 2 ,item 3

Not a big problem, just an aesthetics issue, but still reporting.

casrou:
Seems that the space still is missing before comma for the 'Authors' field. I believe it ought to have been a part of https://github.com/advplyr/audiobookshelf-app/commit/dd91bc666792f9ebaea6de2bb9c272cb38f0cd58 that fixes #777 🙂 

Also cleans up an inconsistency in what was used as `key` for the separator-`span`.

### DIFF
--- a/pages/item/_id/index.vue
+++ b/pages/item/_id/index.vue
@@ -91,7 +91,7 @@
         <div v-if="narrators && narrators.length" class="truncate text-sm">
           <template v-for="(narrator, index) in narrators">
             <nuxt-link :key="narrator" :to="`/bookshelf/library?filter=narrators.${$encode(narrator)}`" class="underline">{{ narrator }}</nuxt-link
-            ><span :key="${narrator.id}-comma" v-if="index < narrators.length - 1">,&nbsp;</span>
+            ><span :key="index" v-if="index < narrators.length - 1">,&nbsp;</span>
           </template>
         </div>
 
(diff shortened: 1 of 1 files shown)


pages/item/_id/index.vue
```diff
@@ -91,15 +91,15 @@
         <div v-if="narrators && narrators.length" class="truncate text-sm">
           <template v-for="(narrator, index) in narrators">
             <nuxt-link :key="narrator" :to="`/bookshelf/library?filter=narrators.${$encode(narrator)}`" class="underline">{{ narrator }}</nuxt-link
-            ><span :key="index" v-if="index < narrators.length - 1">,&nbsp;</span>
+            ><span :key="${narrator.id}-comma" v-if="index < narrators.length - 1">,&nbsp;</span>
```
advplyr, line 94: The reason `index` is used here is because `narrator` is a string. So `narrator.id` doesn't exist. It is the same for genres.
genres & narrators are just arrays of strings

casrou: Ah, my bad!

advplyr:
Thanks!

PR #851, opened 2023-08-30
Fix separator between authors

lkiesow:
This patch fixes a small problem with the separator between several authors which is `Author A , Author B`, but should be instead `Author A, Author B` like with the other metadata fields.


advplyr:
Thanks!

PR #852, opened 2023-08-30
Don't use non-breaking spaces

lkiesow:
This patch converts the non-breaking spaces in the metadata table to breaking spaces to avoid awkward wrapping. For example, due to the `&nbsp;` which is part of the author separator, you might end up with wrapping like this:

    Author    |
    A, Author |
    B         |

instead of

    Author A, |
    Author B  |


advplyr:
Thanks, I just updated this in a commit since it was easier then merging https://github.com/advplyr/audiobookshelf-app/commit/2366c938b56c7b738d5afaf8c290cd236e656dac

PR #906, opened 2023-10-08
OAuth2 Support

Sapd:
This PR Provides oauth2 support

It works out-of-the-box with the current changes in https://github.com/advplyr/audiobookshelf/pull/1636

* Fully works on iOS and Android
* ~~Should work on Android, too. But needs testing (I dont have an dev env for that)~~ (Testing done successfully)
* It should - in theory - also work with Google and co, but needs testing. If there is a error I would check if the callback URLs are correct as first step

It works this way:
1. oauthRequest
    * API request to `audio.example.com/auth/openid`
    * This will try to redirect to the idp/authentik. Instead of redirecting Im catching it and open this link in a browser. Also Im patching the redirect_uri parameter to `&redirect_uri=audiobookshelf://oauth` 
    * Also what its very important (and required by passport - oauth only requires it as GET parameter instead in a Session), it will generate a state variable and also save it into a Cookie. This Cookie needs to be sent in the request later to `audio.example.com/auth/openid/callback`. Saving the Cookie is automatically done with `CapacitorHttp.get` as it uses a native function, it won't work in any case with `fetch` or `axios`
2. Browser.Open
    * This is the redirect to Authentik, sending also the openid parameters and additionally the same state variable from above. The user logs in. In any case authentik will then redirect to audiobookshelf://oauth and pass along a code and state parameter via GET parameters
    * Cookies are not required to be saved or passed (oauth works completely without session but just GET parameters)
3. audiobookshelf://oauth
    * This link will send a message to the app including state and code variable generated by Authentik
4. oauthExchangeCodeForToken
    * This calls `audio.example.com/auth/openid/callback` with the state and the code (and the cookie from the first request bc of passport)
    * Browser is closed
    * Token is returned and then handled like with user/pass login

Error handling is also done:
* this.showAuth = true is not required to be set on error I noticed
* My error handling might not be consistent with the existing code and the toasts might need translation
* The error handling in the backend is currently not sufficient. On a state error the message will be just forbidden without any logs (will post details in the other PR). Can be important when you test google integration or so

I used `App.addListener` on top in `ServerConnectForm.vue`. I'm not really familiar with Vue so I saved the `this` pointer of the component simply in the variable `serverConnectForm`. Maybe there is a better way. Also in the function I catched for other not implemented URLs, given that the function can be called multiple times this catch might not make sense at this point (so probably either  the `else` should be removed or the function should be centrally moved somewhere else in a routing file).

### DIFF
--- a/android/app/capacitor.build.gradle
+++ b/android/app/capacitor.build.gradle
@@ -9,8 +9,8 @@ android {
 
 apply from: "../capacitor-cordova-android-plugins/cordova.variables.gradle"
 dependencies {
-    implementation project(':byteowls-capacitor-oauth2')
     implementation project(':capacitor-app')
+    implementation project(':capacitor-browser')
     implementation project(':capacitor-clipboard')
     implementation project(':capacitor-dialog')
     implementation project(':capacitor-haptics')
--- a/android/app/src/debug/res/values/strings.xml
+++ b/android/app/src/debug/res/values/strings.xml
@@ -3,7 +3,7 @@
     <string name="app_name">audiobookshelf</string>
     <string name="title_activity_main">audiobookshelf</string>
     <string name="package_name">com.audiobookshelf.app</string>
-    <string name="custom_url_scheme">com.audiobookshelf.app.debug</string>
+    <string name="custom_url_scheme">audiobookshelf</string>
     <string name="add_widget">Add widget</string>
     <string name="app_widget_description">Simple widget for audiobookshelf playback</string>
     <string name="action_jump_forward">Jump Forward</string>
--- a/android/app/src/main/assets/capacitor.plugins.json
+++ b/android/app/src/main/assets/capacitor.plugins.json
@@ -1,11 +1,11 @@
 [
 	{
-		"pkg": "@byteowls/capacitor-oauth2",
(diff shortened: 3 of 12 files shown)


Sapd:
Now also adapted and successfully tested on Android too

Sapd:
Made some security improvements for the oauth2 flow. HTTPS is now also forced for it (requirement by oauth2 RFC)

---

Also because https is a requirement for that, I reworked the connection flow a bit. 
If the user does not type in a protocol, like `audiobooks.example.com` it will try HTTPS first. If that does not work, because there is no HTTPS server listening (then it is safe to do so), it will also try HTTP.
On top I did multiple validations of the resulting connection url and unexpected or unsafe redirects. The error messages the user gets were also improved by far.
There was also a bug, where the server would redirect from HTTP to HTTPS, but audiobookshelf would just save the http address, which was fixed.

I tested it on iOS, android and with an ABS behind https and one behind plain http and ofc with wrong URLs too.

---

What now is still missing is I need to introduce a PKCE flow. This is required by oauth2 for mobile apps, because otherwise a malicious app could also register the in-app-link `audiobookshelf://` and intercept the code. I will wait for that until we have evaluated node-openid-client and the branch ready for that/

advplyr:
Thanks!!

Issue #966, opened 2023-11-30
[Enhancement]: Add a low feedback / notificaition setting


tonyedwardspz:
### Is your feature request related to a problem? Please describe.

I love this software. I've built an offline first content consumption pattern around it, and it was the starting point for my meandering through the world of self-hosted open source.

However I do have one gripe. 

It's always trying to tell me stuff. Lot's of stuff. It's great hearing about how awesome ABS is at getting things perfect every time. Well. Almost every time :)

For example, the toasts and confirmation dialogs can be a bit much when downloading lots of short podcast episodes in quick succession. The downloaded media link status is irrelevant, personally speaking. I can completely see why these things are in the app and are hugely useful to lot's of people.

Kinda an extension of #718.

### Describe the solution you'd like

Introduce a low feedback mode. 

After toggling the mode on through the "user interface settings" section of the settings menu, the following would happen:

- Download confirmation dialog for podcasts are skipped
- Download success toasts for podcast are disabled
- Ditto both the above for audiobooks
- Downloaded media link status on podcast pages are hidden

A [pull request](https://github.com/advplyr/audiobookshelf-app/pull/967) will be made shortly for a starter implementation of the above for Android. 

I can think of other things to add to the list, but wanted to get some input from others before going any further.

After using a more hacky version of the implementation (just delete all the things) for a few days, I'd describe my experience of using the app as calmer.

### Describe alternatives you've considered

Maintaining my own fork and getting stuck on an outdated version of everything at some point in the future.

I mean. I probably won't tbf.

### Additional context

In some ways, this is a great accessibility feature to have in the app.  Kinda related reading - [https://www.smashingmagazine.com/2020/09/design-reduced-motion-sensitivities/](https://www.smashingmagazine.com/2020/09/design-reduced-motion-sensitivities/)

advplyr:
Added in [0.9.69-beta](https://github.com/advplyr/audiobookshelf-app/releases/tag/v0.9.69-beta)

PR #967, opened 2023-11-30
Feature - Add a low feedback mode

tonyedwardspz:
This pull requests implements the functionality mentioned in #966 

After toggling the mode on through the "user interface settings" section of the settings menu, the following has been changed:

- Download confirmation dialog for podcasts are skipped
- Download success toasts for podcast are disabled
- Ditto both the above for audiobooks
- Downloaded media link status on podcast pages are hidden

I'm not set on any of this and expect it to change. The word choices can probably be vastly improved.

Happy to take feedback to shape this feature. If it expands much beyond a few simple prompts and dialogs, it might be wise to group things under a couple of settings toggles.

Low feedback setting tiggle

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/data/DeviceClasses.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/data/DeviceClasses.kt
@@ -122,7 +122,8 @@ data class DeviceSettings(
   var sleepTimerLength: Long, // Time in milliseconds
   var disableSleepTimerFadeOut: Boolean,
   var disableSleepTimerResetFeedback: Boolean,
-  var languageCode: String
+  var languageCode: String,
+  var enableLowFeedbackMode: Boolean
 ) {
   companion object {
     // Static method to get default device settings
@@ -145,7 +146,7 @@ data class DeviceSettings(
         autoSleepTimerAutoRewindTime = 300000L, // 5 minutes
         disableSleepTimerFadeOut = false,
         disableSleepTimerResetFeedback = false,
-        languageCode = "en-us"
+        enableLowFeedbackMode = false
       )
     }
   }
--- a/android/app/src/main/java/com/audiobookshelf/app/data/DeviceClasses.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/data/DeviceClasses.kt
@@ -144,7 +144,8 @@ data class DeviceSettings(
         autoSleepTimerAutoRewind = false,
         autoSleepTimerAutoRewindTime = 300000L, // 5 minutes
         disableSleepTimerFadeOut = false,
-        disableSleepTimerResetFeedback = false
+        disableSleepTimerResetFeedback = false,
+        languageCode = "en-us"
       )
     }
   }
--- a/android/app/src/main/java/com/audiobookshelf/app/data/DeviceClasses.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/data/DeviceClasses.kt
@@ -122,8 +122,7 @@ data class DeviceSettings(
   var sleepTimerLength: Long, // Time in milliseconds
   var disableSleepTimerFadeOut: Boolean,
   var disableSleepTimerResetFeedback: Boolean,
-  var languageCode: String,
-  var enableLowFeedbackMode: Boolean
+  var languageCode: String
 ) {
   companion object {
     // Static method to get default device settings
@@ -145,8 +144,7 @@ data class DeviceSettings(
         autoSleepTimerAutoRewind = false,
         autoSleepTimerAutoRewindTime = 300000L, // 5 minutes
         disableSleepTimerFadeOut = false,
-        disableSleepTimerResetFeedback = false,
-        enableLowFeedbackMode = false
+        disableSleepTimerResetFeedback = false
       )
     }
   }
--- a/components/modals/ItemMoreMenuModal.vue
+++ b/components/modals/ItemMoreMenuModal.vue
@@ -388,7 +388,9 @@ export default {
       if (value) {
         const res = await AbsFileSystem.deleteTrackFromItem({ id: this.localLibraryItemId, trackLocalFileId: localFile.id, trackContentUrl: localEpisodeAudioTrack.contentUrl })
         if (res?.id) {
-          this.$toast.success('Deleted episode successfully')
+
+          //this.$toast.success('Deleted episode successfully')
+
           if (this.isLocal) {
             // If this is local episode then redirect to server episode when available
             if (this.serverEpisodeId) {
@@ -414,7 +416,9 @@ export default {
       if (value) {
         const res = await AbsFileSystem.deleteItem(this.localLibraryItem)
         if (res?.success) {
-          this.$toast.success('Deleted successfully')
+
+          //this.$toast.success('Deleted successfully')
+
           if (this.isLocal) {
             // If local then redirect to server version when available
             if (this.serverLibraryItemId) {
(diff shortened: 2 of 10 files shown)


Grog02:
Brilliant idea! I would consider using a word other than feedback for the low feedback mode as it can be confusing having the same word twice but having it mean separate things for some users. Other than that I love the idea of being able to opt out of being shown details on absolutely everything that is going on!

nichwall:
What if instead of adding checks for every toast, a wrapper function is written for the toast `success` function? This wrapper would accept a second argument for `isLowFeedback` and handle that if statement.

That would reduce the number of checks and just be a global option for success toasts. I think the error toasts should not be disabled in Low Feedback Mode yet, but could be in the future.

This could also be extended to groups, since the wrapper toast function would just accept the relevant setting (for the toast) instead of a global Low Feedback Mode.

nichwall:
On that note, maybe the "Low Feedback Mode" should just be "Enable Success Messages", and defaults to Enabled.

It should not be called "Disable [item]" because that Enable/Disable caused a lot of confusion with other settings in the past.

tonyedwardspz:
Good spot @Grog02 

Good idea in the wrapper @nichwall. Likewise for the error messages, I wanna know when things go wrong and assume they go right.

Agree with avoiding disable wording. "Enable Success Messages" makes sense if we sperate that functionallity out as a seperrate setting. This implementation does slightly more that disable success messages as is.

nichwall:
My guess would be `plugins`

advplyr:
I actually think it's fine if we remove the download success toasts. The pattern so far has been I overuse the toasts in the beginning and they slowly get trimmed back.
Are there other toast messages you would like to see hidden? Maybe those are fine to also remove

I'm not sure about hiding the alert messages about local media not being syncable to the server. I can foresee debugging with someone in the future about a sync issue come to find out they had that hidden. Maybe we don't show the successfully linked alert box anymore and when not connected to a server we also don't show anything.

nichwall:
> I'm not sure about hiding the alert messages about local media not being syncable to the server. I can foresee debugging with someone in the future about a sync issue come to find out they had that hidden. Maybe we don't show the successfully linked alert box anymore and when not connected to a server we also don't show anything.

I agree with never hiding the connected to a different server/user message since people are still confused by how multiple servers work and the website guides are not translated at all (the apps supporting localization will help with this confusion). It also provides an easy way to remember which of the several servers/users you used to download media if that's applicable, and if someone is only using one server/user then they'll just never see it anyway.

I think keeping the "Linked to server [server]" should remain in offline mode but could be hidden by the setting. Or, maybe only show this message if there's more than one server connection?

I think the "connected to the correct server" could be removed entirely, though. It's nice to have but I agree it's not necessary.

tonyedwardspz:
> Are there other toast messages you would like to see hidden? Maybe those are fine to also remove 

The others were linked to server sync status, but I can't think when they pop up cause I've not seen them for a few days.

After thinking, I was leanig towards two toggles:

- Hide success toasts
- Hide media link status

The latter would hide the main server link message on the podcast page, along with related toasts. 

I'm keen to hide the main link media message completely. It's positioning and coloring presents itself as the most important piece of info on the page, but it's not for many users.

I do like the suggestions tho. Removing it when possible would help, and hiding it under certain conditions (offline) would be good. If we're going to not allow hiding, can the status be better integrated into the UI? What about condensing the message into another line under Author / Genre? The info is still present, but visually deprioritised.

Screenshot 2023-12-05 at 06 19 41

Even with moving the info, I'd still push for it to be present as little as possible, particuarly when everything's ok / app is in disconnected state as you both discuss. It'd still be good to hide it totally tho.

nichwall:
I agree that presenting the information in a less visible way would be nice. I'm not sure when you joined the project, but this is how it used to look. The new messages were added in `0.9.67`. (I covered up my URL, but it was in the same color of white as the other text)





The issue was people kept getting confused and posting in Discord and GH about how sync was broken, it doesn't work, etc, but they were listening without connecting to the server so it couldn't ever sync back. Then, they'd listen on another device, and then their progress would get messed up. These new messages were an attempt to make it more visible that they were not connected, but also to provide information to users who don't understand the functionality.

I think it still needs to be more prominent. Maybe the setting could toggle between the big message with the warning and the little message without a warning, and have it on the big message by default? I don't want people just turning it to low priority things and then getting confused why things don't work.

advplyr:
I think the media link alerts can be addressed without requiring a toggle setting. The download success toasts can be removed also. Any other unnecessary toasts could be removed also if they are found.

The only thing a setting may be useful for is the confirmation that comes up when clicking download. For podcast episodes we may not need the confirmation since the files are smaller. Audiobooks should keep the confirmation on download.

Since your initial request is related to podcast episode downloads I wonder if a setting is necessary at all if the above is updated.

tonyedwardspz:
This sounds like a good evolution of the UI taking in the discussion 👍. Happy to move my branch towards this and give it a go.

@nichwall I've been around for a few versions of the Android app. I remember the previous version of the sync notification, but never totally understood what it was tbh. As I'm not fussed about stats, it was never a problem regardless of what it said. It was something I could safely ignore rather than have to think about every time I open a podcast.

I totally get the frustration of repeated questions in discord. I've lived that admin life. It's quite common and one of the trade-offs of using a chat platfrom as a knowledge repository mixed with a community hub. It biases towards chat rather than retreival to solve problems.

Through this discussion, I've realised that my OP could have been more focused around syncing. A toggle to say stats don't matter, I don't want to know when sycncing goes wrong. Framing a toggle around that would likely avoid confusion about what the toggle does. That would be sperate to this tho.

tonyedwardspz:
Based on our conversation (thanks for taking the time to have it btw), my branch currently:

Toasts / Prompts
- [x] The download success toasts have been removed
- [x] Podcast downloads don't create a confirmation dialog
- [x] Audiobook download confirmation dialog remains

Media link status
- [x] Removed the successfully connected media link messages
- [x] Hides the media link status if not connected to the server

This setup seems to work for me with the limited use so far.

I've also commented a few toasts related to succesfully deleting things. These would be my suggestions for the moment. 

Others are tightly coupled to media sync status and are warning/error messsages. It only makes sense to remove these for users who explicilty choose to ignore the stats part of the app ( and therefore won't end up is discord asking for help with syncing ). That would suggest a toggle is needed, but is outside of the scope of this pull imo.

nichwall:
Based on a comment from Sapd in Discord, what if instead the media link info box changes to a popup and was formatted like before?

Instead of showing a green "connected" or the white IP, instead always show the IP/server connection, with a green "Connected" or red "Not Connected". There would then be an info bubble next to the connection state that users can tap on to get the messages currently displayed as a popup. Those could be color coded if desired, but I don't think the color coding would be necessary.

That would still provide information and make it less intrusive. The info bubble would also make it more clear that this is useful sync information versus a string without a label.

advplyr:
I realized while testing this that the latest page was never fully setup with local downloads. I'll have to fix this before merging because without the toast there is no indication that the episode was downloaded.

advplyr:
This is good to go now. We will only show an alert now if the user is connected to a server and it is not the same server as the downloaded book they are viewing. We can make further adjustments if we need to with more feedback.

If we can continue to remove unnecessary toasts and alerts we definitely should, so hopefully we don't need a setting for this. The less settings and noise the more intuitive the app should be hopefully.

Thanks!

tonyedwardspz:
Brilliant. Thank you for accepting this. I'm really pleased that my first ever open source pull is for ABS 🙌

I really appreciate everyone's time in helping me put this together.

PR #969, opened 2023-12-04
SSO: Add redirect_uri to oAuth auth request

Sapd:
Also:
- Add client_id with the name of the app (for possible future use)
- Add response_type simply to provide all required oauth2 parameters for reference
- For the URL to be opened in Browser: Forward the callback/redirect_uri URL provided by the server
- Use a hardcoded variable for HTTPS enforcement, for easier debugging

### DIFF
--- a/components/connection/ServerConnectForm.vue
+++ b/components/connection/ServerConnectForm.vue
@@ -107,7 +107,8 @@ export default {
         state: null,
         verifier: null,
         challenge: null,
-        buttonText: 'Login with OpenID'
+        buttonText: 'Login with OpenID',
+        enforceHTTPs: true // RFC 6749, Section 10.9 requires https
       }
     }
   },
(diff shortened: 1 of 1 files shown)


advplyr:
Can we make this backwards compatible with server v2.6.0 so that the app can be released before the server? Or we can release the server first and make that compatible with app v0.9.68.

Sapd:
> Can we make this backwards compatible with server v2.6.0 so that the app can be released before the server? Or we can release the server first and make that compatible with app v0.9.68.

@advplyr 
* If the server is released first, it will expect a redirect_uri (which comes with this change)
* If the client is released first, the server will expect a is_rest (which requirement will be removed with the server change)
* Also I think a problem could be the change around line 201 which now takes the redirect_uri of the server, making it not compatible with older server versions (I think, bc in the old version it probably sent a link to the ABS callback url instead of the app one)

In theory one could postpone the isRest parameter (I removed it to show that the app just works with standard oauth parameters on the new server version). But the last point with the callback URL is the problem. If possible I would recommend releasing them at nearly the same time.

advplyr:
I think that'll do it

PR #989, opened 2023-12-15
Update it.json

burghy86:
first translation, I reserve the right to fix in future releases as soon as I notice errors. thank you very much indeed


advplyr:
I'm not sure what you mean. Should I wait to merge this or are you still working on it? You can open a PR as a draft if you are working on it.

burghy86:
I made a mess, with patch1 I had translated the entire file, I thought I had done the pool but I saw that it hadn't been accepted yet, you can see I did something wrong.  so I started translating it again but I found patch 1 and sent it to the pool.  this is the correct one.  sorry for the mess

advplyr:
This PR does not match `en-us.json` it is missing a bunch of strings.

The file should have 283 strings

https://github.com/advplyr/audiobookshelf-app/blob/master/strings/en-us.json

PR #372, opened 2022-09-07
Display current chapter in the iOS Now Playing widget instead of the overall book progress

ronaldheft:
* Implements #239 for iOS
* Right now, if a book has chapters it will display the chapter information, but I plan to tie this to "Chapter Track" setting on the player

PR #1002, opened 2023-12-23
iOS Now Playing Chapter Track (updated #372)

benonymity:
This is all of @ronaldheft's work from #372 tied into the chapter track setting of the webView player. The only thing I'm unsure about is whether the setChapterTrack route for AbsAudioPlayer needs to be implemented on the Android side as well. I don't have an Android to test whether the route being called causes any errors, but it compiles and works perfectly on iOS.


benonymity:
Fixes #239

advplyr:
This is working great in my tests. I had to make one fix when using the track in NowPlaying to seek.

The only other part I'd like to improve on that is there is sometimes a 5 - 10 seconds delay when switching chapters before the NowPlaying is refreshed.

Thanks!

PR #1003, opened 2023-12-23
Disable iOS swipeBack in fullscreen audio player

### LINKED ISSUE #681: [Enhancement/RFC] iOS screen swipe back should be disabled on "now playing" screen
### Issue overview
When I am on the "now playing" screen, a swipe right from the left edge of the screen (generally indicating "go back" in iOS UX) causes unexpected behavior. 

https://user-images.githubusercontent.com/45179933/233757822-f9a7b6dc-ad26-4ff1-abf5-9dd7fae702ca.MOV

These are the aspects that make it feel especially unexpected from a UX standpoint:

- As a user of other apps in the iOS ecosystem, I don't expect the "now playing" screen to even accept a "go back" input because (A) there is a downward pointing arrow at the top left, meaning that this tile slides up and down, not side to side -- and because (B) the "now playing" screen feels like it operates "on top of" or "in a higher priority" than the other screens of the app. As a user, I expect it to not interact with the history of my previous windows. 
- It seems like the app is processing the back swipe the same as the browser would (which makes sense given the codebase). This sometimes feels glitchy. For example in the video above, I'm on my collections page, visit authors, bring up the now playing screen, but when I swipe back, I'm taken to the collections page instead of the screen that I was just on. This does not make sense with iOS UX.  

### Proposed solution
I feel like the best and easiest solution is to disable back-swipes on the now playing screen entirely. That way the user only interacts with it through vertical swipes or taps. 

I'd love to hear thoughts on this proposal, or any suggestions others have. 

Thanks for an amazing project! Use this every day.

benonymity:
Simple touchstart intercept when swiping within 20 pixels of the edge of the screen (does not interfere with scrubbing, skipping, etc.) Fixes #681


advplyr:
I've been testing this and this seems like a partial fix. I can still navigate back sometimes and can't see why. Even if the check for < 20 is removed and applied to all events it still can navigate backwards sometimes.

benonymity:
Really? I didn't experience inconsistency in my testing, but I'll try some more/on a different device and see if I can reproduce that.

advplyr:
One thing to note is if there is no page to go back to then it doesn't respond. So you need to navigate each time you try it or navigate a bunch first

components/app/AudioPlayer.vue
```diff
@@ -856,7 +860,7 @@ export default {
     window.addEventListener('resize', this.screenOrientationChange)
 
     this.$eventBus.$on('minimize-player', this.minimizePlayerEvt)
-    document.body.addEventListener('touchstart', this.touchstart)
+    document.body.addEventListener('touchstart', this.touchstart, {passive: false})
```
advplyr, line 863:
```suggestion
    document.body.addEventListener('touchstart', this.touchstart, { passive: false })
```

advplyr:
This does prevent the majority of touches so I want to include this in the release today. It would be nice to figure out how to disable that entirely.

advplyr:
Thanks!

KaiStarkk:
Nice work @benonymity . Particularly helpful when trying to change playback position early in a book.

PR #1004, opened 2023-12-23
iOS Status Bar Tweaks

benonymity:
Forces the iOS status bar (aka the body background color) to mirror the current theme, and hopefully to mirror the color gradient in the fullscreen audio player, ~~though haven't quite figured that out yet.~~ which it now does. So the iOS cutout bar is now consistent with the rest of the UI, though again can't confirm the lack of Android conflicts.


advplyr:
This PR removed the gradient from the item page but I had been wanting to improve that gradient for a while so I redid that. It used to be fixed and would scroll with the page which didn't look good imo so it is now absolute position under the cover image on that page.

For Android this doesn't change the status bar color. That may be possible using the Status bar capacitor plugin but that is a separate PR.

Thanks!

Issue #622, opened 2023-03-04
[iOS] Can't pause from headphones


KaiStarkk:
Pressing the pause button causes the app to rewind a few seconds. It seems to register as two presses (pause then play again).

Tested with Aftershokz Aeropex and AcousticSheep sleepphones, both have same problem. 

Sony wh1000mx3 works correctly. 

~iOS 16.3.1~
~iOS 17.1.1~
iOS 17.2.1
~App version 0.9.66~
~App version 0.9.67~
App version 0.9.68
~Server version 2.5.0~
Server version 2.6.0

Please thumbs up if you have this issue, appears to be preferred way of prioritizing issues.

cmccambridge:
I’m also experiencing this in the iOS app with my soundpeats q30 Bluetooth headset. I have to get my phone out to pause a book, which hurts a lot of my favorite audiobook cases- working out, doing chores, etc.

App: 0.9.63-beta
Backend: 2.2.19

advplyr:
I'm not very familiar with iOS since I use Android but on Android it is similar to how you are describing.

KaiStarkk:
> I'm not very familiar with iOS since I use Android but on Android it is similar to how you are describing.

To clarify, on iPhone it is different from android - when you press pause, it rewinds and continued playing. As though the button was pressed twice. 

Is it possible to implement the fix suggested above by handling togglePlayPauseCommand?

wnoakes:
I see this with my iPhone 13 mini on 17.0.3 using Ultimate Ears Boom 3 and Creative Outlier ONE headphones

KaiStarkk:
@ronaldheft  hoping you might be able to share some wisdom on this one? Appears you’ve had some involvement in the iOS player handlers

ronaldheft:
@KaiStarkk Are you experiencing this with a book that is multi-track (multiple files) or single-track (one file)?

I ask, because I know there are definitely still issues with multi-track in the player, especially around winding back the book a few seconds when resuming.

KaiStarkk:
@ronaldheft just tested, it’s happening with both multi track and single track. 

Essentially it’s impossible to pause from headphones.

Any thoughts on the suggestion above about handling  https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/1618992-toggleplaypausecommand

KaiStarkk:
On review it would make sense to do a sanity check when pausing and playing also. i.e.

```
[...]

PR #1005, opened 2023-12-23
Add togglePlayPauseCommand route to iOS

benonymity:
This is just the code from the comments of #622—credit to @KaiStarkk! I ran it on my device and it seems to be working normally, though I don't have old enough headphones to test the actual toggle route. Though if you sideload the APK this PR generates, you might be able to do so. Should, in all theory, work.

### DIFF
--- a/ios/App/Shared/player/AudioPlayer.swift
+++ b/ios/App/Shared/player/AudioPlayer.swift
@@ -583,12 +583,31 @@ class AudioPlayer: NSObject {
         
         commandCenter.playCommand.isEnabled = true
         commandCenter.playCommand.addTarget { [weak self] event in
-            self?.play(allowSeekBack: true)
+            if (self!.isPlaying()) {
+                self?.pause()
+            } else {
+                self?.play(allowSeekBack: true)
+            }
             return .success
         }
+
         commandCenter.pauseCommand.isEnabled = true
         commandCenter.pauseCommand.addTarget { [weak self] event in
-            self?.pause()
+            if (self!.isPlaying()) {
+                self?.pause()
+            } else {
+                self?.play(allowSeekBack: true)
+            }
+            return .success
+        }
+
+        commandCenter.togglePlayPauseCommand.isEnabled = true
+        commandCenter.togglePlayPauseCommand.addTarget { [weak self] event in
+            if (self!.isPlaying()) {
+                self?.pause()
+            } else {
+                self?.play(allowSeekBack: true)
+            }
             return .success
         }
         


KaiStarkk:
Thanks, it was @cmccambridge not me though. Also just noting that it’s only an iOS bug, RE the APK note. 

Sounds like you haven’t tested that this doesn’t break playback on Android or iOS right? I think we just need testing from someone who can compile from Xcode to a physical iPhone to check that everything still works as normal. @johnpyp on discord was looking into it a few days ago. 

Failing that it’s on my todo list to get a macOS VM going if possible these days.

benonymity:
Oh gotcha, forgot APKs were Android only 🤦‍♂️

Yeah, I compiled it to my physical iPhone with Xcode and playback worked fine, and play/pause with AirPods was also successful, though again I don't have headphones old enough to actually test the toggle command

johnpyp:
Yeah I managed to get it running on my iPhone, but I don't have any headphones to test. I think we should just add the command and publish it.

KaiStarkk:
If anyone is able to compile the IPA and send me a binary, I'll figure out sideloading and test if it fixes the bug.

KaiStarkk:
Can I suggest you just include lines 604-610 in initial testing? Can look at the play + pause handlers afterwords if it doesn't fix the bug.

KaiStarkk:
In hindsight I would propose we keep the change minimal to confirm what is being called when the bug occurs. Also fixed a couple logical issues with returned codes

```
        commandCenter.playCommand.isEnabled = true
        commandCenter.playCommand.addTarget { [weak self] event in
            self?.play(allowSeekBack: true)
            return .success
        }

        commandCenter.pauseCommand.isEnabled = true
        commandCenter.pauseCommand.addTarget { [weak self] event in
            self?.pause()
            return .success
        }

        commandCenter.togglePlayPauseCommand.isEnabled = true
        commandCenter.togglePlayPauseCommand.addTarget { [weak self] event in
            if (self?.isPlaying()) {
                self?.pause()
	            return .success
            } else {
                self?.play(allowSeekBack: true)
    	        return .success
            }
            return .commandFailed
        }
```

KaiStarkk:
@advplyr @johnpyp @benonymity see video attached showing a locally built version vs. the TestFlight version. The bug appears to be fixed.

https://github.com/advplyr/audiobookshelf-app/assets/1722064/87f4c7a8-a739-4532-b81b-6b0e2f444e90

Here's the confusing part though: this is before merging the PR.
*i.e. the working version shown in this video is the current build before applying the fix*.

Any ideas? I'm going to use the local version for a few days to see if the bug reappears.

KaiStarkk:
Scratch that, I'm able to reproduce. It only occurs when playing at faster than 1x speed.
Will try building the fix tonight and see if this resolves the problem.

KaiStarkk:
Handling `togglePlayPauseCommand()` doesn't appear to fix the bug.
I also noticed that changing the playback speed causes a play event to occur, as shown in this screencap.
https://github.com/advplyr/audiobookshelf-app/assets/1722064/1a6fb6c2-0066-44ff-a91e-49246d375a77

With that said, I've made the modifications to `playCommand `and `pauseCommand `above as well, and this appears to fix the issue! The PR can I think be merged as-is @advplyr, though @benonymity it might be worth a rebase back to master (not sure if required?)

cmccambridge:
> It only occurs when playing at faster than 1x speed.

Ahh, great catch! I can definitely confirm that my instances of this bug occurred while playing at greater than 1x speed.

I’ve re-confirmed the following just now, and am happy to retest with a new test flight or by sideloading a build (if that’s possible without a Mac?):

- App: Test Flight 0.9.70 (27)
- Headphones: Bose Quiet Comfort 35 II over Bluetooth
- Phone: iPhone 13 Pro
- iOS: 17.2.1
- At 0.5x or 1x speed: pushing the headphones button correctly starts audio when stopped, and stops audio when playing.
- At 1.2x, 1.5x, 2x, 3x: pushing the headphones button again starts audio when stopped, but only skips backward and continues playing if already playing. (I.e. the bug)

Seems the “toggle” function may have been a red herring all along… apologies for the distraction!

KaiStarkk (approved):

KaiStarkk:
> Seems the “toggle” function may have been a red herring all along… apologies for the distraction!
@cmccambridge hey you got us within 10 lines of the problem so chalk it up as a win :)
I can share a build and sideload instructions for testing if you like, but prefer if you drop me a message on the discord or similar for that as IPAs contain PII. Repo owner may also want to code review first, not sure, lmk.

advplyr:
Great! Thanks to everyone helping with this!

Issue #1093, opened 2024-02-20
[iOS] Disable scrubbing on MPRemoteCommandCenter


fidoriel:
### Is your feature request related to a problem? Please describe.
In the past days I accidentally touched the progress bar on the 'Now Playing' widget in numerous situations on iOS (this is not a ABS issue, I have this with a lot of apps, I do not like that touching on the progress bar is sensitive.). I had to find the position in the audiobook where I left off, which is not a great listening experience.

### Describe the solution you'd like
Add an option to settings to disable `commandCenter.changePlaybackPositionCommand.isEnabled`. 

### Describe alternatives you've considered
Just disable the `commandCenter.changePlaybackPositionCommand.isEnabled`. 

### Additional context
I thought about why I would like this to be enabled at all. I was not able to find one.
- It is not possible to navigate through a multi hour audiobook precisely. If I do want to find a chapter, I use the app and not the widget. It is not possible to find the exact position.
- If I want to go back 30s, I would use the 10s back button multiple times, because it is way more precise.
- I can change it to the chapter view, so that not the full audiobook is displayed, only the current chapter. Same reason. I would use the go back button or the app to find the chapter.

Therefor, I do not see any reason why allowing scrubbing through the audiobook on the widget.

nichwall:
Duplicate of https://github.com/advplyr/audiobookshelf-app/issues/526 but also includes specific option

fidoriel:
Oh, thank you, did not find it because of different wordings.

It is a question of wished behavior, philosophy of usage. The fix seems trivial.

advplyr:
Duplicate of #526

nichwall:
> It is a question of wished behavior, philosophy of usage. The fix seems trivial.

It just hasn't been done yet. If you can make and test the change, that would be great.

fidoriel:
> > It is a question of wished behavior, philosophy of usage. The fix seems trivial.
> 
> It just hasn't been done yet. If you can make and test the change, that would be great.

I created and tested a pr. So now we can discuss ;)

Issue #691, opened 2023-04-29
Ability to disable Lock Screen progress bar seek


njakes:
### Is your feature request related to a problem? Please describe.

While an audiobook is playing and the Lock Screen widget is visible you can accidentally skip hours of your book by accidentally touching the seek bar. This is problematic if your screen accidentally turns on while your phone is in hand or pocket. I can see how this would be useful for chapter separated into different files. However with chapterized m4b files it allows you to seek though the entire book from the Lock Screen. 

### Describe the solution you'd like

It would be nice to have the ability to just disable the touch seek function of the Lock Screen progress bar. 

### Describe alternatives you've considered

An alternative would be to make the Lock Screen widget show the current chapter progress instead of the whole book/file. I'm not sure if this would work for those with chapters separated into files though.

advplyr:
Duplicate of #526

advplyr:
If you are on Android we can update that issue to be for both iOS and Android

advplyr:
Also related to #239 for showing chapter tracks with single file audiobooks

PR #1094, opened 2024-02-20
Disable ios/android scrubbing in MPRemoteCommandCenter/MediaBrowserServiceCompat

### LINKED ISSUE #525: Chapters are not updated after download
If you download a book and then edit/update the chapters in Audiobookshelf, the Android (iOS?) App will show the updated chapters in the book details, but not show/use the updated chapters in the player. This inconsistency is pretty confusing.

#### To Reproduce

1. Open book without chapters in App. Download book.
2. Go to Audiobookshelf Web and edit chapters.
3. Open App again, go to book details
    - Book details will show updated chapters
5. Play audiobook
    - Player will **not** show/use updated chapters
  
The ABS App clearly has the updated chapters. It is showing them after all. It should use them for playback.

#### Android Issue

App version: 0.9.60-beta
Android version: 12
Device model: OnePlus Nord 2

### LINKED ISSUE #526: Prevent seeking from iOS widget
### Steps to reproduce

1. Open the look screen/notification center
2. Seek inside the audiobook 
3.

### Expected behaviour
When listing to an audiobook, seeking through the whole audiobook from the iOS play widget (the one you see in the notification center/lock screen) is usually unwanted. This behavior can be supported by other apps such as audible which disable the seeking behavior for audiobooks. This is especially true since audiobooks are quite long and seeking through 20 hours is not a good experience using your fingers (seeking through a single chapter and using the chapter control is another story). 

This behavior is especially annoying since it's quite easy to do on accident. For my current audiobook, this has happened around 6 times during the first 10 hours of playback, and finding where I left off can be pretty hard. 

This is the widget I am talking about: 

Dragging through the little white bar is quite easy to do by accident.

### Actual behaviour
Trying to scroll/seek should not register at all. 

Changing that behavior should be quite easy here:

https://github.com/advplyr/audiobookshelf-app/blob/master/ios/App/Shared/player/AudioPlayer.swift#L601

You should probably change [seekForwardCommand](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/1618981-seekforwardcommand), [seekBackwardCommand](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/1618982-seekbackwardcommand), and/or [changePlaybackPositionCommand](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/1618997-changeplaybackpositioncommand) from [MPRemoteCommandCenter](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter). I am not 100% sure, but you should play and find out what each of those commands do. Unfortunately, I don't have access to xcode to compile and test the app myself right now. 

Generally speaking, it would be very useful to have a better "history" of playback sessions and an easy way to reset it. Currently, you can only find "last start" inside the settings from the webapp. It would be cool if this view was available for every audiobook directly. It would also be very helpful if large jumps create new sessions. For example, if I skip ahead/backwards for more than two minutes (and my current listen time is more then 60 secs), a new session should be created. This would make recovering a session easier because sometimes the ios very does not sync with the web version (as in, if I listen on the web, the ios app does not update the position, not sure if I will make another issue for that in the future). 

### Environment data

Audiobookshelf Version:

- [ ] Android App?
- [x] iOS App?

#### Android Issue
Android version:

Device model:

Stock or customized system:

#### iOS Issue
iOS Version: 16.1.2 (probably not relevant)

iPhone model: not relevant

If you have any more questions, please let me know.

fidoriel:
This could close #526 #1093 #691
Similar to #683. I do not know why #683 this was closed in the first place by the author.

Scrubbing through some kind of audio with a length > 30 minutes is objectively not possible in an accurate way.
Disabling has the benefit that the audiobook position can not be changed by accident.

Tested and Working on iPhone 13 mini, iOS 17.2.1

This is mainly a discussion about usability and philosophy of use, if this is the correct approach without a setting.


fidoriel:
I have changed it to be configurable within the apps settings. Also added an untested version for adroid.
The current behavior is disallowing seeking. After enableing it, it will not change on the current 'Now Playing' widget. The widget will not be repainted. This is the same behavior like on the `jump` buttons, so I think it is consistent and fine.

fidoriel:
Also tested it on android. Unlike iOS, where on the next widget redraw (audiobook change), the android app needs to be restarted so that the change will take effect, because this part of the widget is not redrawn. This could be changed, but I do not know how important that is. There are two possibilities: Add a message to restart or add a settings watcher/draw the widget again.

fidoriel:
Now `LabelAllowSeekingOnWidgetHelp`can be removed. Will you do that? Should I? Thanks for the android fix, never did Kotlin nor android api before.

advplyr:
I'm working on it because I want to change the wording and name of the variables since they can be confusing for android users. I'm testing on iOS now

advplyr:
In Android the term "widget" typically refers to home screen widgets and "Now Playing" is a feature for detecting the names of songs playing. I landed on using the term "Media Controls" to describe that audio player for iOS and Android (discussed in Discord). If users are confused by that terminology we can update it again with feedback. If you want to update the German translation in this PR before I merge it that would be great, otherwise I'll merge it shortly.

I updated iOS to also respond to that setting change in real time. In my testing I had to re-start the audio player to see the setting changed. Now on iOS and Android you can toggle the setting and see change right away while listening.

Thanks for the contribution, this is a nice addition.

fidoriel:
Great, I will update German soon.

fidoriel:
The German translation was not that easy. Translating directly changed the meaning, so I decided on this one. Feedback may improve this.

advplyr:
Yeah it was difficult to find a name for this that works for both iOS and Android. We'll see how users respond to it. Thanks!

bloodscript:
@fidoriel I think your translation is technically correct. Maybe an easier wording can be found? Like "erlaube das vor- und zurückspulen im Benachrichtigungcenter"?

fidoriel:
Yeah, maybe :) that's too late now. But `Benachichtigungscenter` is an iOS term. Android just uses `notifications`. You can create a PR, if you think you can improve that. I do not care how this is named, I just want it to be fixed.

PR #1142, opened 2024-03-20
Add: i18n workflow check

nichwall:
This workflow performs an integration test using https://github.com/audiobookshelf/audiobookshelf-i18n-updater. This workflow only checks that keys are alphabetical, and does not include workflows that attempt to automatically copy keys. If keys are not alphabetical, the action fails and reports which file is not alphabetical.


nichwall:
Added some commits to show the integration test working by breaking the alphabetization of keys. I am not sure why the integration check did not run in 4a32de482b02cf41913923efaf44e8eea11a5745, probably because I pushed the two commits to close together in time.

advplyr:
Thanks! I haven't had a chance to dig into it yet

PR #1174, opened 2024-04-11
Update i18n workflow to use 1.2.0

nichwall:
Updated action to check for extra keys in other languages. Same as https://github.com/advplyr/audiobookshelf/pull/2841


advplyr:
Thanks!

PR #1178, opened 2024-04-14
Update Now Playing info with correct default rate.

### LINKED ISSUE #1177: iOS: Certain headphones can't pause ABS when playback speed is above 1x
### Steps to reproduce

1. Install Bose Music app with compatible headphones
2. Begin playback in ABS with playback speed set to something above 1x
3. Try to pause via headphone controls

### Expected behaviour
- Tell us what should happen
- ABS should pause playback

### Actual behaviour
- Tell us what happens
- Nothing. Playback continues as if no command was registered.

### Environment data

Audiobookshelf Version: 2.8.1

- [ ] Android App?
- [X] iOS App?

#### iOS Issue
iOS Version: 0.9.73-beta

iPhone model: iPhone 15 Pro Max

indiefan:
When rate and default rate don't match, third party apps (e.g. Bose Music) can mistake current status as scrubbing (and disable playback controls until the two match, which currently never happens in ABS until the user changes the playback rate back to 1).

Fixes #1177


advplyr:
Thanks!

PR #1179, opened 2024-04-14
Refactor AudioPlayer to use a Rate Manager protocol.

indiefan:
This is a first of two steps to implement the new (iOS 16+) rate management api in ABS.

A follow-up PR will implement the new rate manager.

FYI: I made this a downstream PR after #1178 rather than detangling them, but if there's good reason to keep them completely separate I can do so.


advplyr:
Thanks, this is working well in my testing. I tested this on iPhone 15 running latest iOS 17 and an iPhone 10 running latest iOS 16.

PR #1201, opened 2024-05-12
App bug report format

nichwall:
New bug report format. We tried to follow a similar structure as the feature request so "feature requests" that are incorrectly filed as "bugs" are easier to relabel. It's not exact but an improvement over how it currently is.

Once again, demo is located [here](https://github.com/nichwall/issue_testing/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=app_bug.yaml&title=%5BBug%5D%3A+) and a screenshot below:

image


advplyr:
Thanks!

PR #1205, opened 2024-05-17
Fix bug iOS AudioPlayer where multiple calls to setupRemoteTransportC…

### LINKED ISSUE #1141: Pause not working from notification card.
### Steps to reproduce

1. Have a audiobook playing
2. Go to settings
3. Change jump back interval 
4. Use pause audiobook from notification card or use AirPods to play/pause

### Expected behaviour
- The audiobook should pause.

### Actual behaviour
- The audiobook jumps back a couple of seconds and continue playing. To pause the audiobook I have to use the controls in the app.
- If I force close the app and open it again play/pause works as expected.

### Environment data

Audiobookshelf Version: 2.7.1

- [x]  iOS App

#### iOS Issue
iOS Version: iOS 17.4

iPhone model: iPhone 14 Pro Max

indiefan:
…ontrols resulted in duplicate MPRemoteCommand targets being applied for the same action.

We call this multiple times to pick up settings changes for certain playback controls. This makes that safe, with the assumption that we don't actually ever want multiple targets defined on the same commandCenter action. If at some point we do want that, then we'll need to refactor this to do more state-management, but I figured it was better to start with the simple approach unless/until we need the more complex one.

Fixes #1141.

### DIFF
--- a/ios/App/Shared/player/AudioPlayer.swift
+++ b/ios/App/Shared/player/AudioPlayer.swift
@@ -584,6 +584,7 @@ class AudioPlayer: NSObject {
         let jumpBackwardsTime = deviceSettings.jumpBackwardsTime
         
         commandCenter.playCommand.isEnabled = true
+        commandCenter.playCommand.removeTarget(nil)
         commandCenter.playCommand.addTarget { [weak self] event in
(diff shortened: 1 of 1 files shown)


advplyr:
Works great, thanks!

PR #1208, opened 2024-05-19
feat: Add Time Marker Detection in Episode Descriptions

mfcar:
Here's the corrected text for your pull request:

Several podcast apps, such as Overcast and Pocket Cast, have a feature that detects Time Markers in the episode description. This feature provides a way to jump directly to that time. It's very similar to chapters, but the time markers aren't embedded in the files; they are automatically detected from the description.

Screenshot 2024-05-19 at 14 23 50

This PR adds support for detecting time markers in the AudioBookShelf app as well.

image

The detector identifies time markers in multiple formats (at least minutes and seconds):

```
H:M:SS
H:MM:SS
HH:MM:SS
H:M:SS
```

----

Demo

https://github.com/advplyr/audiobookshelf-app/assets/814828/431ea9b9-2a5f-4b09-aac8-29dc4349c31b

### DIFF
--- a/pages/item/_id/_episode/index.vue
+++ b/pages/item/_id/_episode/index.vue
@@ -329,41 +329,8 @@ export default {
           seekTimeInSeconds = hours * 3600 + minutes * 60 + seconds
         }
 
-        return `<a href="#" class="time-marker" data-time="${seekTimeInSeconds}">${match}</a>`
+        return `<span class="time-marker cursor-pointer text-blue-400 hover:text-blue-300" data-time="${seekTimeInSeconds}">${match}</span>`
       })
-    },
-    clickPlaybackTime(event) {
-      const startTime = event.target.getAttribute('data-time')
-      if (this.playerIsStartingPlayback) return
-
-      this.$hapticsImpact()
-
-      this.$store.commit('setPlayerIsStartingPlayback', this.episode.id)
-
-      const playbackData = {
-        libraryItemId: this.libraryItemId,
-        episodeId: this.episode.id,
-        serverLibraryItemId: this.serverLibraryItemId,
-        serverEpisodeId: this.serverEpisodeId,
-        startTime
-      }
-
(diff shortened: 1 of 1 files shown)


advplyr:
This is great. The RSS feed for Syntax has an issue because they put a link around their timestamps.
https://feed.syntax.fm/rss

image

I couldn't think of a way to handle false positives better. The regex gets complicated when trying to handle these different cases so I'm not sure if it is worth it.

image

nichwall:
Maybe just adding something like the following to ignore matches that have more than 2 digits on the end parts by checking that the preceding or following character is not a digit (like the link with a port in it). Not sure if that's worth it or how often ports would show up in a link.

Before: `\b(?<!\d)`
After: `?(?!\d)\b`

mfcar:
Thanks @advplyr and @nichwall for the feedback.

I was investigating the Syntax RSS feed, and they use an anchor link for the time marker, but the anchor only works on their page. I tested the same feed on other apps and got the following results:

1. PocketCasts: Uses a different color for the anchor links, but when you tap them, nothing happens.
2. Overcast: Uses the same color for the external links, and like PocketCasts, tapping them does nothing.
3. Apple Podcasts: Doesn't render anchor links, only regular links.


Screenshot 2024-05-20 at 22 52 06

---

Based on this, I made some adjustments to the code to handle anchor links in the detector. If a time marker is inside an anchor link (which won't work because we're not on the website page), I replace it with a detectable time marker (the other links remain unchanged).


Screenshot 2024-05-20 at 08 12 42

---

Based on advplyr's test cases, I tested with additional edge cases, and the detector worked well. As nichwall suggested, I changed the script to avoid URLs with a port value, but if the URL contains at least two colons, I can't detect this, or if it is an plain anchor url

Here are some tests that I generated to check:


Screenshot 2024-05-20 at 23 04 41

advplyr:
Great, thanks!

PR #1210, opened 2024-05-23
Add settings to limit Cellular Data (Download/Streaming)

mfcar:
This PR add a new category on the settings page, "Data Settings" with two configurable options:
- Download using Cellular
- Streaming using Cellular

Screenshot 2024-05-23 at 22 52 10

---

These settings provide users with control over their data usage on cellular networks. The available options for these settings are:
- Always (default): This option allows unrestricted use of download and streaming features over cellular data. This option is set as the default to avoid migration issues, as this is the actual standard behavior.
- Never: This option blocks both downloading and streaming over cellular data, displaying an error toast when attempted. [Image 2]
- Ask for confirmation*: This option prompts the user for confirmation before initiating a download or streaming over cellular data. [Image 3]

*If the audio is already downloaded, the confirmation alert is not displayed.

Screenshot 2024-05-23 at 22 51 32

Image 2

Screenshot 2024-05-23 at 22 51 49

Image 3

---

### What is missing:
~~- For now this feature is only implemented on the latest podcast episodes page, needs to be applied on the other pages that contains the play and download buttons~~
~~- The features have been tested in a web browser. I need to test on mobile devices (but I only able to test on iOS)~~ https://github.com/advplyr/audiobookshelf-app/pull/1210#issuecomment-2132397578
~~- Improve phrasing~~

---

### Demo:

https://github.com/advplyr/audiobookshelf-app/assets/814828/e40b24f4-79da-4ab5-83ec-32d1b50330ae

---

Feedback is appreciated


ZLoth:
> This PR add a new category on the settings page, "Data Settings" with two configurable options:
> 
> * Download using Cellular
> * Streaming using Cellular
> 
> Screenshot 2024-05-23 at 22 52 10

A line should be added at the bottom should indicate that "Use of cellular network may incur data charges by your carrier."


> These settings provide users with control over their data usage on cellular networks. The available options for these settings are:
> 
> * Always (default): This option allows unrestricted use of download and streaming features over cellular data.
> * Never: This option blocks both downloading and streaming over cellular data, displaying an error toast when attempted. [Image 2]
> * Ask for confirmation*: This option prompts the user for confirmation before initiating a download or streaming over cellular data. [Image 3]

Default should be "Ask for confirmation", not "Always". Assume that the user is on a data limited mobile network. 

> Screenshot 2024-05-23 at 22 51 49
> Image 3

A third button should be added called "Settings" which takes the user to the options for Streaming/Download settings. Also, the text should be modified so that it states "You are about to stream using cellular data. This may include carrier data charges. Do you wish to continue?"

mfcar:
> Default should be "Ask for confirmation", not "Always". Assume that the user is on a data limited mobile network.

The current behavior is to always allow streaming and downloading. I set this as the default to prevent flooding users who update the app with numerous confirmation messages

> A third button should be added called "Settings" which takes the user to the options for Streaming/Download settings.

I have been investigating this, but afaik, the Dialog capacitor plugin only allows for two buttons in the prompt. It would be necessary to implement a custom component.

nichwall:
I didn't think I've used an app that doesn't default to "always" at install, so I would say that defaulting to Always matches expected behavior for new users. I have limited data, so that is something I check when going through settings for a new app.

I agree about adding "may incur additional charges."

If there is not a way to make a third option, also add "this setting can be changed in the App settings" or something. It could be nice if there was a button that was "always allow for [book/podcast]", but I think that would be a lot more complicated to implement and confuse users about which books had which permissions.

advplyr:
This is a great start.
The default should be "always" for both, as you have it.

I'm not sure about having the "may incur additional charges." text. This seems too obvious to put in a modal to the point where it could be confusing that it is in there. I haven't been on a limited data plan in a long time so I'm not the target audience.

On the frontend I'm already passing down a flag for if the network connection is over a metered (limited data) or unmetered (unlimited data) connection. This is only set for Android and I just realized that on iOS we are showing "unmetered" for everyone which is a bug to fixed. I'm not sure if we can get that information on iOS yet.

https://github.com/advplyr/audiobookshelf-app/blob/f047f6a23b8d27f7d1231a2c9fbaf8e643e2d109/store/index.js#L18

For android we get the unmetered information with this https://developer.android.com/reference/android/net/NetworkCapabilities

A metered connection doesn't mean that data costs money because you can make your wifi network metered but according to android docs unmetered should mean data is not billable

advplyr:
Also for this implementation I think we should look at putting this logic someone we only need to write it once.
This is something I've been meaning to do for a while with all the different Play buttons. 

We may be able to put that check in the `AudioPlayerContainer.vue` `playLibraryItem` function

mfcar:
Thanks for the feedback.

I've made changes to the phrases based on the feedback and centralized the confirmation code in a mixin, making it easier to update.

Following @advplyr  suggestion, I've moved the streaming permission to `AudioPlayerContainer.vue`. Now, the permission check is resolved in the same way whenever the player is called.

> On the frontend I'm already passing down a flag for if the network connection is over a metered (limited data) or unmetered (unlimited data) connection. This is only set for Android and I just realized that on iOS we are showing "unmetered" for everyone which is a bug to be fixed. I'm not sure if we can get that information on iOS yet.

Due to this, I'm checking if the `networkConnectionType` is equal to `cellular`, and it worked well on iOS. However, I'm uncertain if this can be used in place of checking if it's a metered network. I have data cap on my cellular network, but on the `ConnectorIndicator.vue` component I keep receiving the message that I'm on a unmetered connection.

I was able to generate an iOS build and test it on my cellphone. Here's the demo (unfortunately, I don't have an Android device to test):


https://github.com/advplyr/audiobookshelf-app/assets/814828/eadda950-1bbc-43e8-a3bd-7ce2717c65da

advplyr:
> Due to this, I'm checking if the `networkConnectionType` is equal to `cellular`, and it worked well on iOS. However, I'm uncertain if this can be used in place of checking if it's a metered network. I have data cap on my cellular network, but on the `ConnectorIndicator.vue` component I keep receiving the message that I'm on a unmetered connection.

That flag will always be unmetered on iOS because I only added support for it on Android.

mfcar:
> > Due to this, I'm checking if the `networkConnectionType` is equal to `cellular`, and it worked well on iOS. However, I'm uncertain if this can be used in place of checking if it's a metered network. I have data cap on my cellular network, but on the `ConnectorIndicator.vue` component I keep receiving the message that I'm on a unmetered connection.
> 
> That flag will always be unmetered on iOS because I only added support for it on Android.

I created a new pull request to implement support for checking if a network is metered or unmetered on iOS (I'm not fluent in Swift):
#1217

ZLoth:
The reason why I'm advocating "Ask for confirmation" instead of "Always" comes from my days of doing support for a Dial-Up Internet Service Provider. We _always_ provided a disclaimer about checking to see if a number is local call because we were not responsible for the long distance charges. This was particularly important as one calling area with a not-loved telephone provider, and only one phone number was a local call to that area.

It can be safely assumed that someone who is setting up audiobookshelf is technically savvy, especially if they set it up as a docker container, port forwarding, and a reverse proxy. But, they may be sharing their audio library with some family members who aren't as technically competent plus they are on a fixed income and a limited data plan. I checked https://www.nerdwallet.com/p/best/finance/cheap-cell-phone-plans , and there are plans with limited data, although unlimited data is probably preferred if using AudioBookShelf. I want to avoid getting yelled at and having to foot the data overages.

components/tables/podcast/LatestEpisodeRow.vue
```diff
@@ -259,22 +265,12 @@ export default {
...
-            libraryItemId: this.libraryItemId,
-            episodeId: this.episode.id
-          })
-        }
+        this.$eventBus.$emit('play-item', {
+          libraryItemId: this.libraryItemId,
+          episodeId: this.episode.id,
+          serverLibraryItemId: this.libraryItemId,
+          serverEpisodeId: this.episode.id
+        })
```
advplyr, line 273: Is there a reason for removing the local playback here?

advplyr: I see now. There is a bug in `latest.vue` where localLibraryItemId is always null. I'm looking into this

mfcar: I moved to the `AudioPlayerContainer.vue` to centralize the handling of local files and avoid showing the confirmation dialog for these media states.

advplyr: I fixed this and playing local is possible now

advplyr: For latest podcast episodes we were never passing through the local IDs so downloaded episodes were still direct streaming

advplyr:
Thanks! This is working great on android & iOS

ZLoth:
Awesome! Thank you!

Issue #847, opened 2023-08-27
Headphone remote control - enable `next` and `previous` navigation option on Android


sandreas:
I'm not sure this is documented anywhere, but I think currently, it is not possible to enable the ability of navigating to `next` or `previous` via headphone remote control on android.

So there are multiple different implementations, where some of the keycodes share common ground. For android there actually is an *expected behaviour*: https://source.android.com/docs/core/interaction/accessories/headset/expected-behavior

advplyr:
What do you expect to happen when you use a next/prev action?

sandreas:
> What do you expect to happen when you use a next/prev action?

Depends on the content / personal preferences.

Music / audio books with multiple files:
- Next: Next track, if available, stop when no more tracks available
- Prev: If within the first 3 seconds of a track, skip to previous, otherwise skip to the beginning of a track

Files with chapters (audio books, podcasts):
- Next: Next chapter, if available, stop when no more chapters available
- Prev: If within the first 3 seconds of a chapter, skip to previous, otherwise skip to the beginning of a chapter

However, there are some minor tweaks I would love to see in the settings, although they would take a lot of effort to implement:

- Instead of next / prev track or chapter, I would configure next/prev to fast forward / rewind 5 minutes on files with chapters
  - Reason: Chapters are unpredictable - some books have 3 minute chapters, some have 15 minute chapters

For play/pause i would like to see an increasing amount of rewind seconds, that resets after 3 seconds of no action, example:
- First Pause: Rewind 10 seconds and pause
- Play again
- Pause again (within 3 seconds): Rewind 20 seconds and pause
- Play again
- Pause again (within 3 seconds): Rewind 30 seconds and pause
- Play again
- Pause again (within 3 seconds): Rewind 60 seconds and pause
- Play again (more than 3 seconds) => reset the increasing timer
- Pause again: rewind 10 seconds

You could also increase that to 2 times per time step (increase timer after 2 rounds of rewinding). I think in most cases you pause multiple times you want to rewind the book just a few seconds, because you missed something. But the more you do it the more you want to go back.

gazpachoking:
Just weighing in on how I use headphone buttons: I'd _never_ want my headphone buttons to do next or previous track/chapter. That's way too long in an audio book, and way too easy to mess up with no way to get back to where you just were. The only button I would personally use from my headphones is a short rewind 30-60s. When I'm out running or walking, sometimes I need to skip back because I got distracted and missed something.

[...]

PR #1218, opened 2024-05-28
Added extended headset controls (headset button click handling)

sandreas:
Fixes https://github.com/advplyr/audiobookshelf-app/issues/847

In addition I should say:

- This PR contains just another method of handling headset clicks but keeps the existing behaviour intact - so it should not break any existing behaviour
- I created an extra "Headset Settings" headline in the settings
- This currently contains only one setting switch
- In the future it may contain Additional select boxes to set settings for:
  - 1 click = play / pause
  - 2 clicks = next chapter or track (currently jump forward 5 mins)
  - 3 clicks = prev chapter or track (currently jump backwards 5 mins)
  - Hold = jumpBackwards
  - 1 click + hold (fast forward)
  - 2 click + hold (rewind)

Feedback welcome.


android/app/src/main/java/com/audiobookshelf/app/player/MediaSessionCallback.kt
```diff
@@ -258,6 +277,80 @@ class MediaSessionCallback(var playerNotificationService:PlayerNotificationServi
...
+
+        KeyEvent.KEYCODE_MEDIA_NEXT -> {
+          clickCount += 2
+          Log.d(tag, "=== handleCallMediaButton: Media Next, clickCount=$clickCount")
+        }
+
+        KeyEvent.KEYCODE_MEDIA_PREVIOUS -> {
+          clickCount += 3
+          Log.d(tag, "=== handleCallMediaButton: Media Previous, clickCount=$clickCount")
+        }
```
advplyr, line 322: I'm not understanding the logic here. How do we know that KEYCODE_MEDIA_NEXT/PREVIOUS are clicks and not a separate next/prev button press?

sandreas: Android is handling specific shortcuts not emitting exact events, but overriding them by combined events:

- One click: Play/Pause
- Double click: Next
- Triple click: Previous
- Long press: Voice over (on newer Devices)

This cannot be prevented, because it is a system / kernel behaviour. 

So double-clicking does NOT emit 2 raw click events, but 1 single (automatically rewritten) event `KeyEvent.KEYCODE_MEDIA_NEXT`. 

I handle `KeyEvent.KEYCODE_MEDIA_NEXT` as if 2 single clicks had happened and increase the click count by 2. Same for the triple-click, which happens to be emitted as `KeyEvent.KEYCODE_MEDIA_PREVIOUS`  by Android instead of 3 single keydown/keyup event-combos.

<details>
This code wont work on Android Versions < 7.1. There is no possibility to detect `KEY_DOWN` events, because Android internally waits till `KEY_UP` doing nothing and then emits `KEY_DOWN` and `KEY_UP` together with no delay in between. So the scenario of holding down the key, that should emit something like:

- Press down: `KEY_DOWN`
- Wait: `// wait 1248ms`
- Release: `KEY_UP`

on Android 7.0 devices is 

- Press down: `// nothing`
- Wait: `// wait 1248ms`
- Release: `KEY_DOWN`, `KEY_UP`

</details>

advplyr: My understanding is that these events are emitted by the bluetooth device and so depending on the bluetooth device it might emit a KEYCODE_MEDIA_NEXT on 2 clicks if it is setup for that.
Or you might have a bluetooth device that has separate buttons next to play/pause that is for KEYCODE_MEDIA_NEXT.

Did you find any documentation on these KeyEvents that say when/how they are triggered?
I think this needs to be tested on a bluetooth device that has NEXT/PREV media buttons. I'll see if I have one. I think many bluetooth in vehicles have this.

sandreas: > My understanding is that these events are emitted by the bluetooth device and so depending on the bluetooth device it might emit a KEYCODE_MEDIA_NEXT on 2 clicks if it is setup for that.

Ah, now I understand. You mean a [device like this](https://www.amazon.de/exec/obidos/ASIN/B0CQCPVBGF/wwwgeschenke-inspiration-21) - a car bluetooth receiver with multiple buttons (e.g. for `next` and `prev`).

The PR is NON-DESTRUCTIVE, so that there are no behaviour changes, as long as the according setting (`extended headset controls`) is disabled, which it is by default. So no fear of breaking anything, that works fine at the moment.

I own such a car bluetooth recevier with buttons for `play/pause`, `next` and `previous` and tested it. It shows the following behaviour (which I would have expected):

- `play/pause`
  - single click = play/pause
  - double click = next
  - triple click = prev
  - 4 clicks or more (fast execution) = nothing 
  - multiple clicks (slow execution) = multiple play / pause events
- `next`
  - single click = next
  - 2 clicks or more (fast execution) = nothing (translated to 4 clicks or more)
  - multiple clicks (slow execution) = multiple next events
- `prev`
  - single click = prev
  - 2 clicks or more (fast execution) = nothing (translated to 6 clicks or more)
  - multiple clicks (slow execution) = multiple prev events

So I think it works like expected. Even better, on bluetooth devices with A SINGLE button, you can now use it for `next` and `prev`. 

The only unexpected behaviour might be that multiple fast `next` clicks won't do anything, as long as the timer is not hit - you have to click slower, but since this feature is disabled by default, I don't see this as a huge problem.

android/app/src/main/java/com/audiobookshelf/app/player/MediaSessionCallback.kt
```diff
@@ -258,6 +277,80 @@ class MediaSessionCallback(var playerNotificationService:PlayerNotificationServi
...
+
+    if(clickTimerScheduled) {
+      Log.d(tag, "=== clickTimer cancelled ($clickTimerId): clicks=$clickCount, hold=$clickPressed =========")
+      clickTimer.cancel()
+      clickTimer = Timer()
+    }
+
+    clickTimer.schedule(650) {
+      Log.d(tag, "=== clickTimer executed ($clickTimerId): clicks=$clickCount, hold=$clickPressed =========")
+      playerNotificationService.handleClicks(clickCount, clickPressed)
```
advplyr, line 344:
```suggestion
    Handler(Looper.getMainLooper()).post {
      playerNotificationService.handleClicks(clickCount, clickPressed)
    }
```

Calling handleClicks on the main thread here should make it so the PlayerNotificationService doesn't need CoroutineContext.

sandreas:
@advplyr I accepted the change suggestion, but how would we deal with the conflicts? I'm not that experienced in doing code reviews with conflicts on GH.

advplyr:
I can handle the merge conflicts.

That change I suggested is just so that you can remove all the coroutine stuff you added in the PlayerNotificationService. I also see code commented out so I'm not sure if this is still in progress?

sandreas:
> I can handle the merge conflicts.
> 
> That change I suggested is just so that you can remove all the coroutine stuff you added in the PlayerNotificationService. I also see code commented out so I'm not sure if this is still in progress?

Not really in progress but hoping to get corrected by a professional :-)

I'm not a kotlin expert and the coroutine stuff was new to me, I was just trying to get this working without breaking anything.

So feel free to merge and improve the code whereever you think it's worth.

Once the merge is done I'm going to test this   carefully and give feedback.


Future steps I plan to implement to improve this feature:

- next/prev chapter
- settings in form of multiple select items to configure click-codes
- use exoplayer rewind/ff instead of my custom implementation
- add settings for timings (click. Timeouts, ff step size, etc)

For now I just hope this PR gets accepted somehow ;-)

sandreas:
@advplyr Any news? If there is anything I can do to support you, let me know...

sandreas:
No reaction for a long time...

PR #1222, opened 2024-06-04
Update font settings in EpubReader and Reader components

BimBimSalaBim:
Update font settings in EpubReader and Reader components to include a "Font Boldness" slider which applies a stroke to the text # 3020 on main repo


advplyr:
Thanks!

PR #1231, opened 2024-06-10
Fix Android Large Download Failure

troy117:
The updated InternalDownloadManger takes in the downloaded file as the parameter.  This way if it exits, it will use it's byte length as the starting Range parameter for the remaining download.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/InternalDownloadManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/InternalDownloadManager.kt
@@ -70,13 +70,10 @@ class BinaryFileWriter(outputStream: OutputStream,
       val dataBuffer = ByteArray(CHUNK_SIZE)
       var readBytes: Int
       var totalBytes: Long = initialBytesWritten
-      var i: Int = 0;
       try {
         while (input.read(dataBuffer).also { readBytes = it } != -1) {
           totalBytes += readBytes.toLong()
           outputStream.write(dataBuffer, 0, readBytes)
-          i++;
-          if(i%1000==0) Log.d("Writer","Progress Info: $totalBytes")
           progressCallback.onProgress(totalBytes, (totalBytes * 100L) / length)
         }
         progressCallback.onComplete(false)
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/InternalDownloadManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/InternalDownloadManager.kt
@@ -12,7 +12,7 @@ class InternalDownloadManager(private val file:File, private val progressCallbac
   private val maxRetries: Int = 10  //Not sure what number this should be.  But at least 5
 
   @Throws(IOException::class)
-  fun download(url:String, retryCount: Int=0, initialBytesWritten: Long = 0) {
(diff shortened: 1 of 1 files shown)


nichwall:
I spent some time trying to get this and https://github.com/advplyr/audiobookshelf-app/pull/1230 to work both in emulsion and on my phone this morning but was not successful in getting the download to resume.

Is there a specific way you're testing this to show the download resumes? 

I tried:
- killing the app during a download to internal storage and then reopening the app and trying to download again (no change from normal behavior of downloading entire file again)
- turning on airplane mode during download to simulate losing connection (download shows as successful and available offline, but local file size does not match server file size)

troy117:
Using the original code on my personal phone the download would never finish.  It would freeze up.  So I found in the logcat that it would randomly crash with an unexpected end of stream error.  I never encountered a the end of stream error or a way to simulate it during emulation even though I was emulating the exact same settings as my personal phone.  

I was able to test it using my phone and would see that when it encountered the end of stream error it would resume right away without any lack of progress.  I was able to use it to download multiple books without it freezing up.

This is my first time using github and working on a community project.  I should have been a bit more extensive on my testing.  I fixed it for the issue I was having hoping it would be the issue that other people are having.  

I just finished similar testing to what you did and found that when the internet is shut off during emulation it closes the socket and the listeners mark the download as complete even though it is not.  And because it has a file stored, the download button isn't showing up to resume the incomplete file.  

Thank you for bringing this to my attention.  I think I have an idea on how to fix this and will hopefully have an update that will handle these issues.

nichwall:
No worries, I was just trying to review the changes but also haven't really experienced the download issue in the first place myself so not sure how to reproduce it.

I'm not sure if the listener marking the download as finished is due to the current so behavior or your changes. I would expect it is the current behavior, but I'm not as familiar with the mobile apps.

You can continue adding commits to the same branch and this PR will update automatically so it's easier to keep the discussion in one place.

troy117:
I think the hardest thing about this issue was that it is very difficult to replicate the end of stream error that would randomly happen and cause downloads to freeze.  

But your comment has got me thinking about the two other cases that can cause incomplete downloads, loss of internet connection and closing the app partially through a download.  I don't know how often a user might experience these issues, but it seems like a fun challenge to try and fix.  I believe I understand why and how they happen, I'll see if I can come up with solution.

troy117:
I just got a new phone and it would randomly freeze using the audiobookshelf app downloaded form the play store.  

I don't know how to simulate the end of stream, but here is the logs showing that the commit I made continue the download.  
Screenshot 2024-07-01 173139
Screenshot 2024-07-01 173209

nrjohnstone:
Good that people are working on this issue as I've encountered various download timeouts and errors that are probably related to this. As an idea for "testing", in scenarios like this I have deliberately compiled my backend API with "chaos" wrappers around various endpoint (or feature flags etc.)

For example, if we did this with the API, you could enable a "network timeout" chaos parameter that could always stop the download stream from the server end after X seconds .. It can be a bit tricky to emulate a true failure here, and it depends on what API hosting library your using.

Another alternative is to proxy your calls through ToxiProxy which is excellent for this type of thing. You can configure the proxy endpoint to timeout, slow down and various other things to simulate network connectivity being dropped etc..

PR #1241, opened 2024-06-18
example idea, needs feedback and more work

faush01:
Looking for feedback.

The idea is when the app is paused for an extended period of time (say 5 min) the player (play session) closes.
This works, I have tested it with a few virtual devices in Android Studio.

I am looking for feedback on this idea.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/media/MediaProgressSyncer.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/media/MediaProgressSyncer.kt
@@ -33,6 +33,7 @@ class MediaProgressSyncer(val playerNotificationService: PlayerNotificationServi
 
   private var lastSyncTime:Long = 0
   private var failedSyncs:Int = 0
+  private var pausedAtTime:Long = 0
 
   var currentPlaybackSession: PlaybackSession? = null // copy of pb session currently syncing
   var currentLocalMediaProgress: LocalMediaProgress? = null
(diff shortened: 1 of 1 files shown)


advplyr:
I don't think this approach will be suitable because we don't want to close the player after an extended pause. That would force the user to open the app and start a new session.
A better approach would be to open a new playback session (without closing the audio player) when the play button is pressed after an extended pause. This could get complicated though with how the listening sessions are setup now so it may require some refactoring on the backend.

faush01:
I did look at just creating a new play session, but this was complex and the way the system with playback monitor and sync monitor are designed now they are all interlinked with references to each other and copies (clones) of the play session being passed between them, creating a new session looks like it would need to tear down the whole playback monitor and sync manager anyway, in effect it would be the same as killing the current play action and creating a new one.

That is why I thought it would be easier, and safer, to stop the current player so the user can start it again next time they want to listen which does all the setup again so a new session etc.

To me this solution if added as an option and OFF by default would solve some people's issues with session over multiple days, i.e. after 5 min the paused session stops, and it would not be for everyone, but it does solve the issue I am seeing.

I still think this is a valid approach (close playback if paused for X minutes) and something the user can easily understand it they turn it on in the settings, as long as the setting is clear like:

"Close player if paused for X minutes" (default never)

This should be very clear for a user.

I will have a further look at just creating a new play session if paused for x minutes, but this will be in a different PR.

Issue #1158, opened 2024-03-30
Login on iOS using enter button defaults to “Remove this server config?” instead of logging in


trjohnson19:
### Steps to reproduce

1. Be logged out of the Audiobookshelf iOS app 0.97.3-beta after having previously logged in to an account for a given server (that is, there is a saved URI / server config)
2. Fill in username and password (in my case using my Bitwarden password manager in case that is relevant)
3. Press `enter` with the cursor in either the username or the password field

### Expected behaviour
- The app logs in / performs the behavior of the Submit button

### Actual behaviour
- The app presses / activates the trash button to delete the given server config

### Environment data

Audiobookshelf Version: 0.97.3-beta

- [ ] Android App?
- [x] iOS App?

#### Android Issue
Android version:

Device model:

Stock or customized system:

#### iOS Issue
iOS Version: iOS 17.4.1 (21E236)

iPhone model: iPhone 12 Pro

trjohnson19:
Example screenshot:

![image](https://github.com/advplyr/audiobookshelf-app/assets/77356759/67995f66-d865-404b-9a06-d4d162cb805d)

PR #1252, opened 2024-06-30
Connection form "Enter" to submit instead of delete

nichwall:
Fixes https://github.com/advplyr/audiobookshelf-app/issues/1158

This PR switches the order of the "delete" and "submit" buttons for editing a server connection, then uses `flex-row-reverse` to keep the UI looking the same.
Pressing "enter" while being focused on the password field now submits and attempts a connection instead of trying to delete the server config.

Tested on Android, Pixel 6a.


advplyr:
The better way to handle this is to set the button to `type="button"`. The other `ui-btn` component takes a prop `type` so that would need to also be added to the `ui-icon-btn`. I can do that actually it will only take a min

nichwall:
Sounds good. Thanks for fixing these little code things I try. Feel free to just close the PR if you do it in another commit.

advplyr:
I got it here https://github.com/advplyr/audiobookshelf-app/commit/03a46decba444b68590cd7e52e07c25577ecdc93
I set the default for the icon btn to be type "button" because it made sense.

PR #1257, opened 2024-07-16
split groups of save history items that have different status results

faush01:
Simple change to split groups of save history items when the status for items in a group changes.

I noticed that the status (icon) was only showing for the first item in a group, if there were say 150 saves and the last 30 failed then it would still show success.

Example of the changes:

image

### DIFF
--- a/pages/media/_id/history.vue
+++ b/pages/media/_id/history.vue
@@ -71,6 +71,7 @@ export default {
       let lastKey = null
       let numSaves = 0
       let numSyncs = 0
+      let lastSaveName = null
 
       this.mediaEvents.forEach((evt) => {
         const date = this.$formatDate(evt.timestamp, 'MMM dd, yyyy')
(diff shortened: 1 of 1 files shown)


nichwall:
Tested and working on Android 14, Pixel 6a.

Test 1 procedure:
- Started streaming book (let it buffer a bit, got a few sync events)
- Turned off server (lost connection but did not affect app connectivity)
- New save icon with red icon was created and counted

Test 2 procedure:
- Downloaded book
- Started playing, waited for a sync event
- Turned on airplane mode on phone
- Waited for next save event, new event was created as expected
- Paused, turned off airplane mode, synced with server

faush01:
Do you need anything else from me on this?

ISO-B (approved):
Works as expected with Oneplus One, Android 11. Tested with same procedures as @nichwall

advplyr:
Thanks!

Issue #1264, opened 2024-07-19
[Enhancement]: bookshelf loading indicators as toast notifications


Rdkang:
### Describe the Feature/Enhancement

To use toast notifications for the **"Attempting server connection..."** and **"Loading server data..."** messages instead of as text above the bookshelf.

### Why would this be helpful?

The current loading indicators causes the row of books to shift upwards after loading, which i find to be quite jarring. 
By using toast notifications for the messages, it stops the layout shift.

### Future Implementation (Screenshot)

https://github.com/user-attachments/assets/01de8e33-ce9d-482b-a252-42996c168e3f

### Audiobookshelf App Version

Android App - 0.9.74

### Current Implementation (Screenshot)

https://github.com/user-attachments/assets/2e8c0037-88f8-45b5-b34c-a46a853aa781

PR #1265, opened 2024-07-19
refactor: Move Loading Indicators to Toast Notifications

Rdkang:
implements #1264

### DIFF
--- a/pages/bookshelf/index.vue
+++ b/pages/bookshelf/index.vue
@@ -1,10 +1,5 @@
 <template>
   <div class="w-full h-full min-h-full relative">
-    <div v-if="shelves.length && isLoading" class="w-full pt-4 flex items-center justify-center">
-      <widgets-loading-spinner />
-      <p class="pl-4">{{ $strings.MessageLoadingServerData }}</p>
-    </div>
-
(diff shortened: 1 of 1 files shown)


ISO-B (approved):
Seems to work fine.

Tested with Oneplus One Android 11

nichwall:
Would it make more sense to move this from being part of the icon in the top bar, where users can tap to see what state the app is in?



I know at one point in the past someone mentioned reducing the amount of moving things (in general, not just this loading indicator).

I think these connect messages were originally shown there, but people were getting confused about what was actually happening, so I may just be saying to return to how it was before and we'll get the same sort of request to add more information.

advplyr:
I think the toasts are more obtrusive then how it is now. We could return to putting a spinner in the appbar like @nichwall is saying without having text anywhere saying it is connecting.
Or we could go back to blocking user interaction until the server connection attempt is finished

nichwall:
I don't think blocking user input is a good option since there were a lot of complaints about that causing problems in poor service (since that would also prevent disconnecting from the server in the side menu).

I think just updating the connection widget to show "loading" would be good, since most people don't really care if it's still in the connecting or loading phase, only that it isn't complete. If it's taking a long time, they can tap on it to see which part it's stuck on.

PR #1272, opened 2024-07-27
Added ability to download whole series from series view

ISO-B:
**Why/Use case:**
When traveling my kid has phone for listening audiobooks, but doesn't have access to internet so I needed to download books for him. We have quite lot of children books that are sort, but there might be tens of books in single series. This will make downloading those series much faster.

**What is done:**
Added new menu item to toolbar for downloading whole Serie with single click. This should only be visible in serie view. It will only download books(audio and ebook) that are missing from devices. Tested with android device, but I don't have access to IOS device so couldn't test with that.

I have very limited experience with Vue so there might be some dumb mistakes.


pages/bookshelf/series/_id.vue
```diff
@@ -19,10 +23,167 @@ export default {
...
+      })
+
+      if (payload && payload.results) {
+        console.log('Received payload', payload)
+        this.books = payload.total
+
+        for (let i = 0; i < payload.results.length; i++) {
+          console.log(payload.results[i].numFiles)
+          if (!(await this.$db.getLocalLibraryItem(`local_${payload.results[i].id}`))) {
+            this.numFiles += payload.results[i].numFiles
```
nichwall, line 84: This is including any file which is not an audio file as ebook files. Downloading a series which has no ebook files shows multiple ebook files as being downloaded due to cover images.

pages/bookshelf/series/_id.vue
```diff
@@ -19,10 +23,167 @@ export default {
...
+        fetchFinished = await this.fetchSerieEntities(page)
+        page += 1
+      }
+      if (fetchFinished !== true) {
+        console.error('failed to fetch serie books data')
+        return null
+      }
+
+      // Format message for dialog
+      let startDownloadMessage = `Serie "${this.series.name}" has ${this.books} book${this.books == 1 ? '' : 's'}. Download missing ${this.libraryIds.length} book${this.libraryIds.length == 1 ? '' : 's'} with ${this.numFiles} file${this.numFiles == 1 ? '' : 's'}`
```
nichwall, line 140: We can make these strings translatable by setting them up like the following in `strings/en-us.json`.
`SeriesDownloadMessage: "Download missing {0} book(s) with {1} file(s)?",`
We can then use the following to substitute values into the string for display. The `{num}` syntax is replaced by the array index below.
`return this.$getString('SeriesDownloadMessage', [this.libraryIds.length, this.numFiles])`

I think we can also simplify the string to just be what is going to be downloaded, the user doesn't need to see how many things are already downloaded.

nichwall (changes requested):
Working for me on Android 14 on a Pixel 6a. Looks good to me overall. :)

My main comments are:
- While "serie" is technically a word in English, it is pretty much never used in favor of "series". Led me to an interesting read, though! https://english.stackexchange.com/questions/483198/when-did-serie-become-obsolete-in-english
- The download file count is incorrect for ebooks.
- The download button would probably be better in the bar instead of in the dropdown until there are more items to put in the dropdown.

pages/bookshelf/series/_id.vue
nichwall: Maybe the string should instead just be "download missing {0} book(s) of {1} file(s), totaling {2} MB?". That way we can simplify the messages some more and don't need to worry about identifying ebook versus other file types, and users probably want to know how big of a download it would be.

ISO-B: Changed download message to suggested.

pages/bookshelf/series/_id.vue
ISO-B: Changing confirmation message to suggested resolved this

ISO-B:
- Download file count problem should be now fixed as I changed confirmation message to suggested. It makes more sense now
- Moved download button to toolbar
- Changed Serie -> Series. For non native English speaker that word is strange as singular and plural are same and word ends with s. Noticed that thing while writing code for android auto part of app, but forget to correct it in this part.

nichwall:
Thanks! Yeah, English is weird.

I just tested, and noticed if you press "download", "cancel", "download", the size and file count are not reset. The download is still correct, it is only the dialog box which is wrong. That did let me see that "MB" and "GB" were displayed correctly. :)

Another thought, do we want to hide the download button if all books are already downloaded? Or maybe we don't need to hide it and just have the popup say "All books are already downloaded".

ISO-B:
Good catch. Now that you mentioned it it makes sense. Previously I thought that is was because there was cover for each book and count was double. It is now fixed. 

I added popup to tell that everything is downloaded if user tries to download when everything is already downloaded.

nichwall (approved):
Looks good to me on Android. Think we can merge this before the next app release @advplyr ?

advplyr:
This is great, tested on android and ios. Thanks!

PR #1282, opened 2024-08-09
Optimize notification bar display

seedsky:
Optimize notification bar display
1. Change the main title to the title of the currently playing chapter
2. Combine the author and main title for display
3. Rearrange buttons (remove speed adjustment button)

0e2fbf3ebfbef116f06c0ba2328288e


advplyr:
The speed adjustment is mainly for android auto. Unless we can show different actions for android auto I think we should keep it.

springsunx:
Add an option to display the speed control button, which may be better.

PR #1302, opened 2024-08-30
Make it possible to turn the page with volume keys

RaHoni:
I am used to reading EBooks with FBReader but since uploading all EBooks to ABS I wanted to use the ABS-App. Wich works great for the EPub files we have. But I missed the possibility to turn the Pages by pressing Volume down. This was more convenient and in my Opinion more ergonomic. 

The default config is to use the Down-Key for next page and theUp-Key for previous. 

I have a few questions about this PR:

1. I have put the option to change the use of the Volume Keys into the reader settings, is this the right place?
2. I am not sure about the wording to use in the settings.
3. Wich should be the default setting? On as used by FBReader or Off as to not obstruct the change of the Volume for users who don't know the settings?


RaHoni:
@advplyr Is there anything you would want from this PR?

advplyr:
I think the setting belongs in the reader settings and can be enabled by default. 
There were a few updates I made

1. Changing the setting wasn't setting the volume watcher or clearing it. That was only being done on init.
2. `isWatching` is a promise so without await it would always be true
3. When the audio player is open I thought that the volume should be available so I added a check for that
4. Since PDF and Comics don't have settings to open I restricted this to only be for Epubs. It can be extended to PDF and Comics when they have a settings option.
5. Cleaned up the uneven padding in the UI

I'm not sure what good wording would be either so I thought that works.

Thanks!

fawazral:
Hey @advplyr, you noted "When the audio player is open I thought that the volume should be available so I added a check for that". Can there be an option to always have the volume key be the page turn? When I read on my tablet, I control the volume using gestures and exclusively use the volume keys for page turning. Being able to turn the page using the volume keys while an audiobook is playing is the final thing before I ditch using another app to read an ebook!

advplyr:
Hey, we can probably do that. You can open a github issue feature request for it

Issue #141, opened 2022-04-24
Cover Image in Downloaded Audiobook missing in Android Auto


Sacerdoss:
### Steps to reproduce

1. Download Audiobook from Server
2. Start Playing Audiobook
3. Cover of the Audiobook is black in the player. 

### Expected behaviour
- Cover is displayed

### Actual behaviour
- A black image is shown

### Environment data

Audiobookshelf Version:

- [x] Android App?
- [ ] iOS App?

#### Android Issue
Android version: 11

Device model: Galaxy S10

Stock or customized system: Custom Lineage OS 18.1

#### iOS Issue
iOS Version: -

iPhone model: -

B0rax:
On iOS I have the same problem with all audiobooks (not downloaded books)

MidnightSnowleopard:
Downloading with 0.9.42-beta on Android with server v2.0.2 shows logs on server of "Api called without a token" for all calls to download cover.

Can provide more details if needed.

advplyr:
This is fixed with [v0.9.43-beta](https://github.com/advplyr/audiobookshelf-app/releases/tag/v0.9.43-beta)

feerlessleadr:
It looks like this is fixed for the app itself, but the issue is still present in Android Auto (Please let me know if you'd like me to open a separate issue on this).

advplyr:
Partially fixed in [v0.9.60-beta](https://github.com/advplyr/audiobookshelf-app/releases/tag/v0.9.60-beta)

In my testing I got the cover for downloaded items to show in the list views but once starting playback the cover was not showing.

gazpachoking:
Yeah, still having this issue on latest version. I see the small cover before I start playing the book, and covers show large in the background for streaming books, but playing a downloaded book just has a blank background on the player in android auto.

Ansem93:
If you will work on it again I can help it testing if you want :)

SenorSmartyPants:
Still an issue. Cover shows in queue, but not on large nowplaying screen.
Android 10, Android App version 0.9.74-beta, server ABS v2.10.1.
Book is a collection of MP3s, 1 per chapter. Book was downloaded to local in app storage (default storage on android 10).
Here's a couple screenshots. 
![IMG_20240609_171831756](https://github.com/advplyr/audiobookshelf-app/assets/991618/bfb08018-01da-4de1-a521-5ea315b9cd24)
[...]

PR #1303, opened 2024-08-31
Do not call setIconBitmap instead of setIconUri if book is local

Sp4rky001:
Addressing issue #141

Call setIconBitmap instead of setIconUri if book is local (bitmap is not null)


android/app/src/main/java/com/audiobookshelf/app/player/PlayerNotificationService.kt
```diff
@@ -320,11 +320,11 @@ class PlayerNotificationService : MediaBrowserServiceCompat()  {
         val mediaDescriptionBuilder = MediaDescriptionCompat.Builder()
           .setExtras(extra)
           .setTitle(currentPlaybackSession!!.displayTitle)
-          .setIconUri(coverUri)
+//          .setIconUri(coverUri)
```
advplyr, line 323:
```suggestion
```

advplyr:
Nice find! I looked into that bug a few times and never saw that. Thanks!

Sp4rky001:
Nice!  My first actual contribution to an open source project!

PR #1314, opened 2024-09-07
Allow connection attempt to local server when internet is not available

### LINKED ISSUE #566: Unable to play off local server without internet connection
I have various wall mounted tablets and similar devices dotted around my house that are on my local network, but do not have wider internet access. The ABS server is accessible to them. However, I am unable to connect to it as the Android app checks for internet connectivity before logging into the serve.

Could the app attempt to connect to the specified ABS server, rather than the generic connectivitycheck.gstatic.com?

ashanrath:
Removed the !networkConnected condition on the disabled state for the address input field and the submit button. Text is unchanged, however allows a user to attempt to connect to a local resource even if they do not have an active internet connection.

Fixes #566

### DIFF
--- a/components/connection/ServerConnectForm.vue
+++ b/components/connection/ServerConnectForm.vue
@@ -23,9 +23,9 @@
             <span class="material-icons text-fg-muted">arrow_back</span>
           </div>
           <h2 class="text-lg leading-7 mb-2">{{ $strings.LabelServerAddress }}</h2>
-          <ui-text-input v-model="serverConfig.address" :disabled="processing || !networkConnected || !!serverConfig.id" placeholder="http://55.55.55.55:13378" type="url" class="w-full h-10" />
-          <div class="flex justify-end items-center mt-6">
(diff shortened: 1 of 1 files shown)


WylanSwetsCB (approved):

advplyr:
Have you tested everything else like downloading, streaming, etc?

ashanrath:
> Have you tested everything else like downloading, streaming, etc?

Tested streaming and downloading, both work with no issues. Modified APK with these changes incorporated, deployed to Pixel 7a, tested on both Android 14 and 15. Once you're connected to the server everything seems to work as normal. Also tested connections via direct IP address and FQDN, which had been hardcoded to the dnsmasq service on my router for exactly this scenario.

advplyr:
What I'm confused about is there are a few places that use `networkConnected` other than the server connection form.

The library dropdown in the appbar wouldn't show if that was false. The home page shelves wouldn't load if it was false, only downloaded items would show on the home page.

advplyr:
I think this got fixed in https://github.com/advplyr/audiobookshelf-app/commit/fe9168c6cf505e9c632f662ae0db14256e5fbe93.

And for iOS https://github.com/advplyr/audiobookshelf-app/commit/358197db035e2f792f83a55bdf9a0e4984b0f594

Since this was only a partial fix for the server connect page I will close this out but if those commits didn't fix the issue we can revisit

Issue #1083, opened 2024-02-14
Podcasts on android auto are not in order from new to old (the are in the android app)


Whatsek:
### Steps to reproduce

1. Use ABS on Android Auto
2. Open library an choose a podcast
3. Podcasts should be ordered fom new to old

### Expected behaviour
- Podcast list should be in order from new to old

### Actual behaviour
- They are randomly ordered. Can`t see a patern

### Environment data

Audiobookshelf Version:
2.7.2 (and around 4 previous versions, I guess)

- [X] Android App?
- [ ] iOS App?

#### Android Issue
Android version:
14 (issues was also on 13)
Device model:
Nothing Phone 2
Stock or customized system:
Stock
#### iOS Issue
iOS Version:

iPhone model:

Issue #679, opened 2023-04-18
Android Auto - Access to playlist of unlistened edisodes


update-freak:
### Is your feature request related to a problem? Please describe.
When I start driving in my car I want to start the playing with a podcast playlist with unlistened episodes instead of selecting one specific episode from a podcast.

### Describe the solution you'd like
I would be nice to have access to a playlist of the unlistened episodes.

### Describe alternatives you've considered
I tried to create a playlist, but I did not have access to it via Android Auto.

advplyr:
This would be like the "Latest" tab right? I think adding the latest tab to android auto is a good idea.

update-freak:
yes, adding the "Latest" tab to Android Auto would be great.

aureateflux:
Except the "latest" tab sorts in reverse chronological order; there should at least be a way to change the sort of the list.

advplyr:
Wouldn't that be "oldest"?

aureateflux:
> Wouldn't that be "oldest"?

Uhh 🤔 let's see if I can get my diectics in a row here 😂

By reverse chronological order, I meant that —from top to bottom— it sorts newest to oldest. To me, chronological order should sort from oldest to newest (in the order they were published), placing the first item in the most accessible spot at the top of the list. It's entirely referential though!

Display order and play order should be selectable if this is intended to work like a queue across all subscribed podcasts.

advplyr:
The latest page is sorted by the pub date set on the episode, same as an RSS feed would. The most recent pub date will be on the top. Maybe your episodes don't have a pub date?

burned42:
I gave this a try at https://github.com/burned42/audiobookshelf-app/tree/android_auto_add_recent_episodes (mostly out of curiosity to see if I would be able to do it).

This adds a tab to the android auto view similar to the 'latest' tab on the normal app view.

However, I haven't worked with this code base, kotlin or android studio at all before this. So while this seems to work fine for me, it might not cover all use cases the app offers and is probably also not up to the coding standards of kotlin or this project, so at this moment I'd rather not open a PR for this branch.

But maybe/hopefully someone with more experience can take this as a base/inspiration to integrate this feature :)

ISO-B:
I can implement this once #1322 gets merged. I could also add paging for podcast episodes and order selection.

LexManos:
Hello,  since this is still open and #1322 has been merged, figured i would second this request.
My use case is driving and wanting to easily listen to pending podcasts.
[...]

PR #1322, opened 2024-09-13
Added better library browsing for Android Auto

### LINKED ISSUE #1204: [Enhancement]: Allow browsing books by author / series (and a full list) on Android Auto

right now, only a very small (random?) selection of your books is listed in Android Auto. 

I would like to be able to have the same options to browse for a book available on Android Auto that I have on the Phone app (Browse by Series, sort by latest date, and so on), with a way to basically navigate my library by selecting an author, a series, and then the specific book I'm interested in. 

Also - it would be useful if there would be a "List" view in Android Auto instead of Cover only. 

Right now, to switch between books that have not yet been started / are listed on the results in the app, I would have to pull out my phone to interact with ABS, which is a serious safety issue. 

Make Android Auto work similarly to the stock app interface

A good way to see how I would envision the implementation is the app "listen audiobook player" which has a fully browsable library on Android Auto. 

If useful, I can take a few photos when I'm in my car the next time as I would like a similar implementation to how they did it. 

Android App - 0.9.63

_No response_

ISO-B:
This should resolve issue. Fixes #1204 

Each library now has three browsing options: Library, Series, Collections. Library browsing tree is following:

 - Library
   - Shows authors, but if there is 1000 or more authors only show first letters until there is under 100 authors.
     - Series and single books for author
       - Author books from selected series
- Series
  - Show list of series, but if there 1000 or more series only show first letters until there is under 500 series to show.
    - Books in series
- Collections
  - Show collections
    - Books in collection  

Fetched data is cached using variables.


kuldan5853:
Could you maybe make the thresholds customizable in the future? 1000 / 500 are quite a bit to browse through on a car display, one might prefer to always force it to the Alphabetical drawdown and / or change those numbers. 

Not needed for a first step (looking forward to testing this!) but just as an idea for future refinement.

ISO-B:
That was good idea. I added settings for those.

kuldan5853 (approved):

kuldan5853 (approved):

kuldan5853:
@ISO-B I have tried setting up my own build environment after talking with nicholas a while back, but I just have not enough clue what I'm doing to get a successful build going - would you be willing to build me an APK of your repo (including your three pull requests) so I can test this for a bit on my car?

ISO-B:
Sure. If you are on ABS discord server send message to me and I will provide apk for testing

kuldan5853:
Hey, thanks to the awesome @ISO-B I was able to give this a test on real hardware, and what I have been seeing has been very promising. 

First, the great stuff (Library 1 has ~1.000 Books, Library 2 has ~5.000 Books):

- Library browsing works as expected, and performance was pretty great (I tested on a parking lot away from home on an LTE connection for "real world" testing), and I'd say that the responsiveness of the library browsing etc. was great, even better than using the web version for my big library on my PC - Android Auto seems to do a lot of caching and prefetching, and it works great. 

- I initially enabled the new paginated results and they work fine, but since performance was so good I ended up disabling it and just used the Android Auto pagination instead

And here's my findings for improvement:

- Podcasts currently do not work / are not supported

- Libraries with "Books/Comics" are listed but have no entries

- The Series view currently does not use the Sequence Numbers of Books for sorting, it's A-Z only

I had a quick chat with ISO-B about this, and it seems all three of these issues should be pretty easy to fix. 
There was also the idea of adding asc/desc sort orders to the views to make navigating big libraries easier (desc = newest in a series on top).

Overall, very happy with what I've seen and I really hope after a few tweaks this will make it into main.

kuldan5853:
Found a bug (I think) - when you open the app / connect to Android Auto while the phone has limited connectivity / the app can't reach the server, you get stuck on a "Downloads" page without any way to get anywhere else. 

While in this state, the app is still responsive (on the phone), so I could for example start playback via the phone and also got the "now playing" to display on AA - but there was no way to get back to the navigation bar. 

PXL_20241013_174037021 MP

ISO-B:
> Podcasts currently do not work / are not supported

I accidentally removed podcast browsing. Probably managed to get it back, but can't say before I find time to test.

> Libraries with "Books/Comics" are listed but have no entries

Made fix to this using data from library stats endpoint.

I will first test these and after that I will commit to git and provide new test build for Kuldan

ISO-B:
We have been doing quite lot of testing with @kuldan5853. Libraries without audiobooks are now hidden from Android Auto. Series are sorted using sequence values. There is also setting which allows using ascending or descending order for books in series. Authors weren't always on alphabetical order, but are now sorted. Podcasts are now working again. All bugs should be now fixed.

@kuldan5853 confirmed that getting stuck on downloads tab also happens on production version of app. I can take a look for that problem, but it is not related to this PR so it should be fixed with different PR.

advplyr:
I think we should remove the settings for alphabetical drawdown and force a reasonable standard.

Requesting 10,000 series is too much. Have you tested with a library that large? I think this would be a very large payload that could cause memory issues. Even 1,000 is too much I think. We need some form of pagination for the queries or just show the first 100 and try to get the search working. We could also set up an endpoint on the server to send back data we need to build the directory structure including the alphabetized directories. Building that client side means we have to load the entire library.

When I was first trying to build with android auto a few years ago I could never get the search working. I just tested now and it sort of worked for the first time but showed a bunch of empty directories that I had to look into to see if a playable item was in there.

advplyr:
I don't understand why there are 2 different number inputs for the alphabetical drawdown settings. Why would we want 2 different settings there?

kuldan5853:
> I think we should remove the settings for alphabetical drawdown and force a reasonable standard.
> 
> Requesting 10,000 series is too much. Have you tested with a library that large? 

My collection is ~1.000 Books and ~5.000 Books in 2 different libraries, and I turned off the drawdown by setting the limit to 10.000 myself (to let the Android Auto A...Z do it's magic). I only have 200 series on each library though, 

Performance for that was very fast and usable, I think with up to 1.000 series it should sitll be ok (this is all tested on 5G and on my phone). 

However, I do agree that I didn't really understand the two values in the setup and personally didn't use the alphabetical drawdown besides to test it.

ISO-B:
> I don't understand why there are 2 different number inputs for the alphabetical drawdown settings. Why would we want 2 different settings there?

Now that I think about it one would be enough.

> I think we should remove the settings for alphabetical drawdown and force a reasonable standard.
> 
> Requesting 10,000 series is too much. Have you tested with a library that large? I think this would be a very large payload that could cause memory issues. Even 1,000 is too much I think. We need some form of pagination for the queries or just show the first 100 and try to get the search working. We could also set up an endpoint on the server to send back data we need to build the directory structure including the alphabetized directories. Building that client side means we have to load the entire library.

My test library has ~4,000 books with 800 series from 2,500 authors. With that amount is seemed fast enough. Didn't notice any memory issues on server nor app.
There is two reasons why I think alphabetical drawdown is needed:
1. It is much faster and safer to use while driving.
   - It can be used with physical buttons unlike search. My car has scroll wheel that can be used to scroll through list. I can use built in A...Z option with scroll wheel, but it is pain to get to it and requires more scrolling.
   - With search I need to try to hit right letter. You can probably use voice input for search, but it seems to be hit or miss when I have books with multiple different language
   - Android Auto voice commands could also help with searching and safety, but at least for me problem is that Finnish is not supported and while using English for voice commands Finnish words are not recognized correctly.
2. Android Auto has 1000 item limit for lists. In case that library has more than 1000 authors/series/anything you can't use built-in A..Z thing as it cant jump to section that is not available in that list. We could put paging links to be first items in lists but switching between pages just to use that A..Z thingy feels bit odd vs picking first letter from list and selecting correct item from next page.


Reason for doing fetching data as it is now made was just to see if I can make browsing work for android auto. Didn't want to commit making API changes and then notice that I can't get android auto side to work. Having endpoint that does tree structure would make things easier, but it would still make same amount of data as for now code fetches all/most browsable content when started.

kuldan5853:
Ah, right...I forgot about the 1000 items limit, as none of my libraries gets over that one. In that case, I think the functionality as is is pretty decent and advantageous vs. a 1/2/3 style paginated list..

ISO-B:
Also for long run having library fully cached to app(Offline library) would be helpful for Android Auto too. If you have lot of downloaded books you can't easily browse those on AA when you don't have internet connection.

kuldan5853:
@ISO-B one more thing you might want to integrate before the code gets merged is the "recently added" section as you mentioned a while ago - that would make ist mostly feature complete I think.

ISO-B:
Status update:

General:
 - New top menu item Recent is added
 - Library caches are cleared when switching server
 - Streamlined browsing setting to single setting
 - Podcast episodes show publish date in subtitle

Search:
 - Is done using server API
 - Latest search is cache to prevent need to make new request when returning from browsable item.
 - Results are grouped by book, series, author and split by library
 - Only searches libraries with audio content

Library personalized shelves:
 - Recent books, series, authors, podcasts and episodes shelves are listed under Recent tab.  #679
 - Discovery shelves can be found under library tab from corresponding library
 
 Icons:
- Static browsable list items everywhere have hard-coded icons
- Libraries use icons that are defined on server

Start up / Initial loading:
- Initial loading first loads libraries from server. After that personalized shelves and items in progress are fetched simultaneously. Top menu items are update after every stage.
- If network connection is not available when android auto starts app tries to do initial loading again when network connection becomes available again

I will add fix for #1083 before changing PR from draft.

android/app/src/main/java/com/audiobookshelf/app/data/CollapsedSeries.kt
```diff
@@ -0,0 +1,36 @@
...
+  var sequence:String?,
+  var libraryItemIds:MutableList<String>
+) : LibraryItemWrapper(id) {
+  @get:JsonIgnore
+  val title get() = name
+  @get:JsonIgnore
+  val numBooks get() = libraryItemIds.size
+
+  @JsonIgnore
+  override fun getMediaDescription(progress:MediaProgressWrapper?, ctx: Context): MediaDescriptionCompat {
```
advplyr, line 24: I can't tell if this function is being used anywhere. I was looking because I was checking if `__SERIES__` was intentionally different than the mediaId used in the LibraryItem.getMediaDescription

advplyr:
I made the following updates:

1. Instead of loading library stats separately I added a query string on the server to include library stats. This will be in the next release.
2. Fixed the search results showing ebook only books (had to update the `checkHasTracks` function to work for both minified and expanded library items)
3. Fixed `newest-episodes` personalized shelf loading the same podcast library items multiple times
4. Updated `Ascending`/`Descending` to `Sequence Ascending`/`Sequence Descending`. I'm not sure if this setting is really necessary though.
5. Fixed an issue with alphabetic drawdown where multiple items have almost the same name. Example below:

I have 3 series that start with `The E` and no other series that start with `T`.
Previously I would have to click into `T` then `TH` then `THE` then `THE E`.
Now clicking into `T` will show all results because it checks if the next grouping has more than 1 group.

image
image

The downside of this method is `THE` is very common so depending on if the user has other `TH` series it may show a lot of series.

There are ways to improve this like looking ahead and showing the next logical group like: `T` -> `THE EN` and `THE EX`
but I didn't want to go through the process of building that out. 

It could also use the `nameIgnorePrefix` instead which by default would exclude `The`.

advplyr:
Fixed an issue where when the cache gets reset the data wasn't reloaded.
Also reset the items in progress. I found that when disconnecting from the server (using disconnect link in the app) then re-opening android auto it was using the last servers item in progress.

advplyr:
This is good to go now. The only concern I have is the amount of data being loaded. If my prod server is taking more than a few seconds to load on wifi then there are going to be users that take several minutes to load (maybe timeout?).

Some users are already having performance issues in the web client, and that is loading less data than this. I didn't spend time trying to optimize because I think we should release this first and get feedback from users. We'll be able to set up a new endpoint on the server for global library search and some other things to reduce the number of requests.

Overall this is a big improvement to android auto. Thanks for all your work. I am planning on releasing this tomorrow.

kuldan5853:
> This is good to go now. The only concern I have is the amount of data being loaded. If my prod server is taking more than a few seconds to load on wifi then there are going to be users that take several minutes to load (maybe timeout?).

I have tested the beta build for a few months now, and even on a pretty bad 4G connection I still had acceptable load times all things considered. I'm sure there is potential for optimization down the road (especially with local caching of data etc), but for now this has been very usable in my everyday life.

advplyr:
I suspect this update is related to the [heavy battery usage](https://github.com/advplyr/audiobookshelf-app/issues/1446) on android in the latest release.

`onLoadChildren` is called anytime media is played even when not using android auto. The potentially 10+ http requests that were added here get called when the first media is played no matter what, then subsequent plays just refresh the BrowseTree.

The reason this isn't only run on android auto is because of a [change I made](https://github.com/advplyr/audiobookshelf-app/commit/28989f536a4b0aa2a3f00817a69ebe47df51a45d#diff-ab99ecf842635442352cc075a7d94bbac4e5f2344e40b723be302a9f628a4f0aR917) to the `VALID_MEDIA_BROWSERS` when building the widget. Specifically, I added `com.android.systemui` which is the service that is calling `onLoadChildren` during regular non-android auto playback. I believe I added that service when I was trying to get the widget to be able to play with the app closed completely which I was never able to get working properly. From my testing removing that still allows the widget to function as it was.

This wouldn't have been an issue before (although maybe somewhat) because the `onLoadChildren` function wasn't loading much data.

I'm not positive this is the cause of the heavy battery usage because the last release also updated many dependencies and Capacitor to v6.
What I'm planning on doing now is removing `com.android.systemui` and upgrading to Capacitor 7 (just released a few days ago).

ISO-B:
Good to know. Please update us if that doesn't solve the problem. Then I can try to find a better/other way to handle Android auto stuff.

advplyr:
That only solved the problem for users that don't use android auto. Battery drain is a common complaint so we're going to need to handle this better.

PR #1357, opened 2024-10-29
Fixed app crashing when switching user/sever while playing audio

### LINKED ISSUE #1336: [Bug]: Accessing "Switch Server/User" while audio is playing closes app
### What was the Problem?

One thing that I noticed is that while audio is playing, accessing the "Switch Server/User" menu causes ABS app to close (crash?). This does not occur if audio is not playing.

1. Start playing audio from an audiobook
2. Access the Switch Server/User

### What was Expected?

WHAT IS EXPECTED: The audio may be paused, but the Server list should come up.
WHAT REALLY HAPPENS: The Android App closes

### Phone Model

Pixel 9 Pro

### Phone OS

Android 14

Android App - 0.9.76

### Installation Source

Google Play Store

### Additional Notes

_No response_

ISO-B:
Added pausing play and closing stream commands before logout. Fixes #1336

### DIFF
--- a/components/app/SideDrawer.vue
+++ b/components/app/SideDrawer.vue
@@ -171,6 +171,10 @@ export default {
     },
     async logout() {
       if (this.user) {
+        if (this.$store.state.playerIsPlaying) {
+          this.$eventBus.$emit('pause-item')
+          this.$eventBus.$emit('close-stream')
+        }
         await this.$nativeHttp.post('/logout').catch((error) => {
           console.error('Failed to logout', error)
         })


nichwall (approved):
Looks good to me!

ZLoth:
Thank you.

advplyr:
This didn't explain why the crash was happening and why this fixed it. 

I found that the crash was happening due to the server connection config being nulled before the `/close` session request was getting sent to the server. The exception was getting thrown when building the OKHttp Request because there was no server address in the url.

I was confused as to why the `pause-item` event was required before the `close-stream` event in this PR. I think that was just a coincidence. The `/close` request was getting sent out before the server connection config got nulled but it was still possible that it wouldn't. A possible race condition in the future.

To fix these issues I first wrapped the OKHttp Requests in a try catch because these requests shouldn't crash the app. I then added the `close-stream` event before logout, and in the `closePlayback` Kotlin function I kept a copy of the server connection config first so that even if it gets nulled the final sync & close requests will still get sent to the server.

https://github.com/advplyr/audiobookshelf-app/commit/d902417959e286a6c13a43fb8979e63d721922e5

PR #1358, opened 2024-10-29
Ensure that there is files available before playing local content

### LINKED ISSUE #105: Delete Local Audiobook In Use Causes App To Be Unusable
### Steps to reproduce

1. Download an Audiobook
2. Play Local, then Stop. 
3. Delete Audiobook files with other file manager (X-plore for example)
4. Open App

### Expected behaviour
- Player should check that files are available to play. If not, remove queued book and close player.

### Actual behaviour
- App completely crashes on open.

### Environment data

- [X] Android App
- [ ] iOS App? LOL no

#### Android Issue
Android version: 0.9.40-beta

Device model: Z Flip 3

ISO-B:
This should fix #105 . This wont work 100% with books that have multiple books. 

If there is way to know how many files there should be then this can easily be extended for multi file books.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/data/LocalLibraryItem.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/data/LocalLibraryItem.kt
@@ -79,6 +79,16 @@ class LocalLibraryItem(
   }
 
   @JsonIgnore
+  fun hasTracks(episode:PodcastEpisode?): Boolean {
+    var audioTracks = media.getAudioTracks() as MutableList<AudioTrack>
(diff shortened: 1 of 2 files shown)


advplyr:
This only works after re-opening the app because on first load the `cleanLocalLibraryItems` function gets ran that updates the LocalLibraryItem data removing any files that are no longer there.

ISO-B:
Added check that files really exists. Now it prevents playing book if file is removed while app is running.

advplyr:
Thanks! I wanted to push an additional update in this PR but I don't think permission was given.

Typically this is shown.
image

but it wasn't shown on this PR.

PR #1359, opened 2024-10-31
Changed network connection check logic

### LINKED ISSUE #566: Unable to play off local server without internet connection

ISO-B:
Network connection no longer requires internet connection. Socket connection status is used instead for checking if server is reachable. If there is no socket connection available eq. before connecting to server then connection type is used to for network connection check.

This should fix #566

### DIFF
--- a/store/index.js
+++ b/store/index.js
@@ -31,16 +31,16 @@ export const state = () => ({
 })
 
 export const getters = {
-  getCurrentPlaybackSessionId: state => {
+  getCurrentPlaybackSessionId: (state) => {
     return state.currentPlaybackSession?.id || null
   },
-  getIsPlayerOpen: state => {
+  getIsPlayerOpen: (state) => {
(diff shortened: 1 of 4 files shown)


advplyr:
Thanks!

store/index.js
```diff
@@ -165,7 +165,11 @@ export const mutations = {
     state.isNetworkListenerInit = val
   },
   setNetworkStatus(state, val) {
-    state.networkConnected = val.connected
```
advplyr, line 168: I'm confused about this change actually. Was `val.connected` false in your testing when `connectionType` was not `none`?

PR #1360, opened 2024-11-03
iOS devices have always networkConnected status true

ISO-B:
#1359 fixed things only for Android app. Capacitor Network plugin only shows iOS device connected if internet access is available. This fix allows iOS users to use local server without internet access. Socket is used to detect if connection to server is availabe.  networkConnected is only used for add server form and cellular permission check.

Cellular permissions for download and streaming wont work for iOS if device is connected to cellular, but without internet access to server that is used for connectivity check.


advplyr:
Thanks!

PR #1382, opened 2024-11-16
Hebrew translation

arcmagedr:
Add Hebrew translation strings json for the app


arcmagedr:
Just saw that translation is now being done using weblate 😅
Should I close and continue there?

advplyr:
This should work and sync with weblate.
Are the unverified commits intentional?

arcmagedr:
I can probably fix it and do a final commit if that’s important

advplyr:
Since you made this PR someone added translations through weblate. Can you resolve the conflicts?

Issue #1374, opened 2024-11-11
[Bug]: E-book reader settings not all visible (iOS)


thehijacker:
### What was the Problem?

I can not see in full all the optoions on e-book setting dialog.

![photo_5879506810715620966_y](https://github.com/user-attachments/assets/75b20a54-25f2-406f-bce5-8300b8720d79)

They are cut at bottom and I can not scroll down to see them all.

### Steps to Reproduce the Issue

1. Open e-book.
2. Open settings.

### What was Expected?

To see all settings options.

### Phone Model

iPhone 8

### Phone OS

iOS 16.7.10

### Audiobookshelf App Version

iOS App - 0.9.77

### Installation Source

Testflight

Issue #1365, opened 2024-11-06
[Enhancement]: Volume button page turn while audiobook is playing


fawazral:
### Describe the Feature/Enhancement

https://github.com/advplyr/audiobookshelf-app/pull/1302 added the ability to change the page using volume keys. When an audiobook is playing, the page does not turn (which was done intentionally). 

Can there be an option to always have the volume key be the page turn, even when an audiobook is playing? 

### Why would this be helpful?

When I read on my tablet, I control the volume using gestures and exclusively use the volume keys for page turning. 

### Future Implementation (Screenshot)

Maybe an additional setting after the "Navigate with volume keys" setting that is called "Navigate with volume keys during playback: Yes/No"

### Audiobookshelf App Version

Android App - 0.9.76

PR #1383, opened 2024-11-17
Turn page while playing

RaHoni:
This is a fix for #1365. This adds an additional settings that disables the check for playing when enabled.
The default for this setting is off.

But since while testing i noticed that the problem for #1374 was also showing up on the Virtual Pixel 7a. The simple fix was removing the half-screen option. @advplyr this does change the Settings to be full-screen. Is this okay or is there something else I should do?


RaHoni:
@fawazral would you want to test this PR?

fawazral:
@RaHoni I just tested it out and it works as intended! Thanks so much for doing this!

advplyr:
I didn't catch this before but I think we can use the same "On" and "Off" translation for the settings. If it turns out there are some languages that need them separated for whatever reason we can do that, but if we can re-use translation strings we should.

I updated the label for the new ereader setting to be more clear so you'll have to translate that again in German

RaHoni:
@advplyr I fixed the Translation if you want you can merge now

advplyr:
Thanks!

PR #1387, opened 2024-11-21
Feature: Option to keep screen on while reading

### LINKED ISSUE #1207: [Enhancement]: Keep screen on when reading a ebook

Currently, if the screen lock timeout goes off before I’ve finished reading a “page” my phone/ipad will lock the screen. This requires me to touch the screen periodically to prevent this. It’d be nice if the app automatically prevented screen locking while in the in ebook reader. 

It often takes me longer to read a “page” than the screen lock timeout on my phone/ipad. Keeping the screen on when reading an ebook would prevent this.

image

Android App - 0.9.63, iOS App - 0.9.63

_No response_

ISO-B:
Hard work is done using capacitor-community/volume-buttons plugin. Option is disabled by default, but can be enable from Ereader settings. Should work on iOS and Android. This should resolve #1207 

Tested with OnePlus One, Android 11.


advplyr:
Implemented this here https://github.com/advplyr/audiobookshelf-app/pull/1439
Thanks!

PR #1403, opened 2024-12-06
upgrade to Capacitor v6, fix ios build

DrNgo:
Made necessary changes in accordance to directions for the upgrade noted [here](https://capacitorjs.com/docs/updating/6-0)
- Upgrade `@capacitor` dependencies to latest 6
- Register custom plugins for iOS
- Update Android project versions to new minimums
- Fix Android build error with JVM Mismatch ignore


DrNgo:
rebased on current master

PR #1415, opened 2024-12-20
update gradle an 8.11 for android

Vincenzoferrara:
buildDIr is deprecated > rootProject.layout.buildDirectory


advplyr:
I just took care of this as well as a bunch of other issues when migrating to Capacitor 6. Thanks!

PR #1417, opened 2024-12-23
Support servers with subdirectory

mikiher:
This adds support for connecting to servers with subdirectory, e.g. `https://my.server.com/audiobookself`.

This is the last step in preparing for subdirectory support, based on [the proposed plan](https://github.com/advplyr/audiobookshelf/discussions/3535).

The required changes include:
- Allow paths when specifying the server address in the Connect form
- Fix `this.player.src` in `AbsAudioPlayer.js` (all other changes in this file are due to formatting)
- Connect to the correct WebSocket path
- Fix `getLibraryItemCoverSrc` and `getLibraryItemCoverSrcById` in `globals.js`

I tested this using the web client, as well as on my Android phone against a server running with `RouterBasePath=/audiobookshelf` (which should support both subdirectory and non-subdirectory access). 
- I connected using the standard non-subdirectory address, and made sure everything was working
- I then connected using the subdirectory address, and made sure everything was working.

I _did not_ test on iOS, as I don't have an IOS dev environement nor an iOS phone to test on.

### DIFF
--- a/components/connection/ServerConnectForm.vue
+++ b/components/connection/ServerConnectForm.vue
@@ -486,9 +486,8 @@ export default {
     validateServerUrl(url, protocolOverride = null) {
       try {
         var urlObject = new URL(url)
-        var address = `${protocolOverride ? protocolOverride : urlObject.protocol}//${urlObject.hostname}`
-        if (urlObject.port) address += ':' + urlObject.port
-        return address
+        if (protocolOverride) urlObject.protocol = protocolOverride
+        return urlObject.href
       } catch (error) {
         console.error('Invalid URL', error)
         return null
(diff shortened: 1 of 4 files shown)


advplyr:
I tested this on Android/iOS/web including android auto. Works great, thanks!

vb0:
Just to be clear the subdirectory is hard coded to audiobookself in the Android app, or is there generally some reason not to work (as I tested it and it's accepted but then it can't find the server) due to how the relative path is handled in some way?  

The use case is proxying https://my.server.com/a0d1bsh23 -> http://192.168.1.2:13378 - which doesn't work now
while 
https://my.server.com -> http://192.168.1.2:13378 of course works
https://my.server.com/audiobookshelf -> http://192.168.1.2:13378 works since a while

This is useful because a0d1bsh23 (of course something of this sort, not this particularly) is a good authentication token, sure it's fixed and known to all your infrastructure but unknown to any generic attacker and they can't find it being inside TLS.  

Note that (as far as I can tell) no support should be needed from the server, which would get the URL without the token part.

mikiher:
the `/audiobookshelf` path is hardcoded in the server and the web client, not in the Android app. The hardcoding is a workaround implemented in the constraints of the current architecture, not a full solution that enables any arbitrary path.

PR #1418, opened 2024-12-23
Update cap6, changed to a branch 

DrNgo:
Made necessary changes in accordance to directions for the upgrade noted [here](https://capacitorjs.com/docs/updating/6-0)
- Upgrade `@capacitor` dependencies to latest 6
- Register custom plugins for iOS
- Update Android project versions to new minimums
- Fix Android build error with JVM Mismatch ignore


advplyr:
I ended up doing this locally so I could follow the migration guide step-by-step and make sure I understood the updates. I updated the android dependencies to more recent versions and set compileSdkVersion to 35.

For the Android build error I ended up following the gradle [recommendation](https://docs.gradle.org/current/userguide/toolchains.html) of using a toolchain. I'm not positive this is going to stay because I'm still learning about it, but it is working well.
Thanks!

Issue #1399, opened 2024-11-28
[Bug]: Sleep timer "End of Chapter" incorrect at different playback speed


MadDogTen:
### What was the Problem?

When using the "End of Chapter" button to set the sleep timer when playing at a faster speed, it sets it to the time for the normal playpack speed.

### Steps to Reproduce the Issue

1. Play any audiobook at faster speed
2. Use the "End of Chapter" button to set the sleep timer.

### What was Expected?

It should set the sleep timer based on the current playtime end of chapter, accounting for different playback speed. 

Example- End of Chapter is 10 minutes at 1x speed, Sleep timer "End of Chapter" sets it to 10 minutes. If playing that same 10 minute chapter at 2x speed, The sleep timer "End of Chapter" should be set to 5 minutes.

### Phone Model

Pixel 6 Pro

### Phone OS

Android 15

### Audiobookshelf App Version

Android App - 0.9.77

### Installation Source

Other (List in "Additional Notes")

### Additional Notes

Installed via APK from this repository.

nichwall:
I was able to reproduce this on Android app 0.9.77-beta

danielb2:
I can't reproduce this. The timer displays the correct time for the normal playback speed, but the timer also counts down at a faster/slower pace depending on the playback speed. at the end of the day, it does stop at end of chapter

MadDogTen:
> I can't reproduce this. The timer displays the correct time for the normal playback speed, but the timer also counts down at a faster/slower pace depending on the playback speed. at the end of the day, it does stop at end of chapter

That is honestly strange, as it didn't work that way previously, and I have been avoiding the app because of it. I suppose it's possible I only checked the time and didn't let if fully play for this version, but I could swear I did.

Regardless, It still should display the proper sleep time, not count at a different speed.

Issue #1268, opened 2024-07-25
[Bug]: Autosleep timer engaging outside of set time.


iconoclasthero:
### What was the Problem?

The sleep timer was set (1200-0600, later changed to 0000-0600).  Playback was started ca. 11:00, outside of either time range.  The sleep timer was still active.

### Steps to Reproduce the Issue

1. Set sleep timer
2. Start playback
3. Observer sleep timer on.

### What was Expected?

Inactive sleep timer

### Phone Model

OnePlus 10 Pro

### Phone OS

Android 14, stock rom, rooted

### Audiobookshelf App Version

Android App - 0.9.74

### Installation Source

Other (List in "Additional Notes")

### Additional Notes

I probably sideloaded it last time?  Not 100% sure.  I know that I _have_ sideloaded it before.

iconoclasthero:
![Screenshot_2024-07-25-11-30-55-9 (1)](https://github.com/user-attachments/assets/67b2e7b8-2ed4-4b6c-a414-cfeea9783525)
![Screenshot_2024-07-25-11-30-39-6 (1)](https://github.com/user-attachments/assets/5c5143d0-8ba4-4824-aa7c-782798b47c4f)

nichwall:
I was able to reproduce this by setting an auto sleep timer to end 3 minutes in the future and beginning playback (observed auto sleep timer enabled correctly). I closed the media player but kept the app open until I was outside of active hours for the auto sleep timer. Opening the player incorrectly enabled the auto sleep timer with full time remaining.

Completely closing the app and reopening the app correctly did not cause the auto sleep timer to be enabled. Looks like once the auto sleep timer is enabled it is not checked whether it should be disabled without completing closing the app.

iconoclasthero:
Yeah, what he said about completely closing the app seemed to work.  I used app info > force stop to close it.
Changing books, logging into the server didn't have any effect.

Issue #1200, opened 2024-05-11
[Enhancement]: Add option to disable auto sleep timer temporarily


wnhrt:
### Describe the Feature/Enhancement

An option within the audio player to temporarily deactivate the automatic sleep timer (e.g. for the rest of the day, for the next _x_ hours or for the current listening session). The timer should then be reactivated automatically.

### Why would this be helpful?

Some evenings I want to listen for longer without shaking my phone every _x_ minutes. At the moment I have to disable the auto sleep timer, go into the settings, scroll to the bottom and enable the auto sleep timer again. Automating this process would be convenient.

### Future Implementation (Screenshot)

A button next to or instead of the existing one

![Screenshot_20240511-220005](https://github.com/advplyr/audiobookshelf-app/assets/70302993/a081733b-3ccf-4c12-bb2f-869ae266d1f3)

### Audiobookshelf App Version

Android App - 0.9.63

Dr-Blank:
have you considered adding those many hours to your sleep timer instead?

PR #1453, opened 2025-01-25
Android sleep timer cleanup

nichwall:
## Brief summary

This PR scales the sleep timer display by the playback speed when using the End of Chapter timer on Android.

## Which issue is fixed?

Fixes https://github.com/advplyr/audiobookshelf-app/issues/1399 (first 3 commits)
Fixes https://github.com/advplyr/audiobookshelf-app/issues/1268 and fixes https://github.com/advplyr/audiobookshelf-app/issues/1200 (4th commit)

## In-depth Description

This PR updates the `SleepTimerManager.kt` in 3 commits. The first commit is just applying the autoformatting and adding function comments so it is easier to see what is going on.

The second commit simplifies the Increase and Decrease time functions by using built in functions to select max/min of values.

The third commit (which actually fixes the issue) adds scaling to time remaining functions when the "end time" for the chapter is used instead of "time remaining" for the fixed length.

The fourth commit cleans up general function calls to reduce code duplication and simplify the logic.

## How have you tested this?

Built and tested on Android 15, Pixel 6a.

1. Set playback speed
2. Enable sleep timer for fixed length
3. Change playback speed while fixed length timer is running, verify that time does not change
4. Disable sleep timer
5. Enable sleep timer for End of Chapter, verify time remaining is scaled by playback speed
6. Change playback speed, verify time remaining updated to match

This was repeated for various playback speeds, fixed lengths, and different chapters.

## Screenshots

image
image


nichwall:
This new commit does more general cleanup of the sleep timer for Android. It also addresses https://github.com/advplyr/audiobookshelf-app/issues/1268 (sleep timer not disabling outside of active hours) and https://github.com/advplyr/audiobookshelf-app/issues/1200 (only disable auto-sleep timer for the rest of the day).

advplyr:
I tested this pretty thoroughly and it's working well. I removed the separate string for disabling auto sleep timer on ios because auto sleep timer isn't supported yet there and can be implemented the same way when it is.
The code cleanup is much appreciated, thanks!

nichwall:
Ah, thanks. I wasn't sure about the iOS functionality, so good to know that it's Android only for the auto sleep timer.

PR #1461, opened 2025-01-30
Build apk workflow update

nichwall:
This PR updates the Android build workflows to use `upload-artifact@v4` due to `@v3` being deprecated on January 30, 2025 (which is tomorrow at the time of this writing). The `upload-artifact` changes can be seen in this GitHub blog post: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

The `upload-pages-artifact` is also updated according to https://github.com/actions/upload-pages-artifact to ensure that is on the most up to date version. This workflow updates the page hosted at https://advplyr.github.io/audiobookshelf-app/. I wonder if we still need this page because it is not really advertised anywhere and the APK can be downloaded from the workflow itself. However, the page is very nice and looks professional, but not sure about maintaining it if it breaks. A screenshot of the page is shown below.
APK download page

### DIFF
--- a/.github/workflows/build-apk.yml
+++ b/.github/workflows/build-apk.yml
@@ -13,41 +13,41 @@ jobs:
   main:
     runs-on: ubuntu-latest
     steps:
-    - name: checkout sources
-      uses: actions/checkout@v3
+      - name: checkout sources
+        uses: actions/checkout@v3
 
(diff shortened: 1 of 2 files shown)


advplyr:
Thanks! I forgot that page existed. I suppose it is unnecessary in that case

Issue #1463, opened 2025-02-02
[Bug]: Audio resumes from pause when phone is shaken while timer is active


Zeoic:
### What happened?

A couple of days ago, I noticed this issue start. Must have been around the time my app updated to 0.9.79-beta.

The sleep timer doesnt seem to stop after elapsing or leaving the set time window for auto timer. It will continue to respond to the shake to extend feature all day. Additionally, the shake to extend feature is also resuming the paused media.

Note: It seems that leaving the auto timer active and manually turning off the active timer will prevent it from responding to shake again.

App version: 0.9.79-beta
Server version: v2.18.0
Phone Model: Samsung Galaxy Z Fold 5
Android Version: 14

### What did you expect to happen?

Typically when using this feature, the timer elapsing would turn the timer off, with shake to extend no longer activating. 

Additionally, when outside the time window set for auto timer, the timer didnt used to start / respond to shake.

### Steps to reproduce the issue

1. Enable auto sleep timer
2. Listen to ausio in time range
3. Let timer elapse
4. Shake phone
5. Audio resumes
6. Pause audio
7. Shake phone
8. Audio resumes again

### Audiobookshelf version

v2.18.0

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell

```

Strider27:
Same here, I though I was going crazy.

V 0.9.79-beta
Galaxy S24
Android 14

nichwall:
I was able to reproduce the issue of playback resuming by shaking after a manual pause with both the normal and auto sleep timer.

The shake to resume after the sleep timer ended was correctly disabled 2 minutes after ending a sleep timer (user has 2 minutes to shake to restart the current timer).

vb0:
I can reproduce this as well, thought I was going crazy too. Before it was working fine, well I didn't even realize there's a 2 minutes grace period you can shake it again to resume (I see now it's even documented in the UI), that's a good kink that should be left in.

vb0:
[...]

Issue #1097, opened 2024-02-23
[Bug] Sleep Timer "End of Chapter" Android App


DavidPesticcio:
### Steps to reproduce

1. Select a book
2. Select Play
3. Select Sleep
4. Select "End of Chapter"
5. Listen until the end of chapter is reached
6. Wait for player to stop
7. Press Play - does not advance to next chapter possibly due to "rewind" action at end

### Expected behaviour
- While playing within a chapter, and many minutes from the end of that chapter, selecting the sleep timer to finish at "End of Chapter" the player will stop at end of chapter, but will not continue upon pressing play after playback has stopped.

### Actual behaviour
- The player rewinds a few seconds after stopping, pressing play only plays those few seconds up to the end of the chapter, it does not advance to the next chapter as expected.

### Environment data
Audiobookshelf Version:
Server: v2.8.0
App: 0.9.72-beta

- [x] Android App?
- [ ] iOS App?

#### Android Issue
Android version: 12

Device model: Pixel 6 Pro - Stock ROM

DavidPesticcio:
See also: https://github.com/advplyr/audiobookshelf-app/issues/985

iconoclasthero:
Did the end of chapter option go away in 0.9.74?

Issue #1047, opened 2024-01-17
Auto Sleep Timer Goes Negative When Adjusted Manually and Then Refuses to Play


doctordarko:
### Steps to reproduce

I was looking for a way to cancel the auto sleep timer without disabling it.

1. When timer is at less than 5 minutes, decreasing the timer manually results in a negative number (see attached), which then stops the playback.
2. Pressing play again appear to set the timer for 1 second and then stops again.
3. Setting a custom time or a large time delay is the only way to restart the playback.
<img width="115" alt="image" src="https://github.com/advplyr/audiobookshelf-app/assets/57649463/033b75b0-6d60-4baa-ac4c-70f38063e147">

### Expected behaviour
- When timer is decreased manually, it should not go negative but zero and stop the playback as expected.
- Pressing play again should either (a) restart the auto sleep timer as per the settings or (b) the consecutive sleep timer trigger could bypass the auto sleep settings and restart without setting the timer.
- Regardless of which option restarts the play, there is a bug with the timer currently.

Audiobookshelf Version:

Android App 0.9.71-beta
Pixel 7 Pro Android 14

PR #1464, opened 2025-02-02
Android Sleep Timer cleanup part 2

nichwall:
## Brief summary

This PR continues to clean up the Android sleep timer, and fixes an issue introduced by my first PR for cleaning up the sleep timer (https://github.com/advplyr/audiobookshelf-app/pull/1453).

## Which issue is fixed?

Fixes https://github.com/advplyr/audiobookshelf-app/issues/1463 (first commit)
Partially fixes https://github.com/advplyr/audiobookshelf-app/issues/1097 (second commit)
Fixes https://github.com/advplyr/audiobookshelf-app/issues/1047 (third commit)

## In-depth Description

The first commit adds a check of whether audio is playing or if we are in the 2 minute grace period before resetting the sleep timer due to a shake. This prevents pausing and then setting the phone down from restarting playback.

The second commit changes the threshold to go to the end of the *next* chapter to be 10 seconds away from the end of the current chapter to help account for auto-rewind preventing the sleep timer from getting stuck selecting the same chapter. I am not sure if the threshold should be higher than this, because I could see the benefit of having it set to the maximum rewind time, but that may be too long when getting to the end of a chapter.

The third commit adds a check to whether `sleepTimeLength` is the minimum value of 1 second to clear the timer. This occurs when the user tries to decrease the remaining time below 0 seconds remaining.

## How have you tested this?

Built and tested on Android 15, Pixel 6a.

Tested by:
- Setting auto sleep timer and manual sleep timers. Shaking with media paused, not started, and playing.
- Manually reducing time past zero, pressing play (resumes correctly) or shaking (does nothing correctly).
- Used end of chapter sleep timer, waited for it to finish, then shook to continue on to next chapter (Without using the auto-sleep timer rewind, just the normal auto rewind)

## Screenshots
N/A

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/SleepTimerManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/SleepTimerManager.kt
@@ -264,7 +264,7 @@ constructor(private val playerNotificationService: PlayerNotificationService) {
   }
 
   /**
-   * Gets the chapter end time for use in End of Chapter timers. If less than 2 seconds remain in
+   * Gets the chapter end time for use in End of Chapter timers. If less than 10 seconds remain in
    * the chapter, then use the next chapter.
    * @return Long? - the chapter end time in milliseconds, or null if there is no current session.
    */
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks!

Strider27:
Legend!

PR #1462, opened 2025-02-02
Andriod `DeviceManager.kt` cleanup

nichwall:
## Overview
This PR cleans up the `DeviceManager.kt` file.

## Detailed description
- No functional changes were made
- Code was autoformatted
- Javadoc-style comments were added
- Some syntax cleanup, particularly in the default device settings
- Fixed `autoSleepTimerStartTime` being set twice instead of `autoSleepTimerEndTime`

## How was this tested?
Tested on Android 15 with a Pixel 6a. I did not test with downgrading and upgrading the versions.

PR #1468, opened 2025-02-07
Devicemanager settings revert

nichwall:
This PR fixes an issue introduced by my original rewrite in https://github.com/advplyr/audiobookshelf-app/pull/1462.

Some of the settings would be reset upon restarting the app after more testing.


advplyr:
Thanks!

PR #1471, opened 2025-02-08
Remove local only android

nichwall:
## Overview
This PR cleans up a lot of the "local only" related logic from Android. The ABS Android app removed support for media which was not downloaded directly through the app (i.e. scanning local media from the file system) in app version `0.9.67-beta`, released November 2023. "Local only" media is defined as media which is not linked to an ABS server.

The front-end is updated to not include local only checks or displays, and the back-end cleans the database of local only entries when initialized and checking for missing files to allow "local only" checks and logic to be removed.

Local progress for media downloaded from the server is still tracked when connected to a different server, as a different user, or when not connected to any server.

## Detailed description

Notable commit summary:
1. Remove local only items from front end (both checks and the download icon)
2. General auto-formatting of files, no other code changes
3. Delete any local only files in the Paper DB when the database is initialized. I have never had any local only items, so this is only tested in that case.
4. Remove creation and unused `isLocalOnly` from playback sessions and media progress
5. Removing unused functions which were only used to get chapter metadata from local only files. These functions are not called from anywhere.
6. Auto-formatting `FolderScanner` in preparation for commit 7
7. Remove unused `AudioProbeResult`. This class is only used when scanning local only files, so the class was removed as well as all references. The references were not actually used anywhere else, but the data did exist in `AudioTrack`, referenced in `FolderScanner`.

## Testing
This was tested on a Pixel 6a using Android 15.

I have never had local only media and I am not sure how to add this sort of test data into the database within the app easily, so I am not 100% sure that the deletion is in the correct place. I am currently deleting the local only items during database initialization, which I believe would happen before anything else tries to load data. Because this functionality was removed over a year ago I don't expect the changes to the database to affect many people either.


nichwall:
I have updated the main comment to summarize my commits. Additional verification would be appreciated before merging this.

advplyr:
I don't see any issues with adding the delete there. I don't have any local only items anymore either and I don't think it is necessary to add them since this looks straightforward. I did some testing with all the basic download functionality.
Thanks!

Issue #1276, opened 2024-07-31
[Enhancement]: Change chromecast connected icon


Silther:
### Describe the Feature/Enhancement

use the filled out [Chromecast connected icon](https://mui.com/material-ui/material-icons/?query=cast&selected=CastConnected) instead of a green Cast icon.

### Why would this be helpful?

On black and white screens the green icon is not very recognizable 

### Future Implementation (Screenshot)

![image](https://github.com/user-attachments/assets/e53b1e06-382c-4165-835e-f56a5d14c936)

### Audiobookshelf App Version

Android App - 0.9.74

### Current Implementation (Screenshot)

![image](https://github.com/user-attachments/assets/adfc68c1-b7ef-44b0-abc2-796a46a0f247)

PR #1475, opened 2025-02-09
Use `cast-connected` icon instead of green icon

nichwall:
This PR fixes https://github.com/advplyr/audiobookshelf-app/issues/1276. Only the frontend was changed.

## Images
App bar without casting
image

App bar while casting
image

Player while casting
image

## Testing
This was tested on a Pixel 6a using Android 15.


advplyr:
Thanks, looks much better. On light covers it was using successDark for both casting and not so I updated it to use black and I think it looks good.

Not casting:
image

Casting:
image

PR #1499, opened 2025-02-28
Add question to confirm they have checked for awaiting release

nichwall:
This PR adds a question to beginning of the bug report form to ask users to check the `awaiting-release` tag, along with a link.


advplyr:
Nice idea, thanks!

Issue #831, opened 2023-08-03
[iOS] - Audio not fading out aat the end of the sleep timer.


tehguitarist:
### Steps to reproduce

1. Set a sleep timer
2. Wait til sleep timer gets to the end
3. Observe that it just cuts off

### Expected behaviour
- The audio should fade out after the timer expires (and restarting audio should go back to the fade start time).

### Actual behaviour
- Audio ends abruptly

### Environment data

Audiobookshelf Version:

- [ ] Android App?
- [x] iOS App? v0.9.65

#### iOS Issue
iOS Version:
16.5.1

iPhone model:
13

hobesman:
Dues the iOS app have this setting?

![Screenshot_20230808-160600_audiobookshelf](https://github.com/advplyr/audiobookshelf-app/assets/29721765/2d750d77-835d-4833-ae58-46a91ffa57b1)

tehguitarist:
It’s actually got far less. Might be an unimplemented feature then for iOS
![IMG_7546](https://github.com/advplyr/audiobookshelf-app/assets/28495826/31a1e43b-57df-465e-8e14-cd85615d645b)

PR #1508, opened 2025-03-09
IOS: Add support for fading out on sleep timer end

complacentsee:
This addresses issue #831 to bring additional parity to the iOS app.
Playback will start to fadeout during last 60 seconds of the sleep timer. Once faded out, playback will be paused, volume reset, and playback seeked to start of fadeout.


advplyr:
Great, thanks!

Issue #1447, opened 2025-01-22
[Enhancement]: Smart rewind


pickles4evaaaa:
### Describe the Feature/Enhancement

I see that the Android app has this feature but iOS does not. 

Basically: If you pause your book for more than 10 minutes and then resume play, the playback will rewind 10-30 seconds.

### Why would this be helpful?

The listener is able to get a quick recap of where they are at in the book after they have been paused for more than 10 minutes. 

### Future Implementation (Screenshot)

The app "Bound" on iOS has this feature

### Audiobookshelf App Version

iOS App - 0.9.78

complacentsee:
The code for this functionality is largely implemented on iOS and has been in place for some time. However, a bug has been identified in the function. 

## Description

- **Location:** The bug is in the [PlayerTimeUtils.swift](https://github.com/advplyr/audiobookshelf-app/blob/master/ios/App/Shared/player/util/PlayerTimeUtils.swift) file.
- **Issue:** The function `timeSinceLastPlayed` returns a negative time interval, but the related function `timeToSeekBackForSinceLastPlayed` expects a positive number.


## Original Code

```swift
static internal func timeSinceLastPlayed(_ lastPlayedMs: Double?) -> TimeInterval? {
    guard let lastPlayedMs = lastPlayedMs else { return nil }
    let lastPlayed = Date(timeIntervalSince1970: lastPlayedMs / 1000)
    return DatelastPlayed.timeIntervalSinceNow
}
```

## Corrected Code

```swift
static internal func timeSinceLastPlayed(_ lastPlayedMs: Double?) -> TimeInterval? {
    guard let lastPlayedMs = lastPlayedMs else { return nil }
    let lastPlayed = Date(timeIntervalSince1970: lastPlayedMs / 1000)
    // Return a positive elapsed time.
    return Date().timeIntervalSince(lastPlayed)
}
```
Lastly the code for disabling the feature is mostly completed, but the setting isn't enabled in the iOS GUI build, and the settings variable isn't use. 

Not sure if there is more history to this. But I'll submit a PR soon. 

I'll submit a PR for this

pickles4evaaaa:
@complacentsee This is so exciting! Thanks :)

PR #1512, opened 2025-03-11
IOS: Support autoRewind and disableAutoRewind setting

complacentsee:
This addresses issue #1447. Most of the code to support this was added in 2022, but in practice it isn't working as intended, it looks like the tests didn't test that the resultant code when used completely would result in the correct time, just components that passed tests correctly.

This request does the following:
1. Updates the PlayerTimeUtils handle time intervals correctly.
2. Updates the test suite to test the PlayerTimeUtils.calcSeekBackTime at all ranges.
3. Enables the setting for disable auto rewind in iOS UI.
4. Updates PlayerTimeUtils such that the disable auto rewind setting is used.

### DIFF
--- a/ios/App/AudiobookshelfUnitTests/Shared/player/util/PlayerTimeUtilsTests.swift
+++ b/ios/App/AudiobookshelfUnitTests/Shared/player/util/PlayerTimeUtilsTests.swift
@@ -12,9 +12,41 @@ final class PlayerTimeUtilsTests: XCTestCase {
     
     func testCalcSeekBackTime() {
         let currentTime: Double = 1000
-        let threeSecondsAgo = Date(timeIntervalSinceNow: -3)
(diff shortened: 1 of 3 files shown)


advplyr:
I tested and this is working well.

I think the seek times that were originally put in are too aggressive, we should follow the same ones android uses.

https://github.com/advplyr/audiobookshelf-app/blob/bcf40c0b7674cc87b4bd4d4676645d2efe1c9785/android/app/src/main/java/com/audiobookshelf/app/player/PlayerListener.kt#L165-L172

What do you think?
I'm not sure what other apps use or if there is a standard somewhere.

complacentsee:
I think the android seek backs make sense. My personal use is heavy on longer pauses.

advplyr:
Works great, thanks!

Issue #452, opened 2022-12-05
Can't seek back after book finishes (iOS)


benonymity:
### Steps to reproduce

1. Download a book
2. Finish listening so there are 0 seconds left
3. Attempt to seek to an earlier time

### Expected behaviour
- The book should be able to rewind to a previous position

### Actual behaviour
- The player loads endlessly

### Environment data

Audiobookshelf Version:

- [ ] Android App?
- [x] iOS App?

#### Android Issue
Android version:

Device model:

Stock or customized system:

#### iOS Issue
iOS Version: 16.1

iPhone model: iPhone SE (2020)

mym0T:
I am able to reproduce this issue in 0.9.61 (18) on iOS.

Additionally, selecting a previous chapter after the book has finished playing doesn't work either.

Selecting the close player option doesn't resolve the issue.
The only options that I've found are:
1. Kill the app, re-open it, open the now playing window, then select a chapter without tapping one of the seek buttons

2. Minimize or close the player, open the book page if it's not already open, reset progress, _kill the app_, play the book again.

Note: The step of killing the app in #2 is only required because the book still shows "playing" after resetting progress even though it isn't actually playing. Killing the app at that point resets the play button.

advplyr:
I was just able to reproduce this in 0.9.62.

You don't need to kill the app. You will have to close the player out completely then discard progress.

advplyr:
Not specific to downloads

p-rintz:
It also doesn't seem overly specific to the book finishing.

Ive had it happen a few times when pausing an audiobook in the middle and coming back to it later.
If I seek then, the same issues arise.

p-rintz:
Sadly, this doesn’t seem fixed. I still have the issue that the player randomly jumps to different chapters in an audiobook when I seek back on a completed book.

advplyr:
Is it possible the files are VBR mp3 files? Have you tried other files of different formats?

p-rintz:
Looks like CBR mp3 files to me. 320k

PR #1514, opened 2025-03-14
Bugfix issue 452:  iOS seek issue at end of book

complacentsee:
This resolves issue #452 failure to seek back at the end of a book. 
I'm less familiar with the podcast side and honestly haven't tested that. I don't think this would cause any issues, but 78d7ba6 could cause issues as it falls back on session details in the event the player doesn't have a track loaded, which is the case after a book ends. 

This does the following:
1. Ensures the UI  shows paused and session current time stops incrementing.
2. Allows a fall back to session time. 
3. Reloads the track queue if it isn't loaded.

### DIFF
--- a/ios/App/Shared/player/AudioPlayer.swift
+++ b/ios/App/Shared/player/AudioPlayer.swift
@@ -511,13 +511,10 @@ class AudioPlayer: NSObject {
         let audioTrack = playbackSession.audioTracks[currentTrackIndex]
         let startOffset = audioTrack.startOffset ?? 0.0
       
-        // if the currentTrackTime is not a number, then track isn't loaded
-        // fall back on session.
-        var currentTrackTime = self.audioPlayer.currentTime().seconds
+        // if the currentTrackTime isNan, then fall back on session.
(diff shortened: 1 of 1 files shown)


advplyr:
When the item finishes and you press the seek back button it restarts.
Otherwise it works in my testing for both books and podcasts

complacentsee:
Restarts meaning it seeks back to the start of the book, or restarts as in restarts playback? Just want to make sure we're seeing the same defect. Local files seem do not show that behavior on my tests. But streaming appears to reset back to the start. 
Looks like 78d7ba6 has more work as on a streaming it's returning nan.
`SEEK: Seek to nan from handler`
I'll keep working on this.

advplyr:
Yeah streaming resets back to the start

complacentsee:
This should resolve the issue. Streaming session vs local session appear return slightly different data, but both store currentTime. So if the player doesn't have a currentTime, we can just use the sessions currentTime directly. Although this does make me wonder why go the bother of any of the logic and not just return the session currentTime.

advplyr:
This is working well in my testing, thanks!

PR #1523, opened 2025-03-22
iOS: Bugfix: resolve race condition when rebuilding queue for seeking back 

complacentsee:
This resolves a bug I introduced in c72f7cd  where a race condition exists when seeking requires rebuilding the audioPlayer item queue and the observers monitor that there is an item playing to ensure stopping at the end of the players queue. 
This PR allows the player to keep track if seeking is causing the audioPlayer item queue to be rebuilding, ensuring we don't end the session inadvertently.


advplyr:
Thanks!

PR #1526, opened 2025-03-26
feat: Add more forward/backward durations

skorokithakis:
This PR adds more forward/backward durations (15/20 seconds), as I find 10 too short and 30 too long.

There's a reference to an icon that I'm not exactly sure the location of, any help would be appreciated.

### DIFF
--- a/store/globals.js
+++ b/store/globals.js
@@ -15,6 +15,14 @@ export const state = () => ({
       value: 10
     },
     {
+      icon: 'forward_15',
+      value: 15
+    },
+    {
+      icon: 'forward_20',
+      value: 20
+    },
+    {
       icon: 'forward_30',
       value: 30
     }
@@ -27,6 +35,14 @@ export const state = () => ({
     {
       icon: 'replay_10',
       value: 10
+    },
+    {
+      icon: 'replay_15',
+      value: 15
+    },
+    {
+      icon: 'replay_20',
+      value: 20
     },
     {
       icon: 'replay_30',


skorokithakis:
I've added the durations, and now they work, except icons with arbitrary times don't exist. I've added the time duration as text in the settings, for clarity, but the `forward_media` icon refuses to render. If anyone knows why, I'd appreciate it.

nichwall:
I responded to you on Discord as well, but it is because arbitrary numbers are not defined in Material Icons.
https://fonts.google.com/icons?icon.query=forward&icon.size=24&icon.color=%23e8eaed

pages/settings.vue
```diff
@@ -46,13 +46,13 @@
       <div class="w-10 flex justify-center" @click="toggleJumpBackwards">
         <span class="material-icons text-4xl">{{ currentJumpBackwardsTimeIcon }}</span>
       </div>
-      <p class="pl-4">{{ $strings.LabelJumpBackwardsTime }}</p>
+      <p class="pl-4">{{ $strings.LabelJumpBackwardsTime }}: {{ settings.jumpBackwardsTime }}</p>
```
nichwall, line 49: This will need to be changed to match other localized strings that have numbers (such as below)

`  "LabelNumEpisodesIncomplete": "{0} episodes, {1} incomplete",`

skorokithakis: Thanks, I thought of that but I didn't want the display to wrap. I could add "sec" to it, would that be good?

nichwall: Yeah, even just an `s` would be good so an `m` could be used for minutes whenever that happens (I'm not saying you need to add the minutes as well in this PR, just making it easier to add later)

skorokithakis: Great, I'll do that now, thanks! Not sure how to localize those, though...

skorokithakis: Done, how does that look?

nichwall: It should all be in the same string instead of concatenating multiple strings. The `{0}` allows for string substitution, so if you search for a label in the codebase you can see how it is used

nichwall:
Hey there, is this PR ready to review now? We have since migrated to use Material Symbols.

One thing I noticed when testing this was that the player alignment gets messed up when using one of the new skip amounts, but the scale is correct when using the previously defined values.

skorokithakis:
I'll give it another shot and report back!

advplyr:
Merged here https://github.com/advplyr/audiobookshelf-app/pull/1767

PR #1529, opened 2025-03-31
Replace Material icons with symbols

nichwall:
## Brief summary

Replace Material Icons with Material Symbols, similar to https://github.com/advplyr/audiobookshelf/pull/3138 and https://github.com/advplyr/audiobookshelf/pull/3291.

## Which issue is fixed?

Fixes symbol definitions not existing for https://github.com/advplyr/audiobookshelf-app/pull/1526/files

## In-depth Description

This PR replaces all instances of `material-icons` with `material-symbols` and updates alignment and sizing for the new symbol definitions. Some symbols look a little different due to using the rounded version.

I opted to only use WOFF2, similar to the web client. Additional reading is [here](https://stackoverflow.com/questions/36105194/are-eot-ttf-and-svg-still-necessary-in-the-font-face-declaration/36110385#36110385).

## How have you tested this?

Tested in a web browser (really just the settings page), and on a Pixel 6a.

## Screenshots

The main differences in symbols are below: Seeking uses larger text, the expandable sections use a solid arrow, and the side menu has more rounded symbols.

seeking
expandable
image


advplyr:
Thanks!

PR #1535, opened 2025-04-07
Modification of sleep timer functioning

ooii:
The sleep timer isn’t paused when the user pauses, and the volume keeps decreasing during pause. These changes restore both volume and sleep timer to their initial values when play resumes.

### DIFF
--- a/ios/App/Shared/player/AudioPlayer.swift
+++ b/ios/App/Shared/player/AudioPlayer.swift
@@ -30,13 +30,13 @@ class AudioPlayer: NSObject {
 
     private var status: PlayerStatus
     internal var rateManager: AudioPlayerRateManager
-    
+
     private var playerContext = 0
     private var playerItemContext = 0
-    
+
     internal var playWhenReady: Bool
     private var initialPlaybackRate: Float
-    
+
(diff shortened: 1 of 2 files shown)


pwinnski:
Conceptually, there seem to be competing assumptions here. Either someone sets a sleep timer to one hour because they need to be sure they listen to the audiobook for no more than one hour, and then sleep or whatever. OR someone sets a sleep timer to one hour because they want to listen to exactly one more hour of the book.

If the first, I'd expect the sleep timer to keep running even if the book is paused. If the second, I would not. 

Personally, I suspect the first scenario is more likely, more in line with the concept of a sleep timer, and pausing seems likely when someone finds they are more tired than they realized and aren't even going to make it an hour.

ooii:
Hi,
I understand that the way I want it to work is not the way all the users would have. But, for me, sleep timer contains the word "sleep" and, in my opinion, if the user interacts with the app, it means that the sleep process has been interrupted. 
Of course, it's also a timer and can be just used as a timer.
Maybe the best way is to have the ability to choose the behavior of the sleep timer, if it should reset or not.

PR #1546, opened 2025-04-20
Logs page with AbsLogger plugin

advplyr:
AbsLogger plugin can be called from the frontend javascript and native android/iOS code (currently only for Android).

```js
// Calling from frontend js
AbsLogger.info({ tag: "", message: "" })
```
```kt
// Calling from native android
AbsLogger.info(tag: "", message: "")
```

Features/updates:
- Logs page link added to side drawer (android only).
- Logs page defaults to masking the server address and has an option to unmask.
- Share button that shares a txt file with all logs.
- Copy button that adds all logs to clipboard.
- Clear logs button that deletes all logs from db.
- On app launch, all logs older than 48 hours are deleted.
- AbsLogger emits `onLog` events & logs page is real-time (e.g. watch progress syncing while listening).

Mostly only logs relating to progress syncing or app initialization have been added so far. There are a lot of old/useless logs in the app now so when switching them out with the new AbsLogger it is an opportunity to clean that up.

image
image


advplyr:
This PR also updates the Abs iOS plugins for Capacitor 7 and sets up AbsLogger.swift with stub functions.

Issue #1553, opened 2025-04-23
[Bug]: App crashes when downloading an audiobook


Zibbp:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

Clicking the "download" button for an audiobook crashes the app when attempting to download.

### Steps to Reproduce the Issue

1. Navigate to an audiobook
2. Click the download button

### What was Expected?

The audiobook should download without crashing.

### Phone Model

Samsung Galaxy S22

### Phone OS

Android 14

### Audiobookshelf App Version

Android App - 0.9.79

### Installation Source

Google Play Store

### Additional Notes

App is actually 0.9.80-beta, no option for that.

Server version: v2.20.0 in docker.

The in-application logs do not catch this. I do have ADB debug logs though.

<details>
  <summary>adb logs of crash</summary>

```
04-23 18:22:12.883 27468 28717 D AbsDownloader: Got library item from server ce086423-0e35-4f27-b832-757a7da4c8d9
04-23 18:22:12.884 27468 28717 D AbsDownloader: downloadCacheDirectory=/data/user/0/com.audiobookshelf.app/files/downloads/ce086423-0e35-4f27-b832-757a7da4c8d9
04-23 18:22:12.884 27468 28717 D AbsDownloader: Starting library item download with 1 tracks
04-23 18:22:12.884 27468 28717 D AbsDownloader: Audio File Server Path /api/items/ce086423-0e35-4f27-b832-757a7da4c8d9/file/1280/download | AF RelPath Ruins of the Earth Book 3 - Phantom Deadfall.m4b | LocalFolder Path  | DestName Ruins of the Earth Book 3 - Phantom Deadfall.m4b
04-23 18:22:12.884 27468 28717 D AbsDownloader: TEMP Audio file already exists, removing it from /data/user/0/com.audiobookshelf.app/files/downloads/ce086423-0e35-4f27-b832-757a7da4c8d9/Ruins of the Earth Book 3 - Phantom Deadfall.m4b
[...]

PR #1556, opened 2025-04-24
fix(InternalDownloadManager): add accept-encoding identity to requests

Zibbp:
Fixes https://github.com/advplyr/audiobookshelf-app/issues/1553.
@nichwall pointed me in the correct direction for a fix. I've tested this against my server through a reverse proxy with gzip and without the proxy, directly to the container, both work fine. This disables compression for requests using the internal download manager client. As seen in the issue the okhttp client is sending an accept encoding for gzip by default which causes issues.
And more testing you want to do on your end?

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/InternalDownloadManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/InternalDownloadManager.kt
@@ -17,16 +17,8 @@ class InternalDownloadManager(
 ) : AutoCloseable {
 
   private val tag = "InternalDownloadManager"
-  private val client: OkHttpClient = OkHttpClient.Builder()
-      .connectTimeout(30, TimeUnit.SECONDS)
-      .addInterceptor { chain ->
-          val originalRequest = chain.request()
-          val newRequest = originalRequest.newBuilder()
-              .header("Accept-Encoding", "identity")
-              .build()
-          chain.proceed(newRequest)
-      }
-      .build()
+  private val client: OkHttpClient =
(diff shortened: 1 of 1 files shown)


nichwall:
Can you explain the interceptor and chain section? Is there a reason we need to do that instead of just setting the header?

Zibbp:
My experience with Java is limited so I may be missing something obvious here. The [client builder](https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.Builder.html) doesn't have a header method. Searching around the only way to actually set the header is on the [request object](https://square.github.io/okhttp/3.x/okhttp/okhttp3/Request.html), that has the header method.
https://stackoverflow.com/questions/32196424/how-to-add-headers-to-okhttp-request-interceptor
Can you think of a cleaner way to add the header?

nichwall:
Ah, thanks for the reference! I did some more reading and it looks like we can actually edit the request in the `download` function.
https://stackoverflow.com/questions/41164299/okhttp-enable-disable-gzip-compression-on-requests

This change seems to work with my app, but I have not tested it through a reverse proxy with gzip, just that it does in fact compile.
```
fun download(url: String) {
    val request: Request =
            Request.Builder().url(url).addHeader("Accept-Encoding", "identity").build()
    client.newCall(request)
            .enqueue(
```

Zibbp:
That way is much nicer being just on the download function. I updated & tested. Confirmed I can download files both through the proxy with gzip and directly from the container (no proxy).

nichwall (approved):

advplyr:
Thanks!

PR #1561, opened 2025-04-27
Add: settings for gradle project in subfolder

nichwall:
## Brief summary

This PR adds a new VS Code settings file to allow the gradle project to be in a subfolder (in our case, `android/`).

## Which issue is fixed?

No specific issue

## In-depth Description

This allows VS Code to recognize the `android/` folder as the gradle root. I am not sure if this is the best way to do this, but could not figure out how to do this only from `.vscode/settings.json`.

## How have you tested this?

VS Code no longer complains about imports being incorrect/not existing.

## Screenshots

Prior to adding this new VS Code settings file
image

### DIFF
--- a/android/.vscode/settings.json
+++ b/android/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+  "java.project.root": "android",
+  "java.import.gradle.enabled": true,
+  "java.import.gradle.wrapper.enabled": true,
+  "java.import.gradle.user.home": "${workspaceFolder}/.gradle"
+}


advplyr:
What VSCode extensions are you using? I spent like 2 hours last week trying to get intellisense working for the android code and got no where.

nichwall:
> What VSCode extensions are you using? I spent like 2 hours last week trying to get intellisense working for the android code and got no where.

Extensions that may apply (not including unrelated like C/C++ and Python things)
- `EditorConfig`
- `Gradle for Java` (newly installed when trying to debug, did not have it until this past weekend)
- `Kotlin` by fwcd
- `Prettier`
- `Vetur`
- `Vue - Official`

I assume it is the Kotlin extension.

advplyr:
Okay thanks, I'll test try to get it working again another day. It is much easier to navigate VScode than to navigate android studio

nichwall:
Sounds good. I'm not sure what is auto formatting the Kotlin code, I have not done much research. It may be that same Kotlin extension because I believe prettier doesn't include Kotlin

PR #1562, opened 2025-04-28
Template updates

nichwall:
This PR adds a template for pull requests, similar to how it is done on the server repository. I also added a question of whether the PR targets Android or iOS. (however, I did not use the template for this issue...)

I also updated the issue templates for the new app versions.


advplyr:
Thanks!

Issue #1479, opened 2025-02-13
[Bug]: Download sometimes gets stuck on single files


TrimVis:
### What was the Problem?

Sometimes, especially with a unreliable connection to the server (mobile data/bad wifi), single files will get stuck when trying to download a full audiobook.
By stuck I mean that the percentage displayes does not increase and the stuck file is included in the download limit (e.g. if I set it to download at most 4 files at once, it will only download 3 files at once with one stuck file).

Waiting does not resume the download and on a app restart all downloaded files will be forgotten

### Steps to Reproduce the Issue

1. Connect to a remote server with a slow/unreliable internet connection
2. Download an audibook with many files
3. Some files will get stuck, while all the remaining ones finish normally. However the full audibook download never terminates.

### What was Expected?

I would have expected the app to detect the stuck/stalles file download and correspondingly retry. 
Alternatively the ability to retry downloading of single files could help alleviate this problem (I couldn't figure out how to do that without retrying the full audiobook download)

### Phone Model

OnePlus 12R

### Phone OS

Android 15

### Audiobookshelf App Version

Android App - 0.9.79

### Installation Source

Google Play Store

### Additional Notes

I had this issue appear reliably on mobile data (with a in theory good internet connection) and on wifi (with a not great WiFi signal).
The server is hosted at another place and accessed via a wireguard vpn (I doubt that this is related to the VPN, though).
During all this I always had the app open in foreground, as I was suspecting that battery saver shenanigans were happening (even though I have optimizations disabled/set to ask)

HamletDuFromage:
Same issue here. Got a download stuck at 20%. Stopping the app discards all the progress, and when I started over I managed to get to 64% but it still got stuck.

TrimVis:
If there is any way to provide additional logs, I would be happy to do so :)
Just not aware if this is natively supported by the app or not

nichwall:
There is not currently a way to export logs from the app (you can compile from source to get the logcat data). I have been working on refactoring and cleaning up the download logic for Android the past few weekends, so hopefully there will be something for testing soon.

TrimVis:
Thanks for the update!
You work is appreciated, looking forward to the (beta/test) release

dacmcbibs:
I'm having a similar issue, but for me the download gets stuck if the app loses focus , as long as i leave it in the foreground everything downloads correctly.  I can easily reproduce, so if a logcat would be helpful let me know

nichwall:
[...]

PR #1469, opened 2025-02-07
Download manager cleanup

nichwall:
This PR refactors the `DownloadItemManager.kt` and `InternalDownloadManager.kt` by:
- Adding comments to functions
- Applying autoformatting
- Refactoring functions into smaller functions to make reuse and replacement simpler
- Increases the cache size for writing internal files from the `InputBufferStream`
- No functional changes were made

This PR is part of an effort to document files without introducing functional changes to make things a bit easier to follow.

This was tested by downloading files to both internal and external folders on a Pixel 6a running Android 15.

Issue #1386, opened 2024-11-20
[Enhancement]: Claim orphaned downloads when server address changes, but server itself is the same


notoriousjere:
### Describe the Feature/Enhancement

Add the ability to claim orphaned downloaded files in cases where the server itself is the same as the one that the download belonged to, but it's address merely changed.

### Why would this be helpful?

I recently migrated my docker-managed server from Computer A with IP address 192.168.1.6 to Computer B with IP address 192.168.1.7. I then changed the Static IP of Computer B to 192.168.1.8. For each of these changes, all of my locally downloaded books showed this message:

> Media is linked to an Audiobookshelf serer on a different address (\<address\>). Progress will be synced when connected to this server address.

I had to delete all of my locally downloaded books and re-download them each time my server's IP changed, which is a process I feel the app could handle doing itself.

### Future Implementation (Screenshot)

This could be a button shown on the dialog box. The button would kick off a process to match the orphaned book to the newly connected server (in my case, it'd be the same one it belonged to, but just had a different IP address). This process could succeed or fail. A warning to the user that the downloaded media would no longer be linked to the old server feels like it would apply here.

![mockup](https://github.com/user-attachments/assets/48831b94-1a3a-4ba6-9b25-166e44d1e2dc)

### Audiobookshelf App Version

Android App - 0.9.77

### Current Implementation (Screenshot)

![share_844640594064513293](https://github.com/user-attachments/assets/5d8fe5cc-4bc8-43c9-a863-8bce6896cdeb)

nichwall:
Related to https://github.com/advplyr/audiobookshelf-app/issues/209

brent-holland:
This would be super helpful to keep syncing media with my laptop when I'm on a diff network

bigbeka:
I came across the same issue. Is this a planned feature/enhancement?

TheTerrasque:
I just got hit by this issue too. I don't understand, how is this even a problem? Server lists library, says book X is at progress Y, you open that book on app and moves to that progress, then report progress as it's played / read. Where does it even come in that you also read the same book on a different server before? Why does that matter at all?

nichwall:
> I just got hit by this issue too. I don't understand, how is this even a problem? Server lists library, says book X is at progress Y, you open that book on app and moves to that progress, then report progress as it's played / read. Where does it even come in that you also read the same book on a different server before? Why does that matter at all?

[...]

Issue #1362, opened 2024-11-04
[Enhancement]: Scan local download folder for existing media


DrakeHamString:
### Describe the Feature/Enhancement

Scan the download folder for existing media and if found, then mark the corresponding audiobook as already downloaded and link it.

### Why would this be helpful?

When having to reset the app and re-login, I don't want to download my audiobooks again, that are already on the phone.

I've set up a local folder containing all files. When selecting this folder again, ABS shows that there is no media (but the audiobook files are existing there). When downloading the audiobooks the app just overwrites the existing files.

### Future Implementation (Screenshot)

In local folder option, add a new option to the three-dot-menu called "Scan folder for local media"
![image](https://github.com/user-attachments/assets/20f6421f-0fda-4aad-9f8f-2bd0fd5edf5c)

![image](https://github.com/user-attachments/assets/43a9de9f-3979-406a-a715-eb98b44c90a0)

Of course in the correct languages

### Audiobookshelf App Version

Android App - 0.9.76, Android App - 0.9.75

### Current Implementation (Screenshot)

Local Folder with files in it:
![image](https://github.com/user-attachments/assets/7985a3f5-1ae4-45e5-a89b-4e97a3eb29c8)

App shows no dowloaded audiobooks:
![image](https://github.com/user-attachments/assets/8f1b5278-6768-4451-b86e-8eba96c52a26)

App offers stream option and download option - no local play option:
![image](https://github.com/user-attachments/assets/6e6fec7e-67bc-449f-8bdd-595f994ff5a6)

nichwall:
This functionality was removed in https://github.com/advplyr/audiobookshelf-app/releases/tag/v0.9.67-beta

DrakeHamString:
Why?
This makes the user experience worse

Today I had to reset the app because it was crashing everytime when connecting to the server. After reset, I was able to use it again, but then I had to re-download every media...

realredknight:
This is also useful when using more than one device - it would allow to push media files over USB instead of downloading over WAN one multiple devices.

Willt125:
I would've labeled this as a bug, because it's pretty obvious that if the app has the ability to load local folders, and describes it as loading local media, that the app should load in the local media. I'm seeing this problem as well - Samsung Galaxy A15 5G, selecting a local folder with books does not register as containing books.
[...]

PR #1587, opened 2025-05-26
Android download rewrite

### LINKED ISSUE #34: partial downloads break downloading of an audiobook
This appears to be downloading to the download dir and then copying over to  storage dir. If audiobook shelf app is interrupted between passing download to whatever is downloading and completing moving to the storage dir you will end up with a fully downloaded audiobook in your download dir but will not  be able to play the local copy in the app or initiate a successful download because whatever is actually downloading the book appends -1, -2, etc depending how many copies are in the download dir and when audiobookshelf app tries to copy over the copy not associated with the download request, it fails. You can recreate this behaviour by starting a download but killing the app in between when the download of the audiobook itself is started and when it is finished.

from my previous bug  where I figured out what was happening:
 The issue appears to be the app downloads it expecting it to be named string.mp3 and when the android downloading mechanism finds string.mp3 already in the download dir it puts it in the download dir as string-1.mp3, at which points audiobookshelf app loses its mind and gives all this weird behavior.

### LINKED ISSUE #862: Download can't be restarted and would fill up internal app storage
### Steps to reproduce

1. Start a download in the ABS App.
2. Close the app and swipe it up in the app menu (to end it).
3. Open the app and start the download again.

### Actual behaviour
- Problem 1: All files which are already in place for the first failed download will not get deleted or overwritten. It will look like this:
  - chapter 1.mp3
  - chapter 1 (1).mp3
  - chapter 2.mp3
  - chapter 2 (1).mp3
  - chapter 3.mp3
  - chapter 3 (1).mp3
- Problem 2: When the folder "Internal App Storage" is used, failed downloads can't be clean manual, as it can't be accessed in the file manager app. For custom defined download locations, there is the option to scan or force scan the folder. 
For Internal App Storage this is not available. So when a download is failed, there is no way for the user to clean this data and over time it will get more block space. Then the user needs to reinstall the app to have the app data deleted. 

### Preferred solution
All audiobooks which are selected to be downloaded should be displayed in the specific folder under local media. Also, when the download is not finished yet. 
The download progress, size on the local storage and issues could be displayed there.

Possible actions for the user: 

- Remove (stop the download in case it is still running and delete local data)
- Pause and resume download
- Restart the download if it was interrupted for any reason.

### Other solutions
- Solution 1: Make the download process restart save, so that downloads could be continued after the app was closed.
- Solution 2: When a download is started and there is already local data for this audiobook (due to a failed download) delete the local data first. Maybe prompt the user with a warning, if he wants to continue.
- Solution 3: Enable the possibility to scan and force scan the Internal App Storage download location. In this way, the users can be sure to have all data in their visible.

### Conclusion
The preferred solution would bring a lot of benefit to the app in making the download process and management of offline data a lot easier to handle.
If not possible for some reason, then just solution 2 and 3 would also help a little.

### Environment data

Audiobookshelf Version: 2.3.3

- [x] Android App?
- [ ] iOS App?

#### Android Issue
Android version: 13

nichwall:
## Brief summary

This PR refactors the Android download logic to use a single code path for both Internal App Storage and Shared Storage. This PR also allows for resuming an incomplete download.

**This PR does not add support for downloading with the app in the background (see Further Work below).**

## Which issue is fixed?

Fixes https://github.com/advplyr/audiobookshelf-app/issues/862
Fixes https://github.com/advplyr/audiobookshelf-app/issues/34

This PR is mostly just refactoring without fixing any specific issues. I asked for feedback in the following issues to see if the refactor fixed the issue, but still unsure if those issues are fixed so we should leave them open.

Downloading:
- https://github.com/advplyr/audiobookshelf-app/issues/1479
- https://github.com/advplyr/audiobookshelf-app/issues/1273
- https://github.com/advplyr/audiobookshelf-app/issues/1253
- https://github.com/advplyr/audiobookshelf-app/issues/1167
- https://github.com/advplyr/audiobookshelf-app/issues/215

Associating downloaded media with server connection:
- https://github.com/advplyr/audiobookshelf-app/issues/1386
- https://github.com/advplyr/audiobookshelf-app/issues/1362
- https://github.com/advplyr/audiobookshelf-app/issues/594

## Pull Request Type
Android backend

## In-depth Description
This PR is a continuation of https://github.com/advplyr/audiobookshelf-app/pull/1469. Previously, there were two nearly identical code paths for Internal App Storage (using `File`) and Shared Storage (using `DocumentFile`). We also performed an expensive scan (long running) in Shared Storage once all of the files completed downloading.

As part of this PR, we now:
- Always download to Internal App Storage, then move to Shared Storage if desired.
  - This allows us to reuse the same download logic for both internal and shared storage (using `InternalDownloadManager`). This also simplifies continuing incomplete downloads, discussed below.
- `LocalFile` class updates:
  - Remove `simplePath` because it is one of several slow properties (SAF) and is unused anywhere in the codebase.
  - Added `extension` to improve code reuse.
- Build the `LocalLibraryItem` object from the `DownloadItem` object instead of scanning folder in Shared Storage.
  - This still takes about 50-100ms per file in my testing, but is much faster than scanning the folder and *then* still building the object through SAF.
- Say "Moving and processing files" instead of just "Processing"
  - More descriptive about what is happening. When downloading a single large file to Shared Storage, moving the file takes a long time and processing is very quick. When downloading multiple smaller files, moving is very quick (every file is moved as soon as it finishes downloading), but scanning can take a few seconds. Downloading to Internal App Storage does not require moving files and scans many files almost instantly.
- Do not delete temporary or destination files.
  - This allows continuing a long-running download (large single-file M4B), or multi-file downloads.
  - Every file is still added to the download queue. If the file already exists in Internal Storage, continue downloading from the end of the file. If the file is already complete, then the file completes immediately. If the file already exists in Shared Storage, it was already successfully downloaded and moved and is marked as moved.
  - This also allows users to "redownload" a book after resetting the app or manually creating the file structure without actually downloading the entire book again.
  - This drew inspiration from https://github.com/advplyr/audiobookshelf-app/pull/1230 and https://github.com/advplyr/audiobookshelf-app/pull/1231.
- Check free space in Internal App Storage before downloading any file
  - The "minimum free space" required is the size of the downloaded file multiplied by the `maxSimultaneousDownloads`
  - If there is not enough free space, the AbsLogger shows an error and the file is not downloaded. The download queue can continue with other files, so larger files can still be downloaded in series instead of parallel.

### Further work
Building the `LocalLibraryItem` as each individual file is moved to Shared Storage would reduce the processing time to nearly constant (time to move file + 100 ms), but complicates the logic a lot more with incomplete downloads and the queue system.

I would like to move the download manager to a Foreground service to help with downloads being stopped in the background when a user switches to another app. Further testing would be needed before adding a Foreground service. https://github.com/advplyr/audiobookshelf-app/issues/956

## How have you tested this?

Tested with Android 15 on a Pixel 6a with both books and podcasts.
*This should have additional testing done with other devices, especially older versions of Android.*

## Screenshots
N/A


timwhite:
I find both the 0.9.81-beta and the APK attached to this PR have the same issue, that if you switch away from the app and back, the downloads appear to have frozen and don't resume.
(So I think the issue @troy117 is having isn't new, and I was hoping this PR would fix it)

nichwall:
Yeah, this PR does not implement downloading in the background. This PR *should* allow you to press the download button again and have the download continue from where it left off, but it sounds like that is not working either. Thanks for the feedback y'all. :)

troy117:
After doing more testing it does continue the download from where it left off, after the app is force closed and then after the download is restarted.  I made one change to your code for the version I am running locally on my phone which is to catch the random end of stream exception that happens and continue the download and it works great.

timwhite:
It does work for pressing download again and it resumes part way through a downloaded file, or completed files download basically instantly. So it works for that issue. That in itself makes a huge difference for poor connections, as long as I keep it active. (And knowing to push download again to make it resume was key to testing that, I expected it to resume when I reopened the app)

https://github.com/advplyr/audiobookshelf-app/issues/956 looks to be the best place to track the background/sleep download issue.

nichwall:
Ah, cool. Thanks for clarifying that pressing the download button again is working to continue the download (that was the intention).

I'm hoping to address downloading in the background using Work Managers or similar, but I am trying to keep this PR to refactoring with minimal functionality changes to make the following PRs simpler.

kuci-JK:
Hi, I am excited about any improvements to the download experience :)

I just want to point out something that occured to me while reading your description..

> Always download to Internal App Storage, then move to Shared Storage if desired.

While I think code sharing is great, when user is running low on internal storage this might cause problems, right?

For example:
Let's say I am downloading 4 audiobooks for holidays. My phone has low internal storage, so I have an SD card for media or other large files (3 GB free internal, 10 GB free card).
If each book has 1 GB for example, I turn the download on all of them it might fail because the app would fill up the internal storage even though I would have set the app to save to the SD card.
I would need to download the one by one and each time wait before starting the next one -> that is in my opinion not great.

I feel like the user should be able to save the files on the SD card even when his/hers internal storage does not have enough space. While this might not be an issue for many users, I still think it should be taken into a consideration.

That's just something you can take into consideration while implementing the rewrite :)

nichwall:
> Let's say I am downloading 4 audiobooks for holidays. My phone has low internal storage, so I have an SD card for media or other large files (3 GB free internal, 10 GB free card).
> If each book has 1 GB for example, I turn the download on all of them it might fail because the app would fill up the internal storage even though I would have set the app to save to the SD card.
> I would need to download the one by one and each time wait before starting the next one -> that is in my opinion not great.

As soon as the first file finishes downloading, it is moved from the Internal Storage to the Shared Storage (the SD Card in this example). You should still be able to queue up all 4 books as long as you have enough storage for each individual file.

The issue would be that up to 5 files can be downloaded in parallel. I would need to double check whether this applies to all of the files in the queue or all of the files in the book (meaning the next book does not start until the previous book is finished).

jo-z:
just wanted to toss in a comment saying that downloads randomly stopped working for me on the main app, but work fine again on this one

nichwall:
I added some additional logging and added a check to prevent trying to download files that are too large, but keep the files in the download queue so the files can still be downloaded in series instead of running out of space downloading in parallel. I have updated the top-level comment and marked 2 issues that I believe are fixed by this PR.

I believe this PR is ready for review and merging. Nobody has raised any new issues created by this PR in several GH linked issues or Discord and Reddit discussions.

The main issue people mention is background downloading does not work, but that did not work consistently before this PR anyway and this PR does not attempt to add that functionality.

advplyr:
I started a download to internal storage then locked the screen half way through.
I know background downloading isn't supported but this caused an infinite loop.

Here
```kotlin
while (currentDownloadItemParts.isNotEmpty()) { }
```

image

After closing and re-opening the app starting the download again resumes where it left off.

I'm able to reproduce that consistently when downloading to internal storage. 

When downloading to shared storage the download works in the background. I still have to go through all the code but from your description it sounds like they should function the same since shared storage is downloading to internal storage the same?

nichwall:
Yes, they should both be downloading to internal storage and then moving the shared storage. Is resuming the download to shared storage working when you lock the phone and then restart the app?

I will take a look at fixing the infinite loop there. It is intended that the download tries to resume, but having it just spin there will probably recreate the battery drain issue that was just fixed for Android Auto.

matuszeg:
I wanted to mention that I downloaded the apk for this pull request and it solved my problem of android downloads being incredibly slow. a 800MB file would take approximately 1 hour to download. After I switched to this apk, it downloaded in less than 1 minute. Same server, same internet connection. Only thing I changed was the apk I was using.

nichwall:
> I started a download to internal storage then locked the screen half way through. I know background downloading isn't supported but this caused an infinite loop.  Here:
> ```kotlin
> while (currentDownloadItemParts.isNotEmpty()) { }
> ```
> After closing and re-opening the app starting the download again resumes where it left off.
> I'm able to reproduce that consistently when downloading to internal storage.

I was able to reproduce the infinite loop when locking the phone and added a 10 second timeout after the OkHTTP download socket disconnects so the infinite loop will eventually stop. Unfortunately I have not figured out how to clear the download queue and update the frontend nicely in this case, so that will take some more work.

Currently the download progress percentage and icon at the top of the screen get stuck and cannot resume the download or clear the queue without restarting the app. I will work on getting that fixed before this is ready to merge.

> When downloading to shared storage the download works in the background. I still have to go through all the code but from your description it sounds like they should function the same since shared storage is downloading to internal storage the same?

Yes, downloading in the background *can* continue if the process is not killed by the system, but it is not explicitly added. In my testing I see downloads to both Internal App Storage and Shared Storage can continue for about 5-10 seconds in the background before my Pixel kills the OkHTTP connection. I'm not sure exactly what the difference would be for your phone since we are using the exact same function for downloading to both locations, but it could be related to how the path is constructed. Depending on how hard updating the front end is above, I may end up just implementing full background downloading, but we'll see what time permits over the next few weeks.

> I wanted to mention that I downloaded the apk for this pull request and it solved my problem of android downloads being incredibly slow. a 800MB file would take approximately 1 hour to download. After I switched to this apk, it downloaded in less than 1 minute. Same server, same internet connection. Only thing I changed was the apk I was using.

Thanks for the feedback, I'm glad it's working for you. :)

damccull:
Can we get another test build of this? And maybe see it merged in if it works? :D

damccull:
Ah, I figured out how to built it locally with 'act'.
For anyone else:
1) install docker + act
2) create a folder for artifacts
3) clone this repo
4) run:
    ```
    git fetch origin pull/1587/head:pr-1587
    git checkout pr-1587
    ```
5) run:
    ```
    act push --job build --artifact-server-path path/to/artifacts/folder
    ```
6) browse to artifacts folder
7) unzip the zip
8) untar the tar
9) copy the APK to your android and install it

damccull:
Would be cool to see a detection of internet return automatically and resume the download. Still gets stuck like the original, but when you kill and restart it, it continues from where it left off, so that's good.

nichwall:
Thanks for all of the feedback on this PR, and apologies on never finishing it. After more than a year, I am closing this PR in favor of a new PR since I never got around to fixing those last few bugs, several of which were introduced by attempting to set up for background downloading.

I have just opened #1937 which achieves most of the same functionality in a cleaner way, and also adds background downloading and resumption support if any of you want to check it out.

PR #935, opened 2023-11-10
feat: Restore Android 6 support

spieden:
Anecdotally running just fine on a couple of Sony XPeria Z3Cs with these changes

PR #1591, opened 2025-05-30
Use Moshi to speed up menu on Android Auto

### LINKED ISSUE #1570: [Bug]: Continue tab takes long time to show up on Android Auto

Yes

### What was the Problem?

I am relatively new user of the app (~2 months). I was very happy with the app, my only usability problem was that after attaching to Android Auto I had to navigate to the latest book starting with the Libraries tab. Well, I've always had just two tabs on the android auto screen: Libraries and Downloads. I've then prompted my wife to install the app on her phone. Imagine my surprise, when she got the "Continue" tab and could just resume listening with a single tap! For reference: my phone is a Pixel 6A, her is a Samsung A52s. 

At that time I've used Tailscale VPN to connect with the phone and assumed that maybe it's a connectivity issue of some kind. But then I've downloaded the sources, connected through the Desktop Head Unit and did some digging.

The `ApiHandler.getAllItemsInProgress` function would take a lot of time to finish when stepping through with the debugger. I've added some logging inside the `ItemInProgress.makeFromServerObject(jsobj)` call and I've noticed that https://github.com/advplyr/audiobookshelf-app/blob/cf2f684e80545039d3dc9adaab8f3ecb5c63c6cd/android/app/src/main/java/com/audiobookshelf/app/data/ItemInProgress.kt#L23 takes over 100ms to parse an item. Given that the server returns 16 of them, it quickly adds up.

For comparison I've added Gson to dependencies and did a very quick deserialization test with that and Gson seems to be an order of magnitude faster. This is an excerpt from logcat while running on real device (Pixel 6A) with Desktop Head Unit showing the UI on PC:

```
2025-05-10 23:53:26.807  com.audiobookshelf.app.debug         E  processing items in progress
2025-05-10 23:53:26.807  com.audiobookshelf.app.debug         E  processing item 0
2025-05-10 23:53:26.808  com.audiobookshelf.app.debug         E  serialize to string
2025-05-10 23:53:26.809  com.audiobookshelf.app.debug         E  reading with Jackson
2025-05-10 23:53:27.319  com.audiobookshelf.app.debug         E  read item
2025-05-10 23:53:27.319  com.audiobookshelf.app.debug         E  reading with Gson
2025-05-10 23:53:27.344  com.audiobookshelf.app.debug         E  read item
2025-05-10 23:53:27.344  com.audiobookshelf.app.debug         E  processing item 1
2025-05-10 23:53:27.344  com.audiobookshelf.app.debug         E  serialize to string
2025-05-10 23:53:27.345  com.audiobookshelf.app.debug         E  reading with Jackson
2025-05-10 23:53:27.805  com.audiobookshelf.app.debug         E  read item
2025-05-10 23:53:27.805  com.audiobookshelf.app.debug         E  reading with Gson
2025-05-10 23:53:27.812  com.audiobookshelf.app.debug         E  read item
```

Quick look in the debugger suggests that both Jackson and Gson deserialized roughly the same data (so the Gson-derived object and its subobjects are not empty etc.)

I have no idea why it takes almost 500ms here, maybe because the screen on the device was off? Running the loop over 16 items in `getAllItemsInProgress` took over 7 seconds. I guess the real Android Auto in my car may just timeout.

I cannot imagine Jackson to be an order of magnitude slower than Gson in general, so something else is amiss here. Maybe some configuration issue?

1. Connect the app to android auto.
2. No Continue tab present on the car infotainment screen.

### What was Expected?

Continue tab present to allow quick resume.

### Phone Model

Pixel 6A

### Phone OS

Android 15

Android App - 0.9.81

### Installation Source

Other (List in "Additional Notes")

### Additional Notes

Compiled from source, installed with Android Studio.

golinski:
## Brief summary

This PR uses Moshi serialization library to speed up items loading on Android Auto initialization.

## Which issue is fixed?

Fixes #1570

## Pull Request Type

Android only

## In-depth Description

Jackson object mapping is slow on the initial run as all the classes need to be loaded and their annotations reflexively checked.
Moshi use code generation to create "adapters" the do the same work at build time, which speeds things up considerably. Some custom adapters are also written (in fact: generated first by Moshi and then changed a bit by me) to accommodate the schema used by the backend.

Initially, I wanted to convert everything from Jackson to Moshi, but this was problematic, as there are corner cases and the class hierarchy does not suit Moshi very well. SO I went for a more TDD approach and focused on the classes underlying Recent and Continue tabs only. The rest can be done piece by piece, and even if not, then this change makes the Android Auto load fast enough.

## How have you tested this?

On my phone with the desktop head unit, moreover unit and instrumented test are added using json responses from my server, to make sure the data is the same when deserialized with Jackson and with Moshi.


golinski:
I would love to do the same treatment to the tile that pops up on the Android Auto home screen, as it also seems to take a long while to show up. But I cannot figure the code path that drives it.
image

golinski:
I still see there is a problem. There is a marked improvement with these changes, but I see that the personalized libraries are being loaded twice. Once I've seen in my car that the "Recent" tab was shown, but showed only a spinner. Or the "Continue" tab would show up but be completely empty. It's possible some race condition is still happening.

advplyr:
When I run this the app starts but `AbsDatabase` plugin fails to load with no error thrown besides that it failed to load.

Then I comment out 
```kt
//    DeviceManager.dbManager.cleanLocalMediaProgress()
//    DeviceManager.dbManager.cleanLocalLibraryItems()
```

AbsDatabase plugin loads but the app crashes when deserializing `localLibraryItem`.
```
io.paperdb.PaperDbException: Couldn't read/deserialize file ...
...
at com.audiobookshelf.app.managers.DbManager.getLocalLibraryItems(DbManager.kt:37)
```
I trimmed the unnecessary parts of the error.

Do you not have any local library items downloaded on your test device?

golinski:
I don't have downloaded files, I only stream. So back to the drawing board.

golinski:
I've observed similar problems with Paper DB when I done some slight changes to the model classes hierarchy before. So here I was careful to not change those classes at all.

It's possible that using a single shared instance of the ObjectMapper was the culprit. Jackson claims that a single instance should be thread-safe, but maybe when Capacitor plugins are all initialized at the same time something goes wrong. I've removed that single commit from the PR (as the change is rather orthogonal to using Moshi), rebased and forced push. In your free time check if this indeed helps with the issue you are experiencing.

From what I was able to pick up from what I've read, situation might improve with newer Jackson versions, but from version 2.14 Jackson requires Android SDK 26, while <=2.13.5 should be compatible with Android SDK 19.

BTW, I've been using this version for a while in my car and normally with the UI and I haven't seen the app crashing, but I don't keep any downloads on device.

golinski:
Also one more thing: my experiences with the past approach to this PR were that PaperDB would break in a way you described on a run **after** a serious crash. My intuition was that on a crash occurring on loading the app (completely unrelated to PaperDB) something was not synced properly to the file/database underlying PaperDB. On the next run it would cause a crash when loading data, even if issue was fixed. I haven't really investigated the issue, just wiped the app data in the emulator.

But I haven't seen this with the code in the PR.

nichwall:
> From what I was able to pick up from what I've read, situation might improve with newer Jackson versions, but from version 2.14 Jackson requires Android SDK 26, while <=2.13.5 should be compatible with Android SDK 19.

We can probably raise the minimum SDK version to 26. For context, SDK version 19 is Android 4, and SDK version 26 is Android 8. We dropped support for Android 6 a while ago, so I think it is safe to raise the SDK version. Some people are still using ABS with Android 6 on their old devices, but I think the maintenance burden is outweighed by upgrading the minimum Android version. https://github.com/advplyr/audiobookshelf-app/pull/935

As of February 2025, only 145 people were using Android 8 out of over 65k installs according to Google Play statistics.

Edit to add: the current minimum SDK is 24, which is Android 7. https://github.com/advplyr/audiobookshelf-app/blob/master/android%2Fvariables.gradle

advplyr:
Thanks @nichwall. I did another check of the data

Version | SDK | Installs
------ | ------ | -------
Android 8.1 | 27 | 153
Android 8.0 | 26 | 122
Android 7.1 | 25 | 65
Android 7.0 | 24 | 98

Total installs: 70,871

If we raised the minimum SDK to 26 that would affect 163 installs (~0.2%). That seems reasonable if we get performance improvements for the other 99.8%.

PR #1611, opened 2025-06-22
Add toggle to disable bookmark text auto focus

abguymon:
## Brief summary

Adds a toggle for the feature of auto selecting the bookmark text input. The reason being some users (like myself) don't want the keyboard to pop up when setting a bookmark and don't care about the text.

## Which issue is fixed?

None

## Pull Request Type

Affects both apps. 
Affects both the frontend and backend of the apps (adds a new item into the device settings data store)

## In-depth Description

Recently a change was made to auto select the text input when creating a bookmark. This negatively affected my use of the app as when I'm listening at night I would like to just quickly set a bookmark and then go to sleep. Auto selecting the text input makes the bright keyboard pop up and makes it harder to select create bookmark. I have added a toggle which allows users to disable the auto select so they can go back to how the app worked before. This way it still defaults to the new method but for users that would like to go back they now have an option.

## How have you tested this?

1. Open the app
2. Start listening to an audiobook
3. Create a bookmark (This should auto select the text input)
4. Go to settings 
5. Enable the toggle for disable auto focus
6. Create a bookmark (This should **not** auto select the text input)

## Screenshots

**BEFORE**

https://github.com/user-attachments/assets/3a5dd2f8-0a80-4454-a194-cad23a528e92

**AFTER**

https://github.com/user-attachments/assets/0ae9693c-cb68-4070-bdb9-5006dc4c4105

### DIFF
--- a/components/modals/BookmarksModal.vue
+++ b/components/modals/BookmarksModal.vue
@@ -160,18 +160,29 @@ export default {
       this.newBookmarkTitle = this.$formatDate(Date.now(), 'MMM dd, yyyy HH:mm')
       this.showBookmarkTitleInput = true
 
-      // Only auto focus if the setting is enabled
-      const deviceSettings = this.$store.state.user.deviceData?.deviceSettings || {}
-      if (deviceSettings.bookmarkAutoFocus) {
-        this.$nextTick(() => {
-          if (this.$refs.noteInput?.$refs.input?.$refs.input) {
-            this.$refs.noteInput.$refs.input.$refs.input.focus()
-            setTimeout(() => {
-              this.$refs.noteInput?.$refs.input?.$refs.input?.select()
-            }, 10)
-          }
-        })
-      }
+      // Try to get fresh device data from Capacitor plugin
+      this.$db.getDeviceData().then(deviceData => {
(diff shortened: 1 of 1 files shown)


advplyr:
I don't think something this minor should be a setting. Let me see if there is another solution

advplyr:
I resolved this in #1612 and mentioned we could go for a different design if users prefer the keyboard to open.

I don't think this PR works either because the input had autofocus on it. So the first time it opens it would autofocus even if this setting was enabled.

nichwall:
I agree that this should not be a setting and removing the auto focus entirely is fine, but still feels weird to me to not have a pop-up with only a text box not being auto focused. If we still want an auto focus option, maybe we could have a separate button that is just "create bookmark with current datetime as title" or something.

Is there a reason your keyboard is set to the light theme instead of a dark theme if you don't want a bright white box appearing in the evening/at night?

abguymon:
@advplyr ah yes this resolves it for me. You got to it just before me. I'll close this

@nichwall I know for my case just creating a bookmark with current datetime as title is all I need. I would imagine that's a common usecase

PR #1660, opened 2025-08-13
Fix OIDC button not appearing during re-login when refresh token is missing

jtbrush56:
## Problem
The OIDC login button was not appearing when manually reconnecting to servers that support OpenID authentication, specifically when refresh tokens were missing or expired.

## Root Cause  
The `connectToServer` method only called `/ping` and `/api/authorize` endpoints during reconnection, but never fetched authentication methods from `/status`, leaving `authMethods` empty.

## Solution
- Added call to `getServerAddressStatus()` when token authentication fails
- Populates `authMethods` array before showing auth form  
- Includes robust error handling for graceful fallback
- Reuses existing proven logic for consistency

## Changes
- **File:** `components/connection/ServerConnectForm.vue` (lines 474-490)
- **Backward Compatible:** ✅ No breaking changes
- **Error Handling:** ✅ Graceful fallback to local auth

## Result
Users can now see and use the OIDC button when reconnecting to servers after refresh token expiration.

### DIFF
--- a/components/connection/ServerConnectForm.vue
+++ b/components/connection/ServerConnectForm.vue
@@ -471,6 +471,21 @@ export default {
         // Will NOT include access token and refresh token
         this.setUserAndConnection(payload)
       } else {
+        // Authentication failed, fetch server auth methods before showing auth form
+        try {
+          this.processing = true
+          const statusData = await this.getServerAddressStatus(this.serverConfig.address)
+          if (statusData && statusData.data) {
+            this.authMethods = statusData.data.authMethods || []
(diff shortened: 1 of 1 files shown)


nichwall:
I believe this is included in https://github.com/advplyr/audiobookshelf-app/commit/4be1598ecaff6b6f154ced9e11623d955f398fea

jtbrush56:
I think you might be right. We just handled it different ways. If this PR isn't needed, it can be closed.

advplyr:
Ah yeah I resolved this for the next release. Thanks!

PR #1666, opened 2025-08-20
Allows widget on Samsung cover screens

mfulgo:
## Brief summary
Adds meta-data to the widget, allowing it to be used on the cover screen of Samsung devices such as the Flip.

The widget renders as a 4x4 grid (not an optimal layout) but functions for launching the most recent audiobook or controlling playback.

## Which issue is fixed?

n/a

## Pull Request Type

Android, frontend (widget)

## In-depth Description

(see summary)

## How have you tested this?

Built the APK locally and installed on a Flip 7. I was able to add the widget as page on the cover screen (when the phone is closed) and confirmed the home screen widget remained functional.

## Screenshots

image

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -36,6 +36,10 @@
           <meta-data
             android:name="android.appwidget.provider"
             android:resource="@xml/media_player_widget_info" />
+
+          <meta-data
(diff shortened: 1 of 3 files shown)


nichwall:
I am not able to test this on a physical device but nothing seems to be broken on my Pixel 6a. Seems good to me.

advplyr:
Thanks!

Issue #1323, opened 2024-09-16
 [Enhancement]: Search function within the ebook reader


mrjfilippo:
### Describe the Feature/Enhancement

To be able to search within an ebook. Showing how many results, highlighting the results, jumping to the next result, and seeing a list of results.

### Why would this be helpful?

An easy and convenient way to find a section of a book.

### Future Implementation (Screenshot)

FBReader on Windows implementation includes most except the result list.

![image](https://github.com/user-attachments/assets/aed34f7c-175e-408b-a611-83d28ed80989)

Calibre's reader opens up a whole side menu for it:

![image](https://github.com/user-attachments/assets/817f1697-2b6f-4d5a-8b6e-3d245bdf1f69)

A mix of both would be interesting. An magnifying icon at the top (or search bar), and menu come up from the bottom like the chapter and settings menu already in the reader.

![Screenshot_20240916-225833_Audiobookshelf](https://github.com/user-attachments/assets/9220140c-3cba-4704-b56e-c46e97af7a27)

### Audiobookshelf App Version

Android App - 0.9.76

KaiStarkk:
Appears possible to implement, see [this example ](https://github.com/futurepress/epub.js/wiki/Tips-and-Tricks-(v0.3)#searching-the-entire-book) in upstream's wiki.
It's on my holiday list to get around to some e-reader improvements like this too.

advplyr:
Searching is implemented in the web client already

SmoochSmackers:
This feature is literally the only thing that this app is missing. 

Use-case:

1. If a narrator has an accent that makes a specific word particularly difficult to understand, one can search a phrase leading up to said word in the ebook to find the word. Or...

3. Get the spelling of a character's, or alien species', name for texting or searching. This would be a godsend.

This feature would be a godsend.

Absolutely flawless app otherwise!

PR #1667, opened 2025-08-22
add search inside epub files. #1323

reda-akherfi:
## Brief summary

add the search functionality inside the epub reader on the official android client.

## Which issue is fixed?

it solves this issue: #1323

## Pull Request Type

I only tested it on android, but I guess it must work on IOS because of capacitor.

## In-depth Description

you can click on the search icon, enter a query in the search modal and then select among the matches. there are also, two more buttons as part of the controls: a next and previous occurrence buttons. the search query gets highlighted. the modal remembers the last search query per book. 

## How have you tested this?

opened multiple epub books from my library, and tested all the new features again and again untill I  was satisfied with the results.
tested it on my physical Android 10 phone and  a virtual Android 14 one.
## Screenshots

WhatsApp Image 2025-08-22 at 11 25 54_ce373864
WhatsApp Image 2025-08-22 at 11 25 56_027d1d4a


advplyr:
I'm testing in the web client and it feels buggy but I'm not sure if it was intentional. For example, if I have the search modal open and tap outside of it I would expect it to close, but it opens up the top bar instead.

It is also preserving the previous search even after changing books.

I don't think it's useful for the CFI to show in search results.

Have you checked out the epub search in the web client? It may be useful to base the search off of that implementation or plan to use your implementation the web client so they function the same.

PR #1672, opened 2025-08-27
Made the book selectable.

### LINKED ISSUE #606: Crashes immediately on Android TV 12
Love your app, but it crashes immediately on ATV. Any plans for support on ATV?

### Steps to reproduce

1. Install
2. Open
3. Observe crash

### Expected behaviour
- Tell us what should happen

At least open

### Actual behaviour
- Tell us what happens

Crashes immediately 

### Environment data

Rooted Lineageos 19 kernel version 4.9.113

Audiobookshelf Version: 0.9.62-beta

- [ X] Android App?
- [ ] iOS App?

#### Android Issue
Android version: ATV 12

Device model: ODroid-n2+

Stock or customized system: Custom rooted Lineageos 19 ATV 12 on ODroid-n2+

monksy:
## Brief summary

This makes the book items selectable and navigable within Android TV situations.

## Which issue is fixed?

Fixes #606 

## Pull Request Type

This affects Android TV / devices without touch support and navigation by elements.

## In-depth Description

Your current setup works for the navigation elements because they used nxut-link/anchors. By default the browser treats those as selectable. This enables tabindexs for the book elements so they are selectable and "clickable" in interfaces that don't allow for touch/clicking. 

## How have you tested this?

Android emulator

## Screenshots

Can't really show something that doesn't work before after pressing. 

After 

image


advplyr:
I got android tv device for android 16 in the android studio emulator for testing.

I'm not seeing any difference running this PR compared to the master branch.
When using the extended controls I cannot select book cards. I can only select navigation tabs.

image

I'm also not sure about the two features you added to the manifest because I don't see any difference with the emulator with those there.

I don't have an android tv to test with

advplyr:
I played around with this some more and did some reading. I think for this to work without re-designing the web ui we have to set up custom listeners for d-pad events and pass those to webview. The javascript frontend would then determine where to shift focus.
There is basically no info on handling d-pad controls in webview but there are a lot of unanswered threads of people asking about it.

monksy:
I don't have an easy way of testing Android TV so I can't correct you what
you said. How I did test this is that I used my keyboard in the emulator
with the direction buttons. I found that the navigation bar was
highlighted... Later I found out that the highlighting was due to tab
order/indexing. By default divs art.

The feature changes are because of the requirements to get approval for
Android TV.

monksy:
According to chat gpt this works despite an weird ordering: (I'm out of my
element here)

tabindex

What it does:
tabindex is a standard HTML attribute that controls focus order in the DOM.
It tells the browser (or WebView in Android TV) which elements are
"focusable" and in what sequence.

Behavior on Android TV:

The WebView will use tabindex to decide where focus should move when the
system interprets DPAD navigation.

If you don’t set it, focus may default to clickable elements (like buttons
or inputs), and the order can feel inconsistent.

With tabindex, you can create predictable focus flows (e.g., left → right →
down

advplyr:
That's mostly incorrect. I did a bunch of testing yesterday and D-Pad doesn't follow tabindex.
I'm going to close this out since it doesn't resolve the issue but I did add the tabindex to the book cards since there may be some cases where that is used on mobile devices.
https://github.com/advplyr/audiobookshelf-app/commit/95bf28d7c6176582063e560c61f18edbf3da6029

bilbospocketses:
apk for sideloading is available for Android TV.

It needs to be tested on devices beyond my Google TV Streamer 4K, but on that device it works pretty well, so it should be fine on standard Android and Google TV devices back to Android 8. Not sure how well it will work on Fire TV, but am interested to hear any feedback.

All feedback is appreciated over on my forked repo so I can continue to tune the app while the devs of the main repo work on reviewing and merging the PR.

APK: https://github.com/bilbospocketses/audiobookshelf-app/releases/

Issues page for feedback (bugs, enahncements, etc): https://github.com/bilbospocketses/audiobookshelf-app/issues

Pull Request: https://github.com/advplyr/audiobookshelf-app/pull/1843

User guide for the ATV app: https://github.com/bilbospocketses/audiobookshelf-app/blob/android-tv-dpad-navigation/docs/TV_USER_GUIDE.md

PR #1675, opened 2025-08-28
Remove busy loops

### LINKED ISSUE #1446: [Bug]: Android battery usage
### What was the Problem?

I'm having battery usage issues, the usage has been much higher than in previous app versions. After about two hours of playtime, it has been running in the background for 6+ hours and accounted for 35% of my battery usage, five times the usage of any other app.

Use the app on a similar phone

### What was Expected?

Less battery usage

### Phone Model

Pixel 7a

### Phone OS

Android 15

Android App - 0.9.78

### Installation Source

Google Play Store

### Additional Notes

I do have the widget enabled, I've removed it to see if that makes any difference.

laurisvr:
## Brief summary
Fixes high CPU usage after connecting to Android Auto during playback.  
Removes busy‑wait loops in `PlayerNotificationService` and in library personalization code, replacing them with callback‑based async handling.

## Which issue is fixed?
Fixes #1446

## Pull Request Type
- Platform: **Android**
- Area: **native service / Android Auto media browser**

## In-depth Description
Previously the code used tight `while` loops to wait for async work:
- `while (libraryPersonalizationsDone > 0) { }`
- `while (!this::browseTree.isInitialized) { }`

If one callback failed or init was delayed (common when Android Auto connects/rebinds services), these loops never exited, pegging the CPU at ~100%.  

This PR removes the spin‑loops:
- Personalization loading now uses a counter (`AtomicInteger`/callback) to invoke completion when all libraries finish.
- BrowseTree initialization inside `onLoadChildren` uses `result.detach()` + a small helper (`waitForBrowseTree/onBrowseTreeInitialized`) to defer sending results until `browseTree` is ready.

This avoids ANRs and runaway CPU usage, and aligns with MediaBrowserService’s async contract.

## How have you tested this?
- Built and ran Audiobookshelf Android app locally in debug.
- Reproduced bug on master: while playback is active, connect Android Auto via Desktop Head Unit → CPU spikes >100% and app freezes.
- Applied fix:
  - Reconnected Android Auto during playback.
  - CPU remains stable, UI responsive.
  - Verified browse menus display correctly after `browseTree` initializes.
  - Verified personalization logs show all libraries finishing.

## Screenshots
N/A – changes are to background service / logs only.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/media/MediaManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/media/MediaManager.kt
@@ -15,6 +15,7 @@ import org.json.JSONException
 import org.json.JSONObject
 import kotlin.coroutines.resume
 import kotlin.coroutines.suspendCoroutine
+import java.util.concurrent.atomic.AtomicInteger
 
 class MediaManager(private var apiHandler: ApiHandler, var ctx: Context) {
   val tag = "MediaManager"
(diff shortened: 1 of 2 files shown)


nichwall:
I can confirm that high battery usage went away on Android 16 with a Pixel 6a after using Android Auto.

The time of the app being active in the background still goes up, so I think the app might still be getting stuck at the event wait, but it does not have high battery usage anymore which is great.

advplyr:
Nice, thanks!

laurisvr:
> I can confirm that high battery usage went away on Android 16 with a Pixel 6a after using Android Auto.
> 
> The time of the app being active in the background still goes up, so I think the app might still be getting stuck at the event wait, but it does not have high battery usage anymore which is great.

Thanks for testing and confirming! I agree there’s still an underlying issue — the event isn’t being resolved as expected. Previously this led to a busy spin‑loop and high CPU usage, but with the change it no longer spins, so the battery drain is gone. That said, the service still seems to remain “awake” while waiting on that unresolved event.   

The next step will be to dig into why the event isn’t completing so the service can safely return to idle. And honestly, this whole file has grown pretty complex over time, so some refactoring would also help make these kinds of lifecycle and event‑handling problems easier to reason about.

PR #1679, opened 2025-09-04
feat: add logs on iOS

clayreimann:
## Brief summary

- enable logs page on iOS
- wrap long identifiers in logs UI
- persist logs in Realm
- update editorconfig, podfile, and readme

## Pull Request Type

iOS only

## In-depth Description

Adding the logs to the iOS build to help with tracking down player sync issues.

## How have you tested this?

Running on my device

## Screenshots

image

### DIFF
--- a/ios/App/App/plugins/AbsLogger.swift
+++ b/ios/App/App/plugins/AbsLogger.swift
@@ -54,7 +54,14 @@ public class AbsLogger: CAPPlugin, CAPBridgedPlugin {
         entry.level = "\(level)"
         entry.timestamp = Int(Date().timeIntervalSince1970 * 1000)
         try Database.shared.saveLog(entry)
-        self.notifyListeners("onLog", data: ["value": entry])
+        
+        self.notifyListeners("onLog", data: [
+            "id": entry.id,
+            "tag": entry.tag,
+            "message": entry.message,
+            "level": entry.level,
+            "timestamp": entry.timestamp
+        ])
     }
     
     public func info(tag: String = "\(#file)", message: String) throws {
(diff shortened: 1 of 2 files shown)


Vito0912:
Wouldn't it make more sense to replace the current logger or connect to it (if possible) like it was done with Android?

clayreimann:
> Wouldn't it make more sense to replace the current logger or connect to it (if possible) like it was done with Android?

Not sure what you mean. Like replace the `let logger: AppLogger` instances in each class and just funnel all logs through this the AbsLogger? I've started doing that, but funneling everything to a single AppLogger loses some of the granularity in filtering since it's all coming from the same place. Like not creating a new SLF4J logger per class (which is customary on all the java/scala codebases I've worked on).

It might be a good idea to have an option to suppress the console logging for other native code calling the logger though, to prevent double log lines.

Edit: though of a way to both replace the current logging with the plugin and maintain specificity. Will update PR.

advplyr:
I just had to fix the onLog listener. It is working great. The logs will need to get cleaned up eventually though. On the Android side I chose to migrate only some logs to the new logger to make it easier to debug the sync errors that were getting reported.

Thanks!

clayreimann:
@advplyr this should be happening during app start in [Database.init](https://github.com/advplyr/audiobookshelf-app/pull/1679/files#diff-d0318a04b57012cef1d61a8570f7349a4f8ce68e8e82e67de6d6b0daeace6c94R18)

Edit: Unless you mean that the things that are being logged need to get cleaned up, then yeah probably (or just make more judicious use of log levels)

advplyr:
Yeah cleaned up

PR #1683, opened 2025-09-10
Add chapter duration to chapters table on book page

sir-wilhelm:
This adds the chapter duration to the chapter table on the book page.

## Brief summary

This aims to resolve #1163 by adding chapter duration to the chapters table on the book page.

## Which issue is fixed?

Fixes #1163.

## Pull Request Type

This changes the front-end of the android and iOS app.

## In-depth Description

It just adds the duration to the ChaptersTable view, the same way it's done on the web page.

## How have you tested this?

I tested this in Android Studio, on 2 different size emulated screens.

## Screenshots

<details><summary>Original table</summary>

pixel5:
Screenshot_px5-chaptersTableOriginal
small phone:
Screenshot_small-chaptersTableOriginal
</details>

<details><summary>Table with Duration</summary>

pixel5:
Screenshot_px5-chaptersTableDuration
small phone:
Screenshot_small-chaptersTableDuration
</details>

Issue #1163, opened 2024-04-03
The ability to see the duration of chapters in the book's chapter list vs. seeing the timestamp of where in the book the chapter starts


cslamar:
### Is your feature request related to a problem? Please describe.
Currently, when you are in the book detail page, viewing the chapters shows the timestamp of where the chapter starts in the overall audio file (eg. 1:12:35). 

### Describe the solution you'd like
 It would be nice to see, or to be able to choose between, the duration of the chapters (eg. 43min) instead of the timestamp position.

### Additional context
Screenshot included as a reference to show the current behavior:

![chapter-screen](https://github.com/advplyr/audiobookshelf-app/assets/3410502/f54a9d09-6ad4-4492-940d-207edc0d8393)

advplyr:
I'm looking at this and not seeing a good way to fit both of them in there so like you said we would have to add a setting to switch between the two. I'm not sure where to put that setting. It doesn't seem like something that would be on the settings page. Adding a menu button on the table header feels like overkill, so not sure about this one.

cslamar:
My two cents would be to put the toggle in a global settings menu.  My rational is that it's either you want it to be displayed, or you don't.  Probably not something that someone would be toggling in the book view or the player.

sir-wilhelm:
I've gotten use to going to the web view if I wanted to see the duration, but it would be nice to have a setting to use duration instead (or just show both like the web view). I could also see this setting being used on the chapter view of the player.

<details><summary>Potential chapter duration view (mimic webpage)</summary>

current app:
![Image](https://github.com/user-attachments/assets/1250147f-ac34-4b88-9067-6ce3740ed1a3)
webpage:
![Image](https://github.com/user-attachments/assets/73b84bc8-ba2e-447e-b637-9494fcba24cf)
</details>

<details><summary>Chapter view of player (would be nice to have a setting to view duration instead)</summary>

![Image](https://github.com/user-attachments/assets/01169206-e7f0-443f-a8c3-a6a08ba6c516)
</details>

sir-wilhelm:
It'll be easy to add the duration to the chapter table on the book page (mirroring how it looks on the webpage), so I'll make a PR for that.

I think I have changes ready to go for the mobile app too, but I can only test the android version. I'll make a 2nd PR for that change.

If you want the global settings option to be used for both we can trash the first PR and I can update the 2nd one to use the setting on the chapters table too (but IMO it looks fine w/ both columns -- I tested on a small/large emulator).

PR #1685, opened 2025-09-10
Adds a player preference to use chapter duration in the chapters modal view

sir-wilhelm:
This is a sister-PR to #1683, where it continues to add options/places to use chapter duration instead of start time in views.

## Brief summary

This adds a player preference `"Use chapter duration"` that will swap out the chapter Start time w/ Duration.

## Which issue is fixed?

Fixes/expands on #1163

## Pull Request Type

This impacts both android/iOS clients via Vue changes.

## In-depth Description

The duration shown does take into account the playback speed, and will be adjusted as needed.

## How have you tested this?

I only tested the android version in Android Studio, so I'll need help testing this on iOS/a review to know if this needs more changes.

## Screenshots

<details><summary>Original chapters modal with Start time</summary>

Screenshot_px5-chaptersModalOriginal
</details>

<details><summary>Player chapters modal with Duration</summary>

1x playback speed:
Screenshot_px5-chaptersModalDuration1x
2x playback speed:
Screenshot_px5-chaptersModalDuration2x
</details>

<details><summary>Player settings page with new option:</summary>

image
</details>

### DIFF
--- a/components/app/AudioPlayer.vue
+++ b/components/app/AudioPlayer.vue
@@ -99,7 +99,7 @@
       </div>
     </div>
 
-    <modals-chapters-modal v-model="showChapterModal" :current-chapter="currentChapter" :chapters="chapters" :playback-rate="currentPlaybackRate" :use-chapter-duration="playerSettings.useChapterDuration" @select="selectChapter" />
+    <modals-chapters-modal v-model="showChapterModal" :current-chapter="currentChapter" :chapters="chapters" :playback-rate="currentPlaybackRate" @select="selectChapter" />
     <modals-dialog v-model="showMoreMenuDialog" :items="menuItems" width="80vw" @action="clickMenuAction" />
   </div>
 </template>
@@ -146,7 +146,6 @@ export default {
         useChapterTrack: false,
         useTotalTrack: true,
         scaleElapsedTimeBySpeed: true,
-        useChapterDuration: false,
         lockUi: false
       },
       isLoading: false,
@@ -206,11 +205,6 @@ export default {
             text: this.$strings.LabelScaleElapsedTimeBySpeed,
             value: 'scale_elapsed_time',
             icon: this.playerSettings.scaleElapsedTimeBySpeed ? 'check_box' : 'check_box_outline_blank'
-          },
-          {
-            text: this.$strings.LabelUseChapterDuration,
-            value: 'use_chapter_duration',
-            icon: this.playerSettings.useChapterDuration ? 'check_box' : 'check_box_outline_blank'
           },
           {
             text: this.playerSettings.lockUi ? this.$strings.LabelUnlockPlayer : this.$strings.LabelLockPlayer,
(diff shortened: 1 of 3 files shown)


clayreimann:
@sir-wilhelm as is this crashes on iOS. Looks like a DB migration is required. I don't have time just now to provide the patch, but can probably get to it this weekend.

sir-wilhelm:
@clayreimann thanks for testing. If you could provide the patch to get it working on iOS that would be great.

sir-wilhelm:
After looking at git history for other Device Settings changes, I think I found where the schema is done, i pushed a change with what I think will fix it for iOS if you want to pull [AppDelegate.swift changes](https://github.com/advplyr/audiobookshelf-app/commit/6e66169c0de2608ac0a09b38b12ccf73b63d3132#diff-b3d33b076cd75a857a3a4420b5cdbaa1d565191370a89f6f91466878cab222ee).

ios/App/App/AppDelegate.swift
```diff
@@ -73,6 +73,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                         newObject?["version"] = ""
                     }
                 }
+                if (oldSchemaVersion < 21) {
+                    self?.logger.log("Realm schema version was \(oldSchemaVersion)... Adding useDurationInChaptersModal to ServerConnectionConfigs")
+                    migration.enumerateObjects(ofType: ServerConnectionConfig.className()) { oldObject, newObject in
```
clayreimann, line 78:
```suggestion
                    migration.enumerateObjects(ofType: DeviceSettings.className()) { oldObject, newObject in
```

clayreimann (commented):
Almost the right migration

clayreimann:
With the migration this works as intended though

28FC39FD-763A-42E2-9471-F0465B1D01EB_4_5005_c
3E479E32-3DDD-453D-907A-94F6D2C28CF3_4_5005_c

sir-wilhelm:
Thanks for the help with the iOS portion of this; I think it's good to go now.

sir-wilhelm:
After a discussion in Discord:
> [4:22 PM]advplyr: I just took a look at 1685 and I think if that setting is a UI only change it would be a better fit for the `localStore`. You can see that is where the theme is getting stored. And in `AudioPlayer.vue` you'll find a bunch of the UI only player settings use localStore.

I'll take a look at making those changes.

sir-wilhelm:
After taking @advplyr's suggestions the change set is quite a bit smaller and only impacts vue code, so it should just work on android and iphone. I only tested it on an android emulator, and updated the Player Settings page image/issue description.

advplyr:
I found the setting to be confusing. It's not clear what "Use chapter duration" is doing.

I decided to always include the duration in the chapter modal and updated the chapter title to support being on 2 lines. This will allow for longer chapter titles to still show clearly. In order to make it clear what the numbers were I added a sticky header that has a drop shadow when scrolled.

image

drop shadow
image

I should have put this in a separate PR since it is completely different but it accomplishes the same goal I think.

Thanks!

Issue #1373, opened 2024-11-11
[Enhancement] Add font selection setting in epub e-reader to match web client


ErnieBoBernie:
### What was the Problem?

After the last app update I am unable to change my font selection in the ABS app. This happens on both android and ios versions.  Seems to coincide with the addition of page turn using volume buttons option. 

### Steps to Reproduce the Issue

Open epub book to read.  Select options to change font from Sans to Serif. There is no option available to change font. 

### What was Expected?

Expected to have an option to change the font format

### Phone Model

iphone 15 plus, and Lenovo m9 tablet

### Phone OS

iOS 18.1 and Android 13

### Audiobookshelf App Version

Android App - 0.9.77, iOS App - 0.9.77

### Installation Source

Google Play Store, Testflight

advplyr:
Changing font has not been added to the mobile apps yet. Currently only available in the web client

weilBaum9:
Are there any updates regarding this feature?

PR #1688, opened 2025-09-16
Add font selection setting in epub e-reader to match web-client (#1373)

weilBaum9:
## Brief summary

Add a font selection setting to the epub e-reader to swap between a font with and without serifs as implemented in the web-client. 

## Which issue is fixed?

Implements #1373

## Pull Request Type

Frontend-change for both IOS and Android App

## In-depth Description

The solution gives the user the option to select between a font with or without serifs. It was done this way to mirror the funtionality and implementation of the web-client. 

## How have you tested this?

1. Started App in Android Studio Emulator
2. Connected to Server
3. Opened Book
4. Swapped between sans and serif font 

I don't have access to any apple device, so I couldn't test the implementation for ios devices. 

### Before:
after_sans

### After: 
<p float="left">
  after_sans
  after_serif
</p>

### DIFF
--- a/components/readers/Reader.vue
+++ b/components/readers/Reader.vue
@@ -262,11 +262,11 @@ export default {
     fontItems() {
       return [
         {
-          text: 'Sans',
+          text: this.$strings.LabelFontFamilySans,
           value: 'sans-serif'
         },
         {
-          text: 'Serif',
+          text: this.$strings.LabelFontFamilySerif,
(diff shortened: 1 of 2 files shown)


nichwall (commented):
Looks good to me and works in my testing.

The only thing is to make the option selections able to be localized.

advplyr:
Thanks!

Issue #749, opened 2023-06-20
Skip op & ed automatically


dboyplus:
I would like to be able to set all the audio tracks in a specific album to automatically skip a certain duration. This would help save time when listening to audiobooks that have fixed-length intros and outros.

Yinghai75:
This is exactly the feature I want the most.
I'd like to be able to set, likely, 15s for intros and 10s for outros etc.

Yinghai75:
iOS users can try the DS Cloud app — it’s much better than the official one!
Even books with 2,000 chapters can start and resume playback instantly, and each book lets you set skip times for the beginning and ending!

PR #1689, opened 2025-09-19
Auto skip intro by chapter start and end property

ZaneYork:
## Brief summary
Add a player logic, when currentTime is not within the chapter's begin and end, skip to next chapter's beginning.

## Which issue is fixed?
Implements https://github.com/advplyr/audiobookshelf-app/issues/749

## Pull Request Type
Player's seek logic for both IOS and Android App

## In-depth Description
The solution is a partial logic of mobile side to auto skip op and ed. Currently, we need a extra script to modify book's metadata to modify chapter's begin and end properties.

Here is the script i'm using now (need to rescan to let audiobookshelf read the change):
```
import json
import os
import shutil
import sys

def modify_json_file(file_path, offset_a, offset_b):
    # backup original file
    backup_path = file_path + '.bak'
    shutil.copy2(file_path, backup_path)
    
    # read json file
    with open(file_path, 'r', encoding='utf-8') as f:
        data = json.load(f)
    
    # modify chapters start and end property
    if 'chapters' in data:
        for chapter in data['chapters']:
            if 'start' in chapter:
                chapter['start'] += offset_a
            if 'end' in chapter:
                chapter['end'] -= offset_b
    
    # write back
    with open(file_path, 'w', encoding='utf-8') as f:
        json.dump(data, f, indent=4, ensure_ascii=False)

if __name__ == '__main__':
    if len(sys.argv) != 4:
        print("Usage: python json_modifier.py <json_file_path> <offset_A> <offset_B>")
        sys.exit(1)
    
    try:
        json_path = sys.argv[1]
        offset_a = float(sys.argv[2])
        offset_b = float(sys.argv[3])
        
        if not os.path.exists(json_path):
            raise FileNotFoundError(f"JSON file not found: {json_path}")
            
        modify_json_file(json_path, offset_a, offset_b)
        print(f"Successfully modified {json_path} and created backup")
        
    except Exception as e:
        print(f"Error: {str(e)}")
        sys.exit(1)
```

## How have you tested this?
* Started App in Android device
* Connected to Server
* Opened Book
* Play to chapter's ending
* Wait untill it auto skip to next chapter's beginning

### DIFF
--- a/components/app/AudioPlayer.vue
+++ b/components/app/AudioPlayer.vue
@@ -158,7 +158,8 @@ export default {
       coverRgb: 'rgb(55, 56, 56)',
       coverBgIsLight: false,
       titleMarquee: null,
-      isRefreshingUI: false
+      isRefreshingUI: false,
+      introSkipping: false
     }
   },
   watch: {
@@ -565,6 +566,7 @@ export default {
 
       this.updateTimestamp()
       this.updateTrack()
+      this.updateIntroSkip()
     },
     updateTrack() {
       // Update progress track UI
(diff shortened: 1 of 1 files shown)


nichwall:
I'm confused how this is supposed to work. Am I correct that what you are doing is running a Python script to manually modify the JSON metadata files and scanning these changes back into the database, and then just skipping ahead to the "first chapter start time" in the client?

The server and player expects the first chapter of every book to start at time 0, so needing to run a script to modify JSON files without going through a supported API endpoint is not a good way to handle this. This should be an actual setting on the client to handle skipping introductions, but this would need to happen on the server first to allow marking chapters as "introduction" or similar. Another option would be adding a setting to the client that just always skips the first X seconds of all books.

PR #1704, opened 2025-10-20
feat: add playback queue

andykelk:
## Summary

Adds a playback queue feature allowing users to line up items to play sequentially. The queue persists across sessions and is managed at both the JavaScript/Vue layer and the native Android layer for seamless background playback.

- **Queue management UI**: New `QueueModal` component with drag-to-reorder, play from position, and individual item removal. Accessible via a queue button in the audio player, side drawer, and item more menus
- **Add to queue**: Queue buttons added to item/episode more menus, podcast latest episode rows, and the episode detail page
- **Persistent storage**: Queue is saved to Capacitor `Preferences` (SharedPreferences on Android) and restored on app launch
- **Native Android auto-advance**: `QueueManager` reads the queue directly from SharedPreferences so the Android `PlayerNotificationService` can advance to the next item on playback completion, even when the app is backgrounded
- **Progress preservation**: Current playback position is saved in the queue item so playback resumes from where it left off
- **Native→JS sync**: When the native layer advances the queue, it emits `onQueueChanged` so the Vue store reloads from storage and stays in sync

## Changed files

- `store/index.js` — queue state, getters, mutations and actions
- `store/globals.js` — `showQueueModal` flag
- `plugins/localStore.js` — `getPlaybackQueue` / `setPlaybackQueue` methods
- `mixins/queueMixin.js` — shared `addItemToQueue` helper used across components
- `components/modals/QueueModal.vue` — queue management modal
- `components/app/AudioPlayer.vue` — queue button + `onQueueChanged` listener
- `components/app/SideDrawer.vue` — queue menu item
- `components/modals/ItemMoreMenuModal.vue` — "Add to Queue" action
- `components/tables/podcast/LatestEpisodeRow.vue` — queue button
- `pages/item/_id/_episode/index.vue` — "Add to Queue" action
- `layouts/default.vue` — mount `QueueModal`, load saved queue on init
- `android/.../data/PlaybackQueueItem.kt` — data model matching JS queue item structure
- `android/.../managers/QueueManager.kt` — reads/writes queue from SharedPreferences
- `android/.../media/MediaProgressSyncer.kt` — mark finished even when timer is stopped
- `android/.../player/PlayerNotificationService.kt` — native queue auto-advance on playback end
- `android/.../player/AbMediaDescriptionAdapter.kt` — `invalidateCache()` for cover art refresh
- `android/.../plugins/AbsAudioPlayer.kt` — `onQueueChanged` event bridge
- `strings/en-us.json` — `ButtonAddToQueue`, `ButtonClearQueue`, `LabelQueue`, `MessageQueueEmpty`

## Test plan

- [ ] Add an audiobook to the queue from the item more menu; verify it appears in the queue modal
- [ ] Add a podcast episode to the queue from the episode page and latest episodes list
- [ ] Drag items to reorder in the queue modal; verify order is persisted after closing and reopening
- [ ] Delete an item from the queue; verify it is removed
- [ ] Play an item from the queue modal while something is already playing; verify the current item is added back to the front of the queue
- [ ] Let playback finish; verify the next queue item starts automatically
- [ ] Force-close the app and reopen; verify the queue is restored from storage
- [ ] Test background auto-advance (screen off, app backgrounded) on Android


android/app/src/main/java/com/audiobookshelf/app/player/AbMediaDescriptionAdapter.kt
```diff
@@ -91,4 +91,9 @@ class AbMediaDescriptionAdapter (private val controller: MediaControllerCompat,
       }
     }
   }
+
+  fun invalidateCache() {
```
nichwall: This function is not referenced anywhere, is it intended to be used?

Atinoda:
Thank you @advplyr for Audiobookshelf, and @andykelk for this feature!

I was keen to get using this (currently listening to a podcast with frustratingly short episodes...) so I rebased the feature branch to master and built an apk. It is working well so far. Will report back with any bugs and some feedback after using it for a while. If there's anything you want me to look at in particular, please let me know.

Rapha149:
Hi,
firstly I want to say that I really appreciate your work on this since a queue is the main thing I miss for podcasts on ABS.
However, I would like to suggest two changes:

1. Add a button for enqueuing an episode in the podcast overview page where the episodes of that podcast are listed, like in the latest page. 

2. For the button in the latest page: I find the popups that appear when adding episodes to the queue a bit annoying, especially when I enqueue multiple episodes, as the popups quickly take up half the screen.  
I would suggest this alternative solution: don't allow multiple instances of the same episode / audiobook in the queue, as in my opinion this wouldn't make much sense anyway for these media types. Then the button on the latest page could get an indicator whether the episode is currently enqueued or not, e.g. a green color when it is. The color of the button would be direct feedback and could replace the popup. If it is pressed again there are two options: either the episode is removed from the queue or the queue display opens.  
In the same manner the "Add to queue" button in the three dot menus could be replaced by "Remove from queue" when the episode or audiobook is already enqueued.

Anyway these are just suggestions, please let me know what you think!

andykelk:
> Hi, firstly I want to say that I really appreciate your work on this since a queue is the main thing I miss for podcasts on ABS. However, I would like to suggest two changes:

Good ideas. I have incorporated these in the newest commit.

Rapha149:
Wow, that was fast! Thank you, it looks very good.

One small thing I also noticed: there is no haptic feedback for any of the buttons. For example, the button for marking an episode as finished provides haptic feedback (in the latest page as well as the three dot menu). Adding this would be a very small change, if I'm not mistaken all you would have to do is make the methods `async` and add `await this.$hapticsImpact()`.

Atinoda:
Hi Andy, thank you for working on this!

My apologies that I never got back to you, but I will echo the comments of Rapha from my experience, and having just built and tested the app the fixes are lovely and work well.

- **Previous and Next track buttons**: These do not seem to work for queue navigation, but I think they did before. Maybe a regression?

- **Auto-enqueue podcast episodes as filtered and displayed**: Something to consider - perhaps a related PR is #1816. Way back when, this was the option / behaviour I expected from the app, and it would align to the web client behaviour. Then I discovered there is no queue (sad), then I discovered you had made one (happy). Along that journey, it seems I forgot where the first footstep was placed.

- **Progress / complete status flags and the sleep timer auto-rewind feature**: My use-case was sleepytime listening to shorter podcasts (multiple episodes and fiddling with the buttons at the end of each was annoying). My (admittedly bleary-eyed) testing seemed to have surfaced an edge case. If the rewind timer horizon goes back beyond the current episode, I don't think that progress and completion flags are are being managed. For example, queue of E1 and E2 both episodes have a 15m length; sleep timer of 25m then rewind timer of 20m; the playback sequence is `{E1 [complete], E2 [progress 10m]}` and the rewind timer should then reset this to `{E1 [incomplete, 5m progress], E2 [progress none]}` Honestly though - any time this is relevant I have literally fallen asleep so this is vaguely remembered! Given how fiddly that is to describe, the logic to address that might be torturous with the required queue trajectory memory and state tracking... so I'm just raising it to you but not necessarily suggesting it is something worth working on.

Thanks again for your work in developing this feature!

PR #1727, opened 2025-11-16
Pause and end playback if the Vue layer receives a progress update wh…

### LINKED ISSUE #1385: [Bug]: iOS: Frequently losing progress
### What was the Problem?

I've been using audiobookshelf for about 6 months quite heavily.  I often move between airpods and my car while listening to an audiobook.  Fairly often I will pause, switch to airpods/car, hit play and nothing will happen.  When I open the app on the phone I see a red toast error at the bottom (I will update with a screenshot once I see it again) and all of my recent progress is lost.  This has happened probably 30 times since I've started using the app.

Sorry, this is not consistently reproducible.

Just today I noticed something interesting.  Here's how my day went:

1. Listen in car on drive to work, paused with car button.
2. Put in airpods, clicked the airpod button to continue playing while walking from car to office.
3. Paused by clicking the airpod button.
4. Worked 8 hours without any listening.
5. Started walk back to car, put in airpods, hit airpod button to "resume" as I did not listen to any other media that day, it started playing the audiobook.
6. Once I got to my car, I knew this bug often happens at this point so I decided to open the app while the audiobook was playing and the app did not have the player visible at the bottom of the screen and the book I was listening to had a big play button even though the audiobook was currently playing.  The iOS "media notification" was visible and accurate, as far as I could tell.
7. Paused with airpod button.
8. Got into car, all progress from my walk from the office to the car was lost.

### What was Expected?

My progress would be saved as I switched between airpods and car.

### Phone Model

iphone 15 Pro Max

### Phone OS

iOS 18.1

iOS App - 0.9.77

### Installation Source

Testflight

### Additional Notes

This has happened on every iOS version I've been on for the last 6 months.  (17 and 18?)  It's also happened on every version of audiobookshelf since I've started using it.

indiefan:
…ile having no played track ref.

## Brief summary

#1385 contains repro steps for disassociating the native player from the vue layer. While not technically the trigger users are experiencing in the wild, it's a good method for triggering a similar state, and one that we should be resilient towards. After using it to harden the Native/Vue sync behavior, I will send a patch to address that specific trigger.

This change recognizes when the native audio player is sending updates while the the Vue layer does not believe audio is being played. Regardless of how we got into that state, I believe we should pause and close playback to avoid losing progress. Fwiw, if this state is expected under normal conditions (and recoverable) then we shouldn't land my PR.

## Which issue is fixed?

Fixes #1385 (progress will not be lost any more).

Pressing "play" from native/bluetooth controls in this state will still result in an error toast (because the native and Vue players have become disassociated). Looking into a follow up to restore/rehydrate the Vue layer from a playing native audio player, but I think this fix should go in first as it at least avoids the outcome where progress is lost.

## Pull Request Type

It affects both I believe because it changes the Vue behavior. I have tested this on iOS, but in theory the change makes sense for both platforms (we probably don't want native playback to continue when the Vue layer believes it to have stopped).

## In-depth Description

Not much more to say than what's above.

## How have you tested this?

Tested on iOS using the repro steps provided on #1385 to trigger disassociation of Native/Vue layers and verify remediation.

## Screenshots

N/A.

### DIFF
--- a/components/app/AudioPlayer.vue
+++ b/components/app/AudioPlayer.vue
@@ -551,9 +551,7 @@ export default {
     timeupdate() {
       if (!this.$refs.playedTrack) {
         console.error('Invalid no played track ref')
-        this.pause()
         this.closePlayback()
-        // TODO: Attempt to rehydrate Vue session without interrupting native playback.
         return
       }
       this.$emit('updateTime', this.currentTime)


components/app/AudioPlayer.vue
```diff
@@ -551,6 +551,9 @@ export default {
     timeupdate() {
       if (!this.$refs.playedTrack) {
         console.error('Invalid no played track ref')
+        this.pause()
+        this.closePlayback()
+        // TODO: Attempt to rehydrate Vue session without interrupting native playback.
```
advplyr:
```suggestion
        this.closePlayback()
```

Pause is unnecessary here and I don't think we should rehydrate here. As I mentioned in https://github.com/advplyr/audiobookshelf-app/issues/1385#issuecomment-3568360052 the only time this should happen is when webview is first mounted (with the exception of the disconnect bug). I believe the major issue is devices killing webview in the background, so when the user opens the app again it mounts webview again. Then we can implement the attempt to sync Vue/Native.

advplyr:
I think closePlayback can be a good solution for now to at least resolve the out of sync issue when the user opens the app again.

When a solution is implement to sync Vue/Native on app mount it could be a potential race condition though. That can be avoided by handling the sync check in AudioPlayer.vue `mounted` and delaying attaching the listeners until after the sync check. 
In my other comment I mentioned adding the sync check in the default layout `mounted`, but it probably makes more sense in AudioPlayer given that the listeners for native events are there.

advplyr:
Thanks for digging into this!

advplyr:
Actually this isn't going to work. The race condition I mentioned is already present. You can reproduce it by playing different media while the audio player is open. If an event comes in at the same time that the switch is happening the component refs are not accessible.

I'm going to roll this back before release unfortunately. I found another potential issue with token refresh on ios. I use a 10 second access token expiration on my dev server and occasionally it was kicking me to the server connect screen.

PR #1728, opened 2025-11-16
Bump Realm dependency in Podfile for iOS.

indiefan:
## Brief summary

Bumps the version for RealmSwift in the iOS Podfile.

So I needed this to get past Xcode errors (something about RealmSwift requiring iOS 26.1). Not sure if it's just my setup, or if others have experienced the same issue. Discord had one member asking about the same error a year ago, but getting no response. Sending it here in case it helps others, but feel free to ignore if fresh installs work for other folks.

I don't know what changed in RealmSwift between 10.54.4 and 10.54.6 (my guess is that 10.54.6 is either the oldest version available, or that my local setup required an update to happen to unbork the framework linking from not updating for a long time).

### DIFF
--- a/ios/App/Podfile
+++ b/ios/App/Podfile
@@ -29,7 +29,7 @@ target 'Audiobookshelf' do
   capacitor_pods
   # Add your Pods here
   
-  pod 'RealmSwift', '~> 10'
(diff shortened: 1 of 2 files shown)


mnkyjs:
Works for me, thanks for your contributions 🚀

PR #1730, opened 2025-11-18
Android: fix Cast volume to control receiver

### LINKED ISSUE #584: Volume control from device when casting not working
### Steps to reproduce

1. Open audiobookshelf app
2. Play a book
3. Cast book
4. Try changing the volume on your phone. This part doesn't seem to work for me (Android 12, OnePlus 7) casting to Google Nest Display.

### Expected behaviour
- when casting, your phone essentially becomes a 'remote control', so I would expect to be able to change the volume. I'm not sure if this is a bug or feature request but I'm assuming a big, since it works with every other app.

### Actual behaviour
- The volume does not change.

### Environment data
- Local server install (not docker) V2.2.15

Audiobookshelf Version:

- [0.9.61 beta ] Android App?
- [ ] iOS App?

#### Android Issue
Android version: 12

Device model: OnePlus 7

Stock or customized system: stock

#### iOS Issue
iOS Version:

iPhone model:

eagleeyetom:
- Route hardware volume keys to Cast via MediaSession VolumeProvider
- Use CastSession.volume (0–1) mapped to 0–100; clamp and step by 1
- Update provider with actual device volume after each change (prevents stalls)
- Drop stream-volume/polling attempts and other leftovers
- Expose get/set/adjust device volume commands to the player

## Brief summary

Currently when you try to change the volume while casting, it changes the volume of the phone/tablet. It's very inconvenient as you need to find a remote to change the volume of the receiver.

## Which issue is fixed?

Fixes #584 

## Pull Request Type

Android/backend

## In-depth Description

Not sure if it's the best solution, but it changes the behavior of the app to the desired one. It tries to change the volume by 1 and if it's not supported, increases the treshold.

## How have you tested this?

Install the APK, cast -> try to change the volume.

## Screenshots

N/A

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/player/CastPlayer.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/player/CastPlayer.kt
@@ -76,7 +76,10 @@ class CastPlayer(var castContext: CastContext) : BasePlayer() {
       COMMAND_GET_MEDIA_ITEMS_METADATA,
       COMMAND_SET_MEDIA_ITEMS_METADATA,
       COMMAND_CHANGE_MEDIA_ITEMS,
-      COMMAND_GET_TRACKS)
+      COMMAND_GET_TRACKS,
+      COMMAND_GET_DEVICE_VOLUME,
+      COMMAND_SET_DEVICE_VOLUME,
+      COMMAND_ADJUST_DEVICE_VOLUME)
     .build()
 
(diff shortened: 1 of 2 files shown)


advplyr:
This is working but really buggy for me. Changing volume on the cast player doesn't reflect back on the phone. When changing volume quickly on the phone, the phone will sometimes be showing a different volume then what is being shown on the cast player.
It is still an improvement so I'm going to include this in the next release. Thanks!

PR #1738, opened 2025-11-26
Android cast: stabilize MediaSession volume handling; simplify Google…

eagleeyetom:
… Cast volume

## Brief summary

Improves the Google Cast functionality to control the volume via a casting device

## Which issue is fixed?

Google Cast should now work properly, increment by 1 step and reflect the correct volume.

## Pull Request Type

Android

## In-depth Description

Use immediate UI updates via VolumeProvider (setCurrentVolume) and dispatch device volume changes without relying on async callbacks.

Simplify Cast volume: remove even-value quantization and step-2 enforcement; add short pending cache (~100ms) to bridge async setVolume.

Align increase/decrease to ±1 to match platform step; clamp to 0..100.

## How have you tested this?

APK install -> Cast -> Change volume

## Screenshots

N/A


advplyr:
Works great, thanks!

PR #1739, opened 2025-11-27
Fix clicking next chapter button

Cutch:
## Brief summary
The next chapter button has a bug, repeated clicks to next chapter does not work if chapters have have decimal places.

## Pull Request Type

Test on android

## In-depth Description

The next chapter button has a bug. It does not allow repeated clicks because seek, seeks to the floor(time) but the chapter identification uses .toFixed(2). Next chapter only works if the time is greater than the decimal value of the chapter.
This seems to be the only option as AbsAudioPlayer.seek does not seem to accept decimal values.

## How have you tested this?
Android studio virtual phone


advplyr:
Thanks, this was fixed more complete in https://github.com/advplyr/audiobookshelf-app/pull/1848

PR #1747, opened 2025-12-09
Introduce Media3 Playback Architecture for Android App Modernization

hypnotoad08:
## Brief summary

This PR adds Media3 as a second Android playback backend alongside the existing ExoPlayer2 implementation. The two backends are separate build flavors (`exov2` and `media3`) selected by `BuildConfig.USE_MEDIA3`. The exov2 flavor stays the production player; the media3 flavor installs alongside it (`.media3` application id suffix) so both can be tested on the same device.

## Which issue is fixed?

ExoPlayer2 is in maintenance mode and no longer receives feature work; Media3 is its successor. This PR sets up the migration path.

## Pull Request Type

**Platform:** Android only
**Component:** Audio playback backend
**Type:** Feature (new build flavor)

### How the history is organized

1. **Build flavors, Media3 dependencies, dual-APK CI** — adds the `player` flavor dimension (`exov2`/`media3`), the Media3 libraries, and CI that builds and publishes both APKs.
2. **Data model extensions** — Media3 `MediaItem` support on the library data classes used for browsing.
3. **Shared playback infrastructure** — code both players use: sleep timer coordination, widget snapshotting, progress sync scaffolding, playback constants.
4. **The Media3 stack behind a `PlayerBackend` abstraction** — `Media3PlaybackService` (a `MediaLibraryService`), Cast via Media3 `CastPlayer`, Media3 notifications, an Android Auto browse tree, and a progress syncer that follows the exov2 sync cadence. `AbsAudioPlayer` picks the backend once at load based on the flavor.
5. **Entry-point wiring** — activity, plugins, widget, and cast entry points route to the active flavor.

6. **Fixes.** The rest of the commits are bugs I hit using the media3 flavor on my own phone, grouped by area:

   *Position and seeking.* media3 queues one `MediaItem` per audio track, so `player.currentPosition` is track-relative. Several seek and metadata paths used it where the book-absolute position was needed. Correct on single-track books, wrong on multi-track ones: a jump near a file edge was clamped at the boundary instead of crossing into the next track.

   *Progress sync.* `lastSyncTime` advanced before the payload was delivered, so most ticks credited about a second and long sessions recorded almost no listening time. It now advances after a local save or a successful server sync, matching `MediaProgressSyncer`.

   *Crashes.* A truncated session file made every launch throw. A session with no audio tracks hit `coerceIn(0, -1)`. `SleepTimerCoordinator` threw when the adapter had no session instead of returning null.

   *Android Auto.* Media ids missing from the memory cache failed to resolve on cold start, `page`/`pageSize` were ignored when serving browse children, and the podcast auto-advance payload was built off the main thread.

   *Cast.* A missing `MediaTransferReceiver` declaration left the notification claiming local playback; adding it to the manifest fixes this for both flavors. A closed session was re-emitted as a phantom book. A transient `IDLE` during transfer was published as real playback state. Handoff had a 14s silent gap because the queue was never pushed to the receiver. Chapter-metadata updates had to stop touching the queue while casting, since `CastPlayer` implements `replaceMediaItem` by reloading the receiver's current item.

   *UI attachment.* When playback starts from Android Auto the app process already hosts the session, but the UI controller never adopted it. It now adopts the service-published session on connect and on timeline changes.

   Some commits also delete dead scaffolding (unused `CastManager` cast-session callbacks, an `onSeekCompleted` chain that dispatched nothing) and cut redundant work: per-second metadata serialization when the payload is unchanged, a per-item disk read on the Downloads browse node, and serialized Android Auto loads that now run concurrently.

7. **Structural cleanup of the new Media3 code.** Ten commits pulling collaborators out of `Media3PlaybackService`: separate queue, error, lifecycle, progress-sync, widget, notification-metadata and command-policy types, and a `PlaybackPositionModel` that owns the track-relative to book-absolute conversion. The 31-member service-host interface becomes three per-consumer interfaces. Only `player/media3/` is touched, and no behavior changes.

Three commits restore master's indentation in files the branch touches. They are large and contain no logic changes.

Commits have been squashed where it made sense, so the history reads as discrete changes rather than the order things were actually written. Each one compiles both flavors.

### Key architectural difference between the flavors

- **exov2:** MainActivity binds to `PlayerNotificationService` on startup (traditional foreground service model); the plugin waits for a service-ready callback.
- **media3:** no service binding in MainActivity. `MediaLibraryService` is system-managed per Media3's design; connecting a `MediaController` creates the service, and Media3 promotes it to foreground when playback starts. The plugin initializes its `PlaybackController` when needed.

The Capacitor plugin interface is unchanged, so no web or iOS changes are required.

### Effect on the exov2 flavor

exov2 stays the default and keeps the production application id, so releases keep updating existing installs with the same player. The media3 flavor installs under its own app id and only ships if it is promoted later.

Logic both backends need was extracted into shared code rather than duplicated: sleep timer handling behind `SleepTimerCoordinator`, network state in `NetworkMonitor`, a widget snapshot instead of handing the widget the service itself, the player state the progress syncers read behind `PlaybackStateHost`, and `PlaybackMetricsRecorder` so both stacks log the same rollout metrics. `PlayerNotificationService` calls into those instead of its own copies. Behavior is unchanged, and no new user-facing features or settings are added for either flavor.

### Behavior parity

The media3 flavor reproduces exov2 behavior: pause auto-rewind with chapter clamping, sleep timer (end-of-chapter, fade-out, shake-to-reset, chime), Android Auto browsing including alphabetical drawdown and series ordering, podcast auto-advance on Auto, Cast handoff, media widget, listening-session sync cadence, and local/downloaded playback.

### Media3 API stability

Media3 session APIs are still marked `@UnstableApi`, so the new code carries the required `@OptIn(UnstableApi::class)` annotations.

### Dependency and build changes

Media3 library version 1.11.0 (Apache 2.0) was added to both flavors since shared code references Media3 types:

- androidx.media3:media3-exoplayer
- androidx.media3:media3-session
- androidx.media3:media3-cast
- androidx.media3:media3-exoplayer-hls

Other build changes: `kotlinx-coroutines-guava` added (Media3's ListenableFuture APIs), `compileSdkVersion` 35 → 36, Android Gradle Plugin 8.8.2 → 8.9.3 (required for compileSdk 36), `kotlin_coroutines_version` 1.10.1 → 1.10.2, `kotlin_version` → 2.2.10.

### Build configuration

```bash
# Build ExoPlayer2 (default)
./gradlew assembleExov2Debug

# Build Media3
./gradlew assembleMedia3Debug

# Install both simultaneously for comparison
./gradlew installExov2Debug installMedia3Debug
```

CI builds both APKs (`audiobookshelf-exov2-*.apk`, `audiobookshelf-media3-*.apk`) and the test page links both.

## How have you tested this?

I've been using the media3 flavor as my daily driver for the past several months to catch real-world edge cases. Both flavors compile and install simultaneously, with no crashes or ANRs in daily use.

Specific testing on hardware (Pixel; Android Auto on both the Desktop Head Unit and a real car head unit; Cast; Wear; BT):

1. **Local playback:** start/pause/resume, seeking, playback speed changes, chapter navigation, downloaded items.
2. **Foreground service:** app backgrounded before playback is ready (Android 12+) — service starts without `ForegroundServiceDidNotStartInTimeException`.
3. **Pause/resume auto-rewind:** single rewind, clamped to chapter start.
4. **Sleep timer:** fixed length (volume fade + pause at zero), end-of-chapter mode, shake-to-reset, notification updates.
5. **Android Auto (DHU and in-car):** library browsing (continue listening, books, podcasts, series), search, playback controls, finished multi-track book restarts from track 1, podcast episode completion advances to the next episode without error.
6. **Casting:** device discovery, transfer to cast device, starting playback while already connected to a cast device, position stability across progress saves and chapter boundaries, play/pause/seek from both ends, cast-device volume from the phone rocker, resume on disconnect.
7. **BT/car buttons:** next/previous on single-file books maps to seek when there is no adjacent track.
8. **Wear:** seek buttons produce a single jump per press.
9. **Progress syncing:** sync across devices, local progress updates, server sync cadence matches exov2 (verified against server sessions), token refresh mid-stream.
10. **Notifications:** Android 13+ permission flow, playback controls from the notification, updates during playback.

The same list was re-run after the structural cleanup.


nichwall:
Is this PR still stable? I've been getting emails almost every day about updates, so it doesn't appear to be stable anymore.

Edit to clarify, Every day is a bit of an exaggeration, just nearly every time I have sat down to look at GitHub notifications there is a new one for this PR.

hypnotoad08:
> Is this PR still stable? I've been getting emails almost every day about updates, so it doesn't appear to be stable anymore.
> 
> Edit to clarify, Every day is a bit of an exaggeration, just nearly every time I have sat down to look at GitHub notifications there is a new one for this PR.

@nichwall 

Apologies for the churn. The frequent updates were partly from keeping the branch in sync with master, but also from continuing to refine the Media3 conversion after reopening the PR.

As I worked through it, I found several areas where the original structure could be improved, so I spent a fair amount of time refactoring to better separate responsibilities, improve maintainability, and simplify the overall design. 

I also addressed a number of bugs and follow-up issues discovered along the way, including some specific to Android Auto.

At this point, I'm comfortable with the current state of the PR and don't anticipate further changes unless review or testing uncovers another bug.

PR #1748, opened 2025-12-09
Fixes unaccurate syncing issues and potential fix for extremly large sessions

Vito0912:
## Brief summary

Uses a `Double` instead of a `Long`, because the app used to send `14.9xx`, which always got casted to `14`.
Now it uses precision (as it does for iOS). For this also the syncLastTime has been moved before the acutal request, because the server response time was always deducted from the next listening sync.

Also added a logic, to prevent syncs that are to big. If a sync exceeds 180 seconds (which is at max 3 times syncing), then it resets the listened time back to 0. There is no chance that 180 seconds should be accomplished. This isn't really a fix for larger sessions, but should help a lot not having them at all. At max 3 minutes are lost from listening

## Which issue is fixed?

Discord, potential a few more

## Pull Request Type

Android, backend

## In-depth Description

See Discord conversation. Casting always "removes" a second of listened time

## How have you tested this?

Mobile app

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/media/MediaProgressSyncer.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/media/MediaProgressSyncer.kt
@@ -296,13 +296,6 @@ class MediaProgressSyncer(
         }
       }
     } else if (hasNetworkConnection && shouldSyncServer) {
-      // Do not sync when the sync should not be possible. 180 sec is thrice the max sync amount
-      if (diffSinceLastSync >= 180000L) {
-        lastSyncTime = System.currentTimeMillis()
-        AbsLogger.info("MediaProgressSyncer", "sync: Not sending progress to server. timeListened too big $diffSinceLastSync (title: \"$currentDisplayTitle\") (currentTime: $currentTime)")
-        return cb(null)
-      }
-
       AbsLogger.info("MediaProgressSyncer", "sync: Sending progress sync to server (title: \"$currentDisplayTitle\") (currentTime: $currentTime) (session id: ${currentSessionId}) (${DeviceManager.serverConnectionConfigName})")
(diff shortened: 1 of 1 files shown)


nichwall:
What happens if someone is listening offline for more than 3 minutes and tries to sync? I haven't gone through the code yet, just based off of the description.

Vito0912:
That's actually a very good point. I didn't thought about that. I will need to check the code tomorrow.
I assumed, because I did that in my app, that the syncing part is separate from the local syncing, but now where you say it, it might not be. But on the other hand there are DB calls. Anyway, will check and maybe fix then

hypnotoad08:
this will break app startup,  would need to convert the device table as well as the playbacksession table from longs to doubles before trying to access the data in the dbmanager.

Vito0912:
I sadly do not have any sessions saved. @hypnotoad08 could you try again now. I tried to use a Number, which in theory should work, hopefully?
I don't do any dev in Kotlin at all normally, so this could also be completely wrong.

@nichwall I removed the cap again, because I tried a bit and there is no part, where it couldn't be infinity in theory. This also makes it very easy for sync bugs to happen, as the listened time is just kept in one variable and not really validated at all.

So from my understanding, if the app gets paused by the OS (e.g. when the audio stops playing), but the sync was not yet performed (due to whatever reason), then `lastSyncTime` would be still the old value. Now the user openes the app again, the app unpauses and tries to sync the progress again. Now the progress is `today - lastSyncTime` adding x hours to the session.
This would also make sense, why it is happening to users after a night e.g., because they have their phone locked and laying. If the audio ends and the OS pauses the app, before the sync happens aboves scenario seems most likley to me. What do you think?

Vito0912:
If this is the case, this PR should also reduce that, because the time is set before the http request, making it more likely that the variable gets updated. But this is only, if my theory is correct

Issue #1130, opened 2024-03-10
[Android] Chromecast doesnt continue scanning connections


patienttruth:
### Steps to reproduce

1. Open podcast or audiobooks
2. Play file
3.

### Expected behaviour
- based on screenshots and feature list, I'd imagine I'd expect to see a Chromecast icon.

### Actual behaviour
- no cast icon is present

### Environment data

Audiobookshelf Version: 2.8.0

- [ X] Android App?
- [ ] iOS App?

#### Android Issue
Android version: 14

Device model: Pixel 6

Stock or customized system: GrapheneOS

# Additional Information:
Initially I was on 2.7.2, however I updated to 2.8.0 while preparing this report. Upon initial boot I restarted the Android App, and I saw the cast icon, and was able to select a cast target. It showed as connected, however it would not start playing.

Eventually the app stopped and when I opened it back up the cast icon no longer shows up.

advplyr:
The cast icon only shows up when it sees a device that can be casted to. Make sure your device can see a chromecast device.

patienttruth:
> The cast icon only shows up when it sees a device that can be casted to. Make sure your device can see a chromecast device.

My other apps (Antennapod, etc) can see and connect / play to the chromecast devices. If it's normally a functional feature I'll dive into my networking as there have been some other bugs that somehow aren't affecting the other apps.

patienttruth:
I'm not sure if this provides insight, but I found an error in my Home Assistant Cast Log. At fist I thought "why would this be showing in a HA log, but I'm starting to thing that it's aware of the media sources that cast, and therefore is able to log when the Chromecast has a related error.

```
Failed to cast media https://bookshelf.lab.X.com/api/items/X/file/X?token=X. Please make sure the URL is: Reachable from the cast device and either a publicly resolvable hostname or an IP address
```

Is it necessary for audio bookshelf to be publicly accessible? If so I apologize as I must have missed that.

advplyr:
I'm not sure about that, I think it depends on the device. Have you tested it?

advplyr:
Were you able to resolve this issue?

sydlexius:
I can reproduce this issue on my S22 Ultra.  I can appreciate that if my device let its wifi radio sleep that it might lose contact with nearby devices, but my expectation would be that if I switch between Audiobookshelf accounts that it should make another sweep/scan of my environment for other devices.  As it stands, I basically have to force stop the app and relaunch it to do this.

[...]

Issue #817, opened 2023-07-29
[Android App] No cast button


shlomitan:
### Steps to reproduce

1. Go into android player
2. Play a book (or don't, doesn't matter) 
3. No cast button anywhere

### Expected behaviour
- Cast button to show up so I can select device to cast. When I went to the web player the button was there.
- Checked on my wife's phone, the button is there! 

### Actual behaviour
- I can't find the button :(

### Environment data

Audiobookshelf Version: server 2.2.23, app 0.9.65

- [*] Android App?
- [ ] iOS App?

#### Android Issue
Android version: 12

Device model: Samsung s10e

Stock or customized system: stock

#### iOS Issue
iOS Version:

iPhone model:

advplyr:
Maybe this is something with the network your device is connected to? Abs is using the google cast library that is detecting castable devices. If one is found then the button is shown. This seems like something with your device and not a bug within Abs so try experimenting with settings.

Have you installed the app from the google play store?

shlomitan:
I have no idea why, but restarting after reporting here fixed. Didn't help before.
Thanks, and I hope to never complain again!
(BTW - I love abs. You're amazing for making this happen)

LoudSoftware:
Am currently running into this issue, the only way to resolve this is to close the app and force stop it from settings and restart it.

Perhaps it might be better if the cast button always be present (or at least an option to do so in the settings)?

Otherwise, if the current behaviour is intended, an option to force refresh/scan for cast devices (maybe pull down in library to refresh?) could work as well.

In any case, I have a feeling the app doesn't always scan or remember previous scan results (even though the network didn't change)

PR #1762, opened 2025-12-23
Always show cast button on Android

Landynu:
## Brief summary

Very minor code change to always show the cast button instead of conditionally showing it based on 'isCastAvailable'.  This ensures Android users can always cast their audio, despite state changes.

## Which issue is fixed?

Fixes Issue #1130 

Although marked as closed, this could have been the underlying issue with #817 

## Pull Request Type

Platform: Android
Frontend Component - Appbar.vue

## In-depth Description

Previously, the cast button would disappear (or not show at all) if booting the app while off the network, or when changing state such as using Bluetooth. See comments in #1130 .  

This change is no longer conditional on 'isCastAvailable', and now always displays while on Android, and will use the Google Cast SDK to run scans.  Device discovery now happens on demand rather than upon app startup, ensuring the user doesn't have to reboot the app in order to continue their listening session on a castable device.

From what I can tell, iOS uses AirPlay, and this functionality only affected Android devices. As a result, if we wanted to dive further, we could likely retire all of the logic around the 'isCastAvailable' and instead rely entirely on the Cast SDK.

This is a change that solves a problem for all Android users using Google Cast.  This is the best solution because it is now triggered by the user at their discretion.

## How have you tested this?

I compiled the code change in Android Studio and sideloaded onto my Google Pixel 6.  Given the minor change, I tested the sideloaded app alongside the current release and was able to successfully connect to castable devices without the need for rebooting with these code changes.  In the current release, the cast button would not show when connecting to a new network or when previously on Bluetooth.  With the changes, the cast button is always visible and can be polled whenever the user chooses.

## Screenshots

No real change to the UI, as now the cast button is always visible.

Note: This is my first open source contribution, so apologies if I missed any steps.

### DIFF
--- a/components/app/Appbar.vue
+++ b/components/app/Appbar.vue
@@ -21,7 +21,7 @@
       <widgets-download-progress-indicator />
 
       <!-- Must be connected to a server to cast, only supports media items on server -->
-      <button type="button" aria-label="Cast" v-show="isCastAvailable && user" class="mx-2 cursor-pointer flex items-center" @click="castClick">
+      <button type="button" aria-label="Cast" v-show="$platform === 'android' && user" class="mx-2 cursor-pointer flex items-center" @click="castClick">
         <span class="material-symbols text-2xl leading-none">
           {{ isCasting ? 'cast_connected' : 'cast' }}
         </span>


sluggathor:
can we merge this? I would like to continue using the app but the iffy Chromecast situation keeps pushing me towards using other apps that have other drawbacks.

nichwall:
Superseded by https://github.com/advplyr/audiobookshelf-app/pull/2009

Issue #1308, opened 2024-09-02
[Bug]: Can't fastforward .aac file on Android


olek-osikowicz:
### What was the Problem?

As you can see on the screen recording, I tried to fastforward the book to the middle so I could resume when I finished.

https://github.com/user-attachments/assets/0333e14f-0927-4676-b5c6-f9722549bf92

Moving the slider doesn't work (backs to the beggining) neither does using the button to skip 10 seconds.

As you can hear the file is healthy, and it starts playing the book from the beggining.

I suspect it's to blame the file format.

<details><summary>File details</summary>
<p>
Here is the output of `ffprobe -show_format`

Input #0, aac, from './Niezwyciężony.aac':
  Duration: 06:57:39.80, bitrate: 133 kb/s
  Stream #0:0: Audio: aac (LC), 44100 Hz, stereo, fltp, 133 kb/s
[FORMAT]
filename=./Niezwyciężony.aac
nb_streams=1
nb_programs=0
format_name=aac
format_long_name=raw ADTS AAC (Advanced Audio Coding)
start_time=N/A
duration=25059.801256
size=417424242
bit_rate=133257
probe_score=51
[/FORMAT]

</p>
</details> 

### Steps to Reproduce the Issue

1. Add `.aac` file to your library.
2. Play it on android app.
3. Try to fastforward using the sliders or skip ahead buttons.

### What was Expected?

File should be played from requested timestamp.

### Phone Model

Samsung Galaxy S10

### Phone OS

Android 12

### Audiobookshelf App Version

Android App - 0.9.76

### Installation Source

Google Play Store

### Additional Notes

I solved the issue by converting the file to `mp3` format: 

```bash
$ ffmpeg -i ./input.aac -acodec libmp3lame ./output.mp3
```

Then it works without an issue.

nichwall:
A file being able to play does not mean that the file supports seeking. Can you try converting the file from aac to aac using the `-c copy` parameter?

https://www.audiobookshelf.org/guides/ffprobe/#fixing-an-audio-file

PanZWarzywniaka:
I am sorry, I missed that section in the docs, however now I tried to `ffmpeg -i input.aac -c copy output.aac`, but it's still not seekable. Same problem with skipping.


Created separate book folder and copied the file to avoid caching.

PanZWarzywniaka:
I forgot to add that it works perfectly fine on Desktop (web client).

AngelouDi:
[...]

PR #1763, opened 2025-12-24
Enable constant bitrate seeking to fix AAC seek reset in ExoPlayer

randomdude583:
## Brief summary

Fixes an Android issue where seeking in certain AAC audio files would reset playback to the beginning of the file. This could also occur when resuming playback partway through a file after the app had been inactive, which effectively behaved the same as a seek.

## Which issue is fixed?

Fixes [#1308](https://github.com/advplyr/audiobookshelf-app/issues/1308)

## Pull Request Type

Android app

## In-depth Description

On Android, audiobookshelf uses ExoPlayer for audio playback. For some AAC audio files, ExoPlayer is unable to derive a reliable seek map. When a seek is requested in these cases, ExoPlayer may “adjust” the seek back to position 0, causing playback to jump to the beginning after buffering.

This behavior has been observed with certain AAC files and does not reproduce in the web player, which uses the browser’s media stack. Other formats (such as MP3) generally do not exhibit this issue.

This change enables ExoPlayer’s constant-bitrate seeking mode on the DefaultExtractorsFactory for progressive playback (both local and direct HTTP sources). This allows ExoPlayer to perform approximate seeking for media where a seek map is unavailable, preventing the seek-to-start behavior while preserving normal behavior for formats that provide proper indexing.

Notes:
- Seeking becomes approximate for affected AAC files, which is expected for this fallback.
- No regressions were observed with MP3 or other indexed formats.
- HLS playback is unaffected.

## How have you tested this?

I tested this using the TrueNAS Tech Talk podcast (RSS feed: https://www.truenas.com/feed/podcast/truenas-welcome-to-the-open-storage-era/), which publishes episodes encoded as AAC and reliably reproduces the issue.

Testing included:
- Seeking using the playhead
- Using skip forward/back controls
- Pausing playback, exiting the app, and later resuming playback

These tests were performed both while streaming episodes from the server and when playing episodes downloaded locally.

Additional testing was done with audiobooks encoded as m4b and mp3, and no regressions were observed.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/player/PlayerNotificationService.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/player/PlayerNotificationService.kt
@@ -480,6 +480,8 @@ class PlayerNotificationService : MediaBrowserServiceCompat() {
         val dataSourceFactory = DefaultDataSource.Factory(ctx)
 
         val extractorsFactory = DefaultExtractorsFactory()
+        extractorsFactory.setConstantBitrateSeekingEnabled(true)
+
         if (DeviceManager.deviceData.deviceSettings?.enableMp3IndexSeeking == true) {
           // @see
           // https://exoplayer.dev/troubleshooting.html#why-is-seeking-inaccurate-in-some-mp3-files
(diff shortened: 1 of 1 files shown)


nichwall (approved):
I was able to confirm this does fix seeking with the AAC files of The TrueNAS Show as described in the PR description, and they did not work before. I did also double check that enabling MP3 Index Seeking would not fix it and confirmed that MP3 index seeking does not fix this issue (not that it should, just as a sanity check). This will eventually be superseded by a migration to media3, but is a simple fix for now.

Looks good to me!

advplyr:
Thanks!

PR #1767, opened 2025-12-29
Expand time interval options for jump controls

### LINKED ISSUE #1445: [Enhancement]: Increase Skip forward/back time options in Mobile App (to align with web app)

increased options for default skip forward/back times in the web player were add with [#1995 ](https://github.com/advplyr/audiobookshelf/issues/1995) but the mobile still only has 5/10/30 second options. Would be great to have the additional skip times that were added to web app available in the mobile apps as well. 

Thanks! 

additional skip times are helpful to more easily navigate through a podcast,skip past ads, etc. 

Image

iOS App - 0.9.78

_No response_

FreedomBen:
## Brief summary

- Added additional options for customizing jump forward/backward.  Instead of just 5, 10, and 30 seconds, we now have 5s, 10s, 15s, 30s, 60s, 2m, and 5m.
- Switched the jump interval to a list in the settings menu for better UX
- Reordered "Disable auto rewind time" for aesthetics in settings menu (it looked out of place after backwards time and forwards time were turned to dropdowns

## Which issue is fixed?

Fixes #1445

## Pull Request Type

Android and iOS.  No backend.

## In-depth Description

- Added additional options for customizing jump forward/backward.  Instead of just 5, 10, and 30 seconds, we now have 5s, 10s, 15s, 30s, 60s, 2m, and 5m.
- Switched the jump interval to a list in the settings menu for better UX
- Reordered "Disable auto rewind time" for aesthetics in settings menu (it looked out of place after backwards time and forwards time were turned to dropdowns

The only downside of this (afaict), and what created the biggest challenge, is the lack of numeric glyphs in the Material Symbols font.  They only have 5, 10, and 30, which we were previously using.  Given that they don't have others, that means we had to change strategies for displaying.  I switched all to generic backward and forward icons and added the time interval as a label next to it so the user can still tell at a glance what the setting is.  For consistency we don't use the numbered glyphs anymore.  For more info on the Material Symbols font, see: https://fonts.google.com/icons?icon.query=forward

However this does feel like two steps forward, one step back, though I can't think of good options.  Other possibility would be to create SVGs, though I'm not skilled enough to do that.  Personally this feels like a good compromise, but others will probably differ. I considered using the numbered glyphs when the selected value was 5, 10, or 30, but when testing I found the inconsistency jarring and kind of janky.

## How have you tested this?

Yes, tested on Pixel 8 Pro against personal prod server running v2.32.1.  Not tested on iOS (I don't have any Apple devices)

## Screenshots

See comments below

### DIFF
--- a/strings/en-us.json
+++ b/strings/en-us.json
@@ -360,6 +360,6 @@
   "ToastRSSFeedCloseFailed": "Failed to close RSS feed",
   "ToastRSSFeedCloseSuccess": "RSS feed closed",
   "ToastStreamingNotAllowedOnCellular": "Streaming is not allowed on cellular data",
-  "UnitSecondsShort": "{0}s",
-  "UnitMinutesShort": "{0}m"
+  "UnitMinutesShort": "{0}m",
+  "UnitSecondsShort": "{0}s"
 }


FreedomBen:
# Screenshots:

## Selecting a new jump value:

Screenshot_20251229-130045

## The new settings page:

Screenshot_20251229-130040

## Full page player:

Screenshot_20251229-130028

## Mini-player (see bottom):

Screenshot_20251229-130018

nichwall:
Could we instead put the numbers below the skip buttons? I think it looks weird having the number next to the icons.

We will also want to make these strings translatable, similar to how this was attempted in https://github.com/advplyr/audiobookshelf-app/pull/1526

Edit to add: can you also edit the PR body to follow the template?

FreedomBen:
- Numbers moved below the skip button (does indeed look a lot better)
- Strings translatable (I found the i18n system a bit confusing so made a best effort, but appreciate some guidance here if this is not a good direction).  Commit with i18n stuff is https://github.com/advplyr/audiobookshelf-app/pull/1767/commits/642cb07b4d982fae2d5e9e143d932591b63c03be
- Edited PR

## New screenshots

Showing labels below

### Settings menu

Screenshot_20251229-152742

### Minimized player view

Screenshot_20251229-152749

nichwall:
> - Numbers moved below the skip button (does indeed look a lot better)

Thanks, yeah I think that's a lot nicer.

> - Strings translatable (I found the i18n system a bit confusing so made a best effort, but appreciate some guidance here if this is not a good direction).  Commit with i18n stuff is https://github.com/advplyr/audiobookshelf-app/pull/1767/commits/642cb07b4d982fae2d5e9e143d932591b63c03be

Thanks for taking a stab at that. Sorry, I didn't explain it very clearly (on my phone so I can help point you to a better spot in a few hours if the following doesn't make sense). There is a global function for the string translation which allows for number substitution so you don't need to recreate the same thing. An example from the other PR is below.

```
      <p class="pl-4">{{ $getString('LabelJumpBackwardsTime', [settings.jumpBackwardsTime]) }}</p>
```

So, you could keep the `label` in the lookup and then just do something like `$getString(settings.jumpBackwardAmount.label, [settings.jumpBackwardAmount.value])` and then set the label to be `UnitSecondsShort` or `UnitMinutesShort` accordingly. I did not check the actual naming of the variables for this example, so adjust accordingly.

Hopefully that makes sense.

FreedomBen:
Thank you, I think I've got the i18n at least closer :-)  

New commit:  https://github.com/advplyr/audiobookshelf-app/commit/f548163689150ba1db8eba2e7fb12fd8695990de

FreedomBen:
@nichwall don't mean to pester but don't want to slip through the cracks. Is this looking ok?

nichwall:
It looks good to me from glancing through it, but I'll be able to sit and down and go through and actually test all of these PRs this weekend.

FreedomBen:
Awesome, much appreciated!  If it's helpful, the PRs I've sent don't conflict so can be tested altogether.  I threw them on a branch that I'm running on my device currently you can use (again only if helpful):  https://github.com/FreedomBen/audiobookshelf-app/tree/all-changes

nichwall (approved):
This is working for me and everything looks good to me.

FreedomBen:
@nichwall anything else I need to do before merge?  (just making sure it's not waiting on me)

nichwall:
You're good, just need advplyr to look at it and merge it.

advplyr:
I made some updates:
- Reverted the player controls width change because we need to stay within the "safe" space I think they call it
- Removed the jump label when the player is collapsed. I didn't think it looked good collapsed. Too crowded. I think showing it when the audio player is expanded will work.
- Removed the icon in the dialog because the vertical alignment was bugging me and it's unnecessary anyway

Thanks!

PR #1768, opened 2025-12-30
feat(android): add custom headers to all HTTP request paths

### LINKED ISSUE #254: Support adding custom headers to server connection/requests

Zero Trust like app access, in particular Cloudflare Tunnels, are becoming quite popular in the self-hosted world.  They allow you to make your self-hosted applications available externally without opening ports.  However, they typically use a OpenID Connect or SAML style SSO approach to authenticate access which means an intermediate login page appears before connectivity to the application endpoint is then allowed.  As a workaround to client apps that cannot support this additional authentication step, Cloudflare supports the use of service tokens passed with requests as additional headers.
https://developers.cloudflare.com/cloudflare-one/identity/service-tokens/#connect-your-service-to-access

Support adding additional custom headers, custom keys and values, when setting up the server connection to allow us to set additional headers for requests made to the server.  A great app citizen that has done this quite well is LunaSea.
https://docs.lunasea.app/modules/sonarr#custom-headers
https://imgur.com/a/JSC1bYn

The only alternative would be to forward the ports directly from the internet, or to always have a VPN or similar running when remote.  Allowing custom headers to be set for the server connection would allow for very secure remote access.

### Additional context
N/A

yaker-gh:
## Brief summary

Completes the custom HTTP headers implementation for the Android app. The app already had UI scaffolding (`CustomHeadersModal.vue`) and a `customHeaders` field on `ServerConnectionConfig`, but headers were never actually sent with requests. This PR wires them through every HTTP path so the app works behind reverse proxies that require authentication headers (Cloudflare Access, Authelia, Authentik, etc.).

## Which issue is fixed?

Related to discussion: https://github.com/advplyr/audiobookshelf-app/discussions/1313
Closes: https://github.com/advplyr/audiobookshelf-app/issues/254

## Pull Request Type

- **Platform:** Android only
- **Scope:** Frontend (Vue/JS) and Backend (Kotlin)

### Commit 1: Core plumbing (`feat(android): support custom HTTP headers for reverse proxy auth`)

**Android (Kotlin):**

| File | Change | Why |
|------|--------|-----|
| `ApiHandler.kt` | Inject custom headers into `getRequest()`, `postRequest()`, `patchRequest()` | All API calls funnel through these methods — single injection point covers all endpoints |
| `DeviceManager.kt` | Add `customHeaders` getter property | Exposes headers from current `ServerConnectionConfig` to all Kotlin code |
| `AbsDatabase.kt` | Sync `customHeaders` when updating *existing* server configs | Previously only set on new configs — editing headers on an existing server had no effect |
| `DownloadItemPart.kt` | Add custom headers to Android `DownloadManager` requests | External-storage downloads use the system DownloadManager, not OkHttp |
| `InternalDownloadManager.kt` | Accept + apply custom headers to OkHttp download requests | Internal-storage downloads go through a separate OkHttp client |
| `DownloadItemManager.kt` | Pass `DeviceManager.customHeaders` to `InternalDownloadManager` | Wiring between header source and download pipeline |
| `PlaybackSession.kt` | try-catch around local cover bitmap decoding | Prevents crash when cover image decode fails (encountered during testing) |

**Frontend (Vue/JS):**

| File | Change | Why |
|------|--------|-----|
| `ServerConnectForm.vue` | Add "Custom Headers" button to UI; pass headers to ping, status check, and socket connect | Users need to configure headers AND they need to be used during initial connection |
| `plugins/nativeHttp.js` | Add custom headers to all `CapacitorHttp` requests | JS-layer native HTTP calls were missing headers entirely |

### Commit 2: Fix remaining HTTP paths (`fix: add CF Access custom headers to all HTTP paths`)

During real-world testing behind Cloudflare Access, several HTTP paths were found to be missing custom headers, causing:
- **Constant logouts** — token refresh requests were blocked by CF Access, so the app thought the token was invalid
- **Broken ebook/PDF rendering** — pdfjs fetches content through its own HTTP layer without custom headers
- **Broken real-time sync** — Socket.IO WebSocket connections were blocked by CF Access

| File | Change | Why |
|------|--------|-----|
| `plugins/axios.js` | Request interceptor injects custom headers into all axios requests | Axios is used alongside nativeHttp for many API calls — was completely missing headers |
| `plugins/nativeHttp.js` | Add headers to `refreshAccessToken()`; pass `serverConnectionConfig` through call chain | Token refresh hitting `/auth/refresh` without CF headers → blocked → forced logout |
| `store/user.js` | Add headers to Vuex `refreshToken` action | Duplicate token refresh code path (Vuex store action) — same logout bug |
| `components/readers/PdfReader.vue` | Add custom headers to pdfjs `httpHeaders` config | PDF/ebook viewer has its own HTTP layer that doesn't go through axios or nativeHttp |
| `plugins/server.js` | Accept `customHeaders` in `connect()`, pass as Socket.IO `extraHeaders` | WebSocket was connecting without auth headers — blocked by proxy |
| `layouts/default.vue` | Pass `customHeaders` when reconnecting socket on app resume | Auto-reconnect used old `connect()` signature without headers |

### Commit 3: UX fix (`fix: show edit form immediately without re-validating server connection`)

| File | Change | Why |
|------|--------|-----|
| `ServerConnectForm.vue` | `editServerConfig()` now shows the form directly instead of calling `submit(true)` first | Previously, editing a saved server config required the server to be reachable. If the server was down or the address changed, you couldn't edit the config. Now matches `newServerConfigClick()` behavior. |

### Commit 4: Ebook reader fix (`fix: enable CapacitorHttp to route ebook requests through native HTTP`)

The ebook readers (epub, comic, mobi, PDF) use axios/XHR for binary data fetches (`responseType: 'arraybuffer'`/`'blob'`). With `CapacitorHttp.enabled: false`, these XHR requests execute in the WebView and are subject to CORS enforcement. Behind reverse proxies like Cloudflare Access, the CORS preflight `OPTIONS` request is blocked because preflight requests cannot carry custom authentication headers per the CORS spec.

| File | Change | Why |
|------|--------|-----|
| `capacitor.config.json` | Set `CapacitorHttp.enabled: true` | Patches global `XMLHttpRequest`/`fetch` to route through native HTTP, bypassing CORS while preserving binary response types |
| `android/.../capacitor.config.json` | Same change in Android assets copy | Android build reads from this copy |

**Why this is safe now:** This was previously disabled because Capacitor 4's native HTTP broke binary responses (returning base64 strings instead of ArrayBuffer/Blob). Capacitor 5.7+ (PR [#6818](https://github.com/ionic-team/capacitor/pull/6818)) fixed this for GET requests by using a proxy URL approach — the native layer handles the HTTP request but the WebView's real XHR processes the response, preserving proper binary types. The app is on Capacitor 7.

### Known limitations

- **iOS:** No custom header support at all — `ServerConnectionConfig.swift` doesn't have a `customHeaders` property. Out of scope for this PR.
- **ExoPlayer HLS streaming:** Only sends `Authorization` header, not custom headers. Direct play uses token-in-URL so it works regardless.
- **Cover images via `` tags:** Browser limitation — `` elements can't carry custom headers. Works if server doesn't require auth for cover endpoints.

## How have you tested this?

- Configured Cloudflare Access with Service Token authentication (`CF-Access-Client-Id` + `CF-Access-Client-Secret`) in front of audiobookshelf server
- Added headers via the Custom Headers modal in the app
- Verified: server ping, login, library browsing, audiobook streaming, downloads (both internal and external storage), ebook/PDF reading, real-time sync via Socket.IO
- Verified token refresh works without triggering logout
- Verified headers persist across app restarts
- Verified headers update correctly when editing an existing server connection
- Verified ebook readers (epub, comic, mobi, PDF) load correctly behind CF Access after enabling CapacitorHttp
- Verified no regression in audio playback, navigation, or socket connections after enabling CapacitorHttp

## Screenshots
Screenshot_20251229-231947
Screenshot_20251229-231904
Screenshot_20251229-231856
Screenshot_20251229-231816


yaker-gh:
@advplyr, I can drop the nix setup if you don't want that in your codebase.

Let me know if there's anything else I can do to get this merged :)

3gfvgFWESAs44545:
Thank you for the great work. I'm hoping this gets merged soon! :)

MichaelKopt:
Please merge it. For those who use Audiobookshelf with Cloudflare, it is very useful.

Gathaeryx:
@advplyr any chance this could be merged in the next beta release? There only seems to be one release every two months, and the one from two weeks ago didn't include it. This the most discussed issue at the moment (after the iOS app), so it seems useful for plenty of people. If any additional testing has to be done to get this merged, I am willing to help.

nichwall:
Independent testing of PRs is helpful to indicate that the PR is ready to review. The current priority is on the web client rewrite and there are over 100 open PRs between the server and apps, so it may still be a bit.

If this APK works for you, you can also sideload the APK in the meantime.

MichaelKopt:
> If this APK works for you, you can also sideload the APK in the meantime

The issue is getting the APK. I don't have time to set up a build on my local machine, but if I have a beta APK with this feature, I can help with testing.

3gfvgFWESAs44545:
Happy to test as well if an apk can be produced by someone?

yaker-gh:
I can produce the APK. I've been running my fork ever since I added this change. I'll need a couple days then I'll post the link here.

yaker-gh:
There you go! https://gitlab.com/yaker/audiobookshelf-app/-/releases

Tested on a Pixel 9 running GrapheneOS and it works as expected.

nichwall:
I think you may have pushed something incorrectly, there are now over 2000 commits for this PR

yaker-gh:
> I think you may have pushed something incorrectly, there are now over 2000 commits for this PR

Yikes. Yes, I think so too.

nichwall:
If you can revert back to where it was we can approve workflows for the PR soon (TM). The workflows will automatically build APK for PRs already, it just doesn't happen automatically for new contributors.

yaker-gh:
Ooookay. Back to something reasonable.

yaker-gh:
I missed _a lot_ of stuff when first implementing this. It works, but it has a few quirks that I need to fix. I'll be updating this PR with quite a few changes in the coming days.

yaker-gh:
@nichwall, I think that's all of it!

3gfvgFWESAs44545:
Do we need testers for this yet? Can I test this feature?

Scyrizu:
I gave this a shot and got into my server while not on my home wifi,
however book images do not load, and when trying to stream a book i get "playback failed: source error" and the connection status is socket not connected.

yaker-gh:
@Scyrizu, can you post logs? I ran into those issues on my first attempt, but these were fixed on the latest commit. Perhaps I missed something, so I'd love to see logs so I can figure out where this is failing.

nichwall:
I spent a little time looking at this PR again today and I have a few questions/comments. I do not currently have a setup which requires custom headers, so would need to get something set up or have additional testing done.

> ### Known limitations
> 
> - **iOS:** No custom header support at all — `ServerConnectionConfig.swift` doesn't have a `customHeaders` property. Out of scope for this PR.

If iOS is not supported, the UI needs to be gated to not display on iOS users will assume this has been implemented and try to use it. There are examples in the settings and other parts of the UI of how this gating is done for settings not implemented on both platforms.

> - **ExoPlayer HLS streaming:** Only sends `Authorization` header, not custom headers. Direct play uses token-in-URL so it works regardless.

Streaming with custom headers seems like a common feature that should have custom headers like @Scyrizu mentioned? Tokens should not be included in the URL, this was an issue that was fixed in server version 2.22.0 (released May 2025), but even if it was included I'm not sure how that solves the HLS streaming.

> - **Cover images via `` tags:** Browser limitation — `` elements can't carry custom headers. Works if server doesn't require auth for cover endpoints.

Both author and cover images do not require authentication for the ABS server. Are you meaning the proxy needs an ignore pattern for checking the header?

PR #1770, opened 2026-01-02
Add chapter name to history view

### LINKED ISSUE #1766: [Enhancement]: Show chapter titles in history

When looking at the history for a audiobook show chapter titles, not just time stamps.

This would make it much easier to see what chapters I listened to and when.  Mostly because I'll fall asleep listening or zone out and it would really help to see chapter names in the history.  This would make it easier to decide how far back  to rewind to. 

Image

Android App - 0.11.0

Image

FreedomBen:
## Brief summary

Add chapter name to history view

Capturing the chapter name in the history view is helpful because it can greatly decrease the time it takes for the user to grok the location since it is usually more meaningful context indicator than timestamp alone.

## Which issue is fixed?

Fixes #1766

## Pull Request Type

Android, iOS.  No backend

## In-depth Description

**Add chapter name to history view**

Capturing the chapter name in the history view is helpful because it can greatly increase the time it takes for the user to grok the location since it is usually more meaningful context indicator than timestamp alone.

This commit adds the chapter name to the history, displayed just to the left of the timestamp.  If the chapter name is too limit, it is truncated with ellipsis (...)

## How have you tested this?

Yes, tested on Android Pixel 8 Pro.  Not tested on iOS (I don't have any Apple devices)

## Screenshots

Screenshot_20260102-081142

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/data/MediaItemHistory.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/data/MediaItemHistory.kt
@@ -14,6 +14,4 @@ class MediaItemHistory(
         var serverUserId: String?,
         var createdAt: Long,
         var events: MutableList<MediaItemEvent>,
-        // Snapshot of chapters when history was first created; used for offline and stable display
-        var chaptersSnapshot: List<BookChapter>? = null
 )
--- a/android/app/src/main/java/com/audiobookshelf/app/media/MediaEventManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/media/MediaEventManager.kt
@@ -91,12 +91,6 @@ object MediaEventManager {
             getMediaItemHistoryMediaItem(playbackSession.mediaItemId)
                     ?: createMediaItemHistoryForSession(playbackSession)
 
-    // Backfill chapter snapshot if missing
-    if ((mediaItemHistory.chaptersSnapshot == null || mediaItemHistory.chaptersSnapshot!!.isEmpty())
-            && playbackSession.chapters.isNotEmpty()) {
-      mediaItemHistory.chaptersSnapshot = playbackSession.chapters
-    }
-
     val mediaItemEvent =
             MediaItemEvent(
                     name = eventName,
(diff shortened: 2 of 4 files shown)


FreedomBen:
After testing this patch, if you go back to a version that doesn't have the code from this PR, the app will crash when trying to load history.  This patch (which can be applied with `git apply <patch-file>` will recover it.  I considered cleaning it up and committing it but it's such an edge case I'm not sure it's necessary, but posting here in case anybody tests this and wants to go back to a version without this PR:

```
diff --git a/android/app/src/main/java/com/audiobookshelf/app/managers/DbManager.kt b/android/app/src/main/java/com/audiobookshelf/app/managers/DbManager.kt
index b2f8293e..4af48227 100644
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/DbManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/DbManager.kt
@@ -271,7 +271,17 @@ class DbManager {
     Paper.book("mediaItemHistory").write(mediaItemHistory.id, mediaItemHistory)
   }
   fun getMediaItemHistory(id: String): MediaItemHistory? {
-    return Paper.book("mediaItemHistory").read(id)
+    return try {
+      Paper.book("mediaItemHistory").read(id)
+    } catch (e: Exception) {
+      Log.e(tag, "getMediaItemHistory: failed to read id=$id, deleting corrupted entry", e)
+      try {
+        Paper.book("mediaItemHistory").delete(id)
+      } catch (_: Exception) {  }
+      null
+    }
   }
 
   fun savePlaybackSession(playbackSession: PlaybackSession) {
```

nichwall (commented):
This is working in my testing on a Pixel 6a and looks good overall.

I am not sure if it makes more sense to store the chapter title with each `MediaItemEvent` instead of a single snapshot of the chapters. If we are storing a single snapshot of the chapters I think there needs to be an easy way to say "recalculate" or "reset" chapters because the History of an item cannot be deleted.

nichwall:
After thinking about this more, does it make sense to just not store the chapter information in either the History or events? If we're calculating the chapter title at runtime every time the page loads we could also just add the chapter information as another argument and not even store it on the History item.

FreedomBen:
> After thinking about this more, does it make sense to just not store the chapter information in either the History or events? If we're calculating the chapter title at runtime every time the page loads we could also just add the chapter information as another argument and not even store it on the History item.

Yeah the more I think about it, the more I come to same conclusion.  It doesn't seem like the runtime benefits of storing would be worth the syncing/denormalization cost.  I'm going to try reworking this to avoid storing and see how it goes

PR #1772, opened 2026-01-06
added 'sort by progress' entries

### LINKED ISSUE #1715: [Enhancement]: Sort by Progress: Last Updated

A sort by Progress: Last updated like on the web app.

Would be convenient for returning to things you previously listened to with more results than continue listening.

Image

Android App - 0.10.1

_No response_

KiwiHour:
## Brief summary
Added the ability to sort library items by progress, just like in the web app

## Which issue is fixed?
Adds requested enhancement and closes #1715 

## Pull Request Type
This affects both Android and iOS.
Only changes the frontend

## In-depth Description
There is not much more to add, this is just a port from the web app to add missing sorting functionality. It simply uses the pre-existing sorting information provided by the server, and I believe it is the optimal solution.

## How have you tested this?
I have tested this on both an emulated Android device and on my own phone with my actual server (with USB debugging).
On my phone, the sorting functionality works as expected, just like on the web app.

**I have not tested with iOS**. I don't imagine it would have any issues due to the small size of this change. If someone who can test this could so, that would be great.

## Screenshots
**Screenshot from the emulated device on Android Studio**
image

### DIFF
--- a/components/modals/OrderModal.vue
+++ b/components/modals/OrderModal.vue
@@ -1,7 +1,7 @@
 <template>
   <modals-modal v-model="show" width="90%">
     <div class="w-full h-full bg-primary rounded-lg border border-fg/20">
-      <ul class="w-full rounded-lg text-base" role="listbox" aria-labelledby="listbox-label">
+      <ul class="w-full rounded-lg text-base max-h-[70vh] overflow-y-auto overscroll-contain" role="listbox" aria-labelledby="listbox-label">
         <template v-for="item in items">
           <li :key="item.value" class="text-fg select-none relative py-4 pr-9 cursor-pointer" :class="item.value === selected ? 'bg-bg bg-opacity-50' : ''" role="option" @click="clickedOption(item.value)">
             <div class="flex items-center">


nichwall:
Functionally, this is working on my 6a, but we may want to change the modal to handle scrolling, similar to the "Filter" modal. The modal does not quite clip of the top/bottom of my screen but it is close and this is likely a problem with smaller phones.

KiwiHour:
> Functionally, this is working on my 6a, but we may want to change the modal to handle scrolling, similar to the "Filter" modal. The modal does not quite clip of the top/bottom of my screen but it is close and this is likely a problem with smaller phones.

Excellent point, I've added scrollability in the most recent commit

nichwall (approved):
Looks good to me, everything is working on my Pixel 6a.

A minor thing (but I don't think needs to be resolved for this PR) is that the modal does not scroll to the selected item when clicking on the sort button again so the selected item may not be visible initially. This behavior is the same as the Sort modal, though.

KiwiHour:
> Looks good to me, everything is working on my Pixel 6a.
> 
> A minor thing (but I don't think needs to be resolved for this PR) is that the modal does not scroll to the selected item when clicking on the sort button again so the selected item may not be visible initially. This behavior is the same as the Sort modal, though.

That is a good idea, but yeah probably a separate PR. Seems to require a chunk of JavaScript code than a simple tailwind class or something

advplyr:
Thanks!

PR #1774, opened 2026-01-07
Significantly improve Android UI performance when resuming from background

### LINKED ISSUE #1773: [Bug]: App is very slow to resume after being backgrounded

Yes

### What was the Problem?

When the app is foregrounded after spending time in the background, the UI can sometimes freeze and become non-responsive, especially after spending significant time backgrounded (such as when the sleep timer is on with the screen off).  It also shows stale data temporarily during the long resume process and controls are non-responsive.

- Start a book
- Set the sleep timer
- Turn off the screen
- Let the timer complete
- Unlock the phone
- Open ABS app if necessary
- Observe the bug

### What was Expected?

Quickly resume when foregrounded

### Phone Model

Pixel 8 Pro (but happens on older Pixel phones too, and I'm sure all Android phones)

### Phone OS

Android 15

Android App - 0.11.0

### Installation Source

Google Play Store

### Additional Notes

_No response_

FreedomBen:
## Brief summary

Significantly improve UI performance when resuming from background

When the app is backgrounded, diplay events were queued and could build up over time, leading to a sluggish resume when the app was foregrounded as the queue was processed.  As these updates are all stale by the time the app is foregrounded, we should ignore them to prevent a long resume process.

## Which issue is fixed?

Fixes #1773 

## Pull Request Type

Android only

## In-depth Description

- Track if we are foregrounded or not so we know when to ignore display events
- Don't queue metadata updates when backgrounded since there is no display to update and these end up queued
- On resume (when being foregrounded) send current state to UI so it stays in sync 
- Short-circuit player progress updates when backgrounded
- On resume (foregrounding) add short imperceptible delay before sending progress updates to allow web view time to initialize so we don't send the event too soon (fixes race condition)

This does not affect the notification tray indicator.

## How have you tested this?

Yes, tested on Pixel 8 Pro over the course of a few days (~ 12 listening sessions spread across multiple hours, including overnight)

## Screenshots

No UI changes

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/plugins/AbsAudioPlayer.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/plugins/AbsAudioPlayer.kt
@@ -82,6 +82,8 @@ class AbsAudioPlayer : Plugin() {
         }
 
         override fun onLocalMediaProgressUpdate(localMediaProgress: LocalMediaProgress) {
+          // Skip progress updates when app is backgrounded to prevent event queue buildup
+          if (!isInForeground) return
           notifyListeners("onLocalMediaProgressUpdate", JSObject(jacksonMapper.writeValueAsString(localMediaProgress)))
         }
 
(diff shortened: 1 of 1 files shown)


FreedomBen:
@nichwall I've been running this on my phone now for about a week and it's been a big improvement, but in the specific case of running on a very slow network (speed limited Tailscale connection back to home) with very high latency, it still performs roughly the same as it did before these changes (so no worse).  So I think they're still good and I don't have any concerns about merging, but I'm going to continue hunting down the issues in that particular case of slow/high latency network.  I think what's happening is the app is blocking on resume from background, waiting on the network connection.  Regardless, I strongly suspect any fixes from that will belong in a different PR anyway, so if your testing on this one looks good we can merge. 

BTW I'll be around to continue supporting all of my PRs so definitely let me know if you notice anything (before or after merge)

advplyr:
This is working in my tests. Thanks!

Issue #1161, opened 2024-04-02
[Bug]: Mobile app ignores external progress when playing


KRSogaard:
### Describe the issue

When moving from Web to Mobile decide, the mobile deceive will show the correct progress, but as soon as i click pay it will reset.
This happens to me almost every time i listen to an audiobook and switch device, example last night i was listening to a book up until "2:38:00" on my mobile phone, during the day today i have been listening via the web up until "5:07:02".
When i open my mobile app, it shows the correct time "5:07:02" as the current progress, but as soon as i click pay on the mobile device it resets to "2:38:00" and continues from there. To make the problem worse it "reports" this to set server as the current progress, overwriting my actual progress.

### Steps to reproduce the issue

1.  Listen on the mobile app, stop at a give time frame, called X time
2.  Listen on the webapp, stop at any given time frame a little more into the book, called Y time
3. Open the mobile app, see that the time should match where you left off on the web app, should match Y time
4. Click pay, see it reset to X time and continue playing from there

### Audiobookshelf version

v2.8.1

### How are you running audiobookshelf?

Docker, Android App

advplyr:
Please provide both the server version you have installed and the android app version you have installed. It is also helpful if you provide the android OS version you are using.

Also, it is important to add if you are experiencing this with downloaded books or streaming books or both.

KRSogaard:
I was able to catch it on video today: https://www.youtube.com/shorts/o7nfUfMNsHM
I have been listening on my laptop, and wanted to switch to my mobile. I stopped the playback on the web browser at 7:36:40, i then waited roughly 2 min to make sure the server has been updated. 
I then started the screen recording above, You will see that is set the current progress to 7:36:08 (a bit off the 7:36:40 on the web browser but this is ok), i then click the play button and i feel the phone vibrate twice and i see the progress jump back to 6:23:00 the progress i had when i stopped listening on my phone last time. 

The android app is the latest from the Google Play store, version 0.9.73-beta, and i am streaming the book.

advplyr:
In this case since the player has correctly synced the progress from the web in real-time using websockets that should not have jumped back like that. I'll look into it

KRSogaard:
@nichwall no i did not clock the app fully, it was running in the background. However i have seen it do the same thing when i force close the app first. Also it is clear it got the update from the server as it is a the right place before i push play, so i don't think that is the issue.

Inevitable:
[...]

PR #1777, opened 2026-01-10
Fix issue: 1161 ([Bug]: Mobile app ignores external progress when playing on web)

giejay:
## Brief summary
This PR tries to fix the issue which users were seeing when playing an audiobook on the web and continuing on the app. The server progress would not be visible in the app and it would use the local progress, overwriting any server progress.  

## Which issue is fixed?
https://github.com/advplyr/audiobookshelf-app/issues/1161

## Pull Request Type
Android + iOS, the backend

## In-depth Description
Whenever the websocket is disconnected, or the app has not been visible for 30 seconds, it syncs the local progress to the server and vice versa. 

## How have you tested this?
Played an audiobook on the server/web with app opened on my Android Phone
Scrolled through the audiobook in web, app progress was synced
Closed app for > 30 seconds
Saw a websocket disconnect in the logs
Opened app again, progress was synced -> This is not the case without this PR.

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -238,7 +238,7 @@ export default {
       const isMediaOpenInPlayer = this.$store.getters['getIsMediaStreaming'](prog.libraryItemId, prog.episodeId)
       if (isMediaOpenInPlayer && this.$store.getters['getCurrentPlaybackSessionId'] !== payload.sessionId && !this.$store.state.playerIsPlaying) {
         await AbsLogger.info({ tag: 'default', message: `userMediaProgressUpdate: Item is currently open in player, paused and this progress update is coming from a different session. Updating playback time to ${payload.data.currentTime}` })
-        this.$eventBus.$emit('playback-time-update', payload.data.currentTime)
+        this.$eventBus.$emit('playback-time-update', prog.currentTime)
       }
 
       // Get local media progress if exists
@@ -311,11 +311,14 @@ export default {
           headers: { Authorization: `Bearer ${serverConfig.token}` },
           connectTimeout: 6000
         })
-        if (response && response.data && response.data) {
+        if (response && response.data && response.data.mediaProgress) {
           console.log('✅ [default] Successfully refreshed user info/media progress from server on visibility change')
-          this.$store.commit('user/setUser', response.data)
+          response.data.mediaProgress.forEach(async (prog) => {
+            // Update local media progress
+            await this.userMediaProgressUpdated({ id: prog.id, data: prog, sessionId: 'visibility-change-sync' })
+          })
         } else {
-          console.error('[default] No user in /api/me response')
+          console.warn('[default] No mediaProgress in /api/me response')
         }
       } catch (err) {
         console.error('[default] Failed to refresh user info on visibility change', err)
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -328,12 +328,9 @@ export default {
       if (document.visibilityState === 'visible') {
         const elapsedTimeOutOfFocus = Date.now() - this.timeLostFocus
         console.log(`✅ [default] device visibility: has focus (${elapsedTimeOutOfFocus}ms out of focus)`)
-        // If device out of focus for more than 30s then reload local media progress
-        if (elapsedTimeOutOfFocus > 30000) {
-          console.log(`✅ [default] device visibility: syncing local + server media progress after > 30s out of focus`)
-          await this.syncLocalSessions(false).then(() => this.reloadServerMediaProgress());
-        } else if(!this.$socket?.connected){
-          console.log(`✅ [default] device visibility: syncing local + server media progress after websocket disconnect`)
+        // If device out of focus for more than 30s or websocket disconnected then sync local progress with server progress
+        if (elapsedTimeOutOfFocus > 30000 || !this.$socket?.connected) {
+          console.log(`✅ [default] device visibility: syncing local + server media progress ${elapsedTimeOutOfFocus > 30000 ? 'after being out of focus for more than 30s' : 'after websocket disconnect'}`)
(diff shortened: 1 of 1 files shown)


Serph91P:
This would be so great if this works. I hate it when I listen to something on the web player. Stop the playback and try later to start listening on the android app and it would start where I Left off on the android app and not the web player progress.

giejay:
I have been using this branch for the last 3 months now and it works perfectly for me. Not sure what to do to get this merged

Serph91P:
> I have been using this branch for the last 3 months now and it works perfectly for me. Not sure what to do to get this merged

The devs probably have a lot going on so my guess is time is what's needed.

layouts/default.vue
```diff
@@ -295,15 +295,43 @@ export default {
...
+        const response = await CapacitorHttp.get({
+          url: `${serverConfig.address}/api/me`,
+          headers: { Authorization: `Bearer ${serverConfig.token}` },
+          connectTimeout: 6000
+        })
+        if (response && response.data && response.data.mediaProgress) {
+          console.log('✅ [default] Successfully refreshed user info/media progress from server on visibility change')
+          response.data.mediaProgress.forEach(async (prog) => {
+            // Update local media progress
+            await this.userMediaProgressUpdated({ id: prog.id, data: prog, sessionId: 'visibility-change-sync' })
```
advplyr: I don't think this strategy is sufficient because some users may have hundreds of media progress records on their server. Running this every time the app doesn't have focus for 30 seconds would be expensive

giejay: Agreed. What would be better, only syncing the items which have changed in the last few weeks/days or moving the logic to whenever an item is played (and there was an out of focus event).

advplyr: If you look at the `syncLocalSessionsWithServer` function it is loading all load media progress, fetching the user from the server, then looping through and comparing timestamps to see which is more recent.

See that functions implementation on the android side here:
https://github.com/advplyr/audiobookshelf-app/blob/65ec1ec5d46e074500326f91a75e3bfe9f85c1df/android/app/src/main/java/com/audiobookshelf/app/server/ApiHandler.kt#L781-L850

In this PR you are basically doing the same thing twice.
Since I'm not able to reproduce this issue, can you test with only doing the `syncLocalSessions` call?

giejay: So I build the app again, this time with only the syncLocalSessions call and these are the steps to reproduce the issue:

- Play an audiobook using the app
- Pause
- Simulate a connection drop/websocket disconnect (I just turned off mobile data + wifi) but keep audiobookshelf app opened.
- Play the same audiobook using the browser, wait for the sync api call.
- Pause
- Now resume again on the mobile app by turning on the wifi/mobile data
- Inspect the history, it should have the synced state from the server
- But.. when hitting play, it continues using the last state of the mobile instead of the server state. 

So I don't think its an actual issue with the sync, the history shows the sync and shows the correct timestamp of the server. However, if the current player is on the foreground, it ignores that and uses local stored timestamp. 

Here some screenshots:

Mobile data is turned on again, we can clearly see the "Sync" from the server but the current playback of the player remains at 12th second instead of 32.
image

When hitting play, it uses the 13th second of the minute and seeks back 03 (I guess to replay last 10 seconds, thats fine).
image

giejay: Now I scanned through the code again, and I think this is the missing piece: 

fnc: userMediaProgressUpdated
```
      // Check if this media item is currently open in the player, paused, and this progress update is coming from a different session
      const isMediaOpenInPlayer = this.$store.getters['getIsMediaStreaming'](prog.libraryItemId, prog.episodeId)
      if (isMediaOpenInPlayer && this.$store.getters['getCurrentPlaybackSessionId'] !== payload.sessionId && !this.$store.state.playerIsPlaying) {
        await AbsLogger.info({ tag: 'default', message: `userMediaProgressUpdate: Item is currently open in player, paused and this progress update is coming from a different session. Updating playback time to ${payload.data.currentTime}` })
        this.$eventBus.$emit('playback-time-update', prog.currentTime)
      }
```

That's the code that also gets executed when I call the new reloadServerMediaProgress method.

advplyr: yeah so when visibility happens we could check what is the currently playing media item and only sync that

giejay: I modified the code to only reload the currently playing item

advplyr: This could work. I don't think we need to do the full syncLocalSessions in this case.
Have you been able to test this implementation?

giejay: I have and it's working fine now. Want to test it for a bit longer (also with Android auto) and I will also test it with removing the syncLocalSessions call.

advplyr: thanks, can you remove the syncLocalSessions? I'll be working on a mobile app release this weekend and would like to include this if we can reduce it down to only what is necessary for the issue you outlined

giejay: Done!

advplyr: I just realized I already implemented this in AudioPlayerContainer `deviceFocused` function, but it is only pulling server progress for non-downloaded media.
To confirm, the audiobook you are playing while testing is downloaded locally?

giejay: Yes, I usually download it and that's also the case while testing

advplyr:
I'm working on a more comprehensive fix in https://github.com/advplyr/audiobookshelf-app/pull/1864. Specifically, we need to build on the existing focus logic in `AudioPlayerContainer` as well as we need to handle the case where the app never lost focus and only lost network connectivity.

also for future reference, we should be making http requests using the `this.$nativeHttp` plugin because this handles token refreshes.
Thanks for your help debugging this, it has been very helpful

giejay:
Even better! Thanks!

PR #1804, opened 2026-02-14
Add multi-select mode for batch bookmark deletion

### LINKED ISSUE #1075: Remove all bookmarks button
I add frequent bookmarks due to listening while falling asleep. It would be convenient if I could remove these all at once instead of one at a time.

A button that would clear all bookmarks. This could live either in the bookmark pop up or in the triple dot menu for each book.

Another solution would be to be able to select multiple bookmarks for deletion in case the user would like to keep certain bookmarks.

abguymon:
## Brief summary

Adds multi-select mode to the bookmarks modal for batch deletion. Long-press a bookmark to enter selection mode, tap to select/deselect, with select all toggle and batch delete with confirmation dialog.

## Which issue is fixed?

Fixes #1075

## Pull Request Type

Affects both Android and iOS (frontend change only).

## In-depth Description

- Long-press any bookmark to enter multi-select mode
- Tap individual bookmarks to toggle selection
- Select all / deselect all toggle button
- Batch delete selected bookmarks with a confirmation dialog
- The "create bookmark" button is hidden during selection mode to avoid confusion
- New i18n strings added for the UI elements

This implements the alternative solution described in #1075 — rather than just a "clear all" button, users can select specific bookmarks for batch deletion, which gives more control while still solving the problem of clearing many bookmarks at once.

## How have you tested this?

yes

## Screenshots

https://github.com/user-attachments/assets/f4920d3e-7b00-4791-bfa4-a9108d3065df


strings/en-us.json
```diff
@@ -243,6 +244,7 @@
   "LabelSeason": "Season",
   "LabelSelectADevice": "Select a device",
   "LabelSelectMediaType": "Select media type",
+  "LabelSelected": "selected",
```
nichwall, line 247: Wrong capitalization

nichwall (commented):
Hey, sorry it has taken so long to review the PR, we've had a large backlog and the focus has been on the web client migration for the past months.

The functionality itself is working in my testing on Android 16 on a Pixel 6a.

I think the selection bar should override the "Create bookmark" at the bottom instead of a new bar being added to the top, this shifts where each bookmark appears once the first bookmark has been selected. Along those same lines, the multi-select is not persistent like the "Create bookmark" bar when the bookmark list is long enough to require scrolling.

I'm not sure the best way to resolve it in this PR, but when bulk deleting the bookmarks are deleted one at a time which looks a little funny. The API does not currently support bulk deleting bookmarks, so this seems to be the best way to handle it right now.

PR #1808, opened 2026-02-20
feat: Added option to have the sleep timer handle multiple chapters

indeedhat:
## Brief summary
Adds the option for the sleep timer to handle multiple chapters.
If enabled in the settings you will get additional [+]/[-] buttons to increment/decrement the timer by whole chapters.

## Which issue is fixed?
N/A

## Pull Request Type
Android only

## In-depth Description
Often times i want to use the sleep timer to go to the end of the next chapter or the one after that, with the existing setup of 5 minute increments this is pretty hard to achieve, i basically have to work out the expected end time and set a manual timer.

With this PR i have updated the sleep timer UI/logic so that you can increment either by the original 5 minute intervals or to the end of the next chapter. This functionality needs to first be enabled in the settings.

When enabled you get a new set of [+][-] buttons in the sleep timer modal for chapter inc/dec.

- If your current sleep timer is set for the end of a chapter then incrementing will increase it to the end of the next chapter.
- If your current timer does not line up with the end of a chapter then it will increment to the end of the chapter the timer ends within.
- Decrementing will always go back to the end of the previous chapter

## How have you tested this?
This has been tested both in an android emulator in AndroidStudio and on my personal device.

EDIT:
manual tests:
- single m4b file books
- one mp3 per chapter books
- swapping between books resets timer to end of current chapter
- skipping past end of sleep timer pauses playback
- shake to reset extends sleep timer by a chapter if its in the last 10 seconds
- shake to reset extends sleep timer by a chapter if recently finished
- sleep timer does not go past the end of the book
- pressing play after timer has just finished extends timer by chapter
- skipping forward by time increments or full chapters updates remaining time
- when mixing end of chapter and +/- 5 minute increments +/- chapter correctly gets the appropriate chapter boundary

## Screenshots
Settings Page
Screenshot_20260220_185704

Settings Page Info
Screenshot_20260220_185814

Sleep Timer
Screenshot_20260220_190055


indeedhat:
im putting this into draft as i have just discovered that it doesn't play nicely with shake to reset.
EDIT: fixed

nichwall:
I haven't looked through the code, but have you tested with both single-file and multi-file books? The sleep timer resets the timer when receiving a Play event, and switching to a new track will count as a Play event.

indeedhat:
> I haven't looked through the code, but have you tested with both single-file and multi-file books? The sleep timer resets the timer when receiving a Play event, and switching to a new track will count as a Play event.

Ive tested with both books that are in a single m4b and with a book that has each chapter as an individual mp3. I don't have any books that split it up in any other way such as multiple m4b's with multiple chapters each.

In both cases the sleep timer works as i would expect, it spans chapters without losing its place or randomly adding on time.

In the process of testing the multi track book i have just discovered an issue where if you have a sleep timer running and swap to another book it keeps the end time of the previous book rather than resetting it so you can end up with a sleep timer that has serveral hours left on it, ill look into how i can fix this.

indeedhat:
I apparently didn't submit the comment i added last night but i believe the above above issue has now been fixed, the PR description has been updated to list all of the scenarios i could think of to test.

PR #1809, opened 2026-02-21
perf: use server-side extraction for remote comic books

### LINKED ISSUE #3505: [Bug]: Big comics are basically unusable at the moment
### What happened?

Since the server was fixed to be better at importing cbr/cbz comic books (no crashes anymore, thanks!), I have been trying to implement a comic library. 

The built-in reader works okay for smaller books (20, 30mb filesize), but as soon as files get over in the multiple hundred MB range, they either load infinitely (never opening), or if they open (I tested this on a 200mb file), it can take up to 30-60 seconds for a single page turn, while the whole system basically becomes unresponsive. 

This makes using the comic book feature basically unusable at the moment. 

This was the same across the App and the Web client, so I assume it is a backend issue. 

### What did you expect to happen?

Smoother performance / working at all also for larger cb files

### Steps to reproduce the issue

1. Add a big comic book (300+ mb) to your library and try to open it. 

When it (eventually) opens, try to scroll through more than a few pages. 

### Audiobookshelf version

v2.14.0

### How are you running audiobookshelf?

Windows Tray App

### What OS is your Audiobookshelf server hosted from?

Windows

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

_No response_

### Additional Notes

_No response_

mattmanning:
## Summary
Large comic files (300+ MB CBR/CBZ) were causing 30-60 second load times because the entire file was downloaded to the mobile client before extraction could begin.

## Changes
- Uses new server-side `/api/items/:id/comic-pages` and `/api/items/:id/comic-page/:page` endpoints
- Pages are extracted and cached on the server, then sent individually to the client
- Preloads 2 pages ahead, 1 behind for smooth page turning
- Falls back to client-side extraction for local/downloaded files (file already on device)

## Dependencies
**Requires server PR:** advplyr/audiobookshelf#5078

The server PR adds the comic-pages and comic-page endpoints that this mobile app PR depends on. Both PRs need to be merged for the full fix.

## Testing
1. Merge server PR first (or test against a server with that PR applied)
2. Open a large CBR/CBZ comic from the server (not downloaded locally)
3. Pages should load quickly instead of waiting for full file download
4. Page navigation should be smooth with preloading

## Related
Fixes: advplyr/audiobookshelf#3505

### DIFF
--- a/components/readers/ComicReader.vue
+++ b/components/readers/ComicReader.vue
@@ -324,7 +324,7 @@ export default {
         this.fileIno = response.fileIno
         this.extractedItemId = itemId  // Store for use in getServerPageUrl
 
-        this.$toast.success(`Server extraction: ${this.numPages} pages`)
+        console.log(`[ComicReader] Server extraction ready: ${this.numPages} pages`)
         this.loading = false
 
         const startPage = this.savedPage > 0 && this.savedPage <= this.numPages ? this.savedPage : 1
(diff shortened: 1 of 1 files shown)


mattmanning:
**Testing Notes:**

✅ Successfully tested on Android (Pixel 10 Pro)
- Server-side comic page extraction working correctly
- Page turns are instant (previously 30-60 seconds for large comics)
- Preloading of adjacent pages working smoothly
- Fallback to client-side extraction works when server endpoint unavailable

Tested against server running PR #5078 (advplyr/audiobookshelf).

Issue #1419, opened 2024-12-27
[Enhancement]: Add mTLS support


MatBustamant:
### Describe the Feature/Enhancement

Add mTLS with client certificate support

### Why would this be helpful?

It would provide more options and flexibility to secure remote access to the server.

### Future Implementation (Screenshot)

It could prompt the user to select and allow the use of the certificate.

### Audiobookshelf App Version

Android App - 0.9.77

### Current Implementation (Screenshot)

It just fails to connect to the server.

KimbaTV:
This would be great...

the-mentor:
+1 on this

CZDanol:
+1 from me as well

PR #1819, opened 2026-03-09
Ssl cert http

ckbaker10:
## Brief summary

Adds mTLS authentication using certificates 

## Which issue is fixed?

https://github.com/advplyr/audiobookshelf-app/issues/1419

## Pull Request Type

Only affects android
Affects the backend of the app as well

## In-depth Description

I am working with mTLS and a custom CA to limit access to my internet facing services. Thus audiobookshelf needs mTLS for my requirements

## How have you tested this?

Yes I have tested it, the modifications are only shown when the server requests a certificate

## Screenshots

image


coxtor:
Awesome, thank you very much for implementing this!

nut-neek:
Thank you, amazing feature! Is there anything I can help with to get the PR ready for merging?

jasonboukheir:
This is really cool. In my case, I use a split setup, where accessing my instance locally does not require a cert, but accessing it publicly while I'm remote from my network does require a cert. In that case, my cert is optional, but my server returns a 403 on error.

My 2 cents: we should let the user set the cert before connecting. It can just be an advanced option below the server url.

ckbaker10:
that kind of setup is necessary for automation anyway since most tools do
not handle mtls well, I did it this way to allow for multi server setups
since the app allows to add multiple servers even though I guess only very
few people use that

Jason Elie Bou Kheir ***@***.***> schrieb am Mo., 13. Apr.
2026, 23:33:

coxtor:
Why has this been closed ? It would be a great addition to the app.

ckbaker10:
I am currently reworking my prs i have been testing them since creating them with my local built and I've not been happy with the quality yet. 

Before I resubmit them I'd also like to create a PR for unit tests so regressions can be avoided easier

https://github.com/ckbaker10/audiobookshelf-app/tree/android-jvm-test-framework

it's still a WIP though

Issue #1821, opened 2026-03-12
[Enhancement]: Add a text tint slider for font color in the ereader


abouthillier:
### Describe the Feature/Enhancement

Libby, Kindle, and other ereader apps allow for additional customization of the text color, mainly in dark mode. 

### Why would this be helpful?

I'd like to see or implement these added features to aid in low-light reading comfort.

### Future Implementation (Screenshot)

<img width="295" height="604" alt="Image" src="https://github.com/user-attachments/assets/16aabbbb-feb9-449f-a2d6-e5d042382eb7" />

### Audiobookshelf App Version

Android App - 0.11.0

PandaUnit-TWL:
[abs-ereader-text-tint.patch](https://github.com/user-attachments/files/26297388/abs-ereader-text-tint.patch)

PandaUnit-TWL:
## Implementation: Text Color slider for ereader (#1821)

Adds a "Text Color" slider to the ereader settings panel, positioned after "Font Boldness". This lets users dim the text for comfortable low-light reading — a common feature in Libby, Kindle, and other ereader apps.

### How it works

The slider ranges from 20 (dim) to 100 (full contrast), defaulting to 100 so existing behavior is unchanged.

- **Dark/Black themes:** 100 = `#ffffff`, 50 = `#808080`, 20 = `#333333`
- **Light theme:** 100 = `#000000` (inverted — 50 = `#808080`, 20 = `#cccccc`)

The tint value also flows through to `-webkit-text-stroke` so font boldness rendering stays consistent with the dimmed color.

The setting persists to `localStorage` through the existing `settingsUpdated()` → `loadEreaderSettings()` flow, same as all other ereader settings. Older stored settings without `textTint` default to 100 via nullish coalescing (`??`), so no migration is needed.

### Not changed

- **MobiReader** — does not use the same `themeRules` CSS injection system
- **ComicReader** — not applicable (image-based rendering)
- **Bottom status bar** — keeps its existing Tailwind opacity classes, unaffected by the slider

---

### Changes

**3 files changed, 23 insertions, 2 deletions**

---

#### `strings/en-us.json`

Added one string:

```json
  "LabelFontBoldness": "Font Boldness",
  "LabelFontColor": "Text Color",       // ← new
  "LabelFontFamily": "Font family",
```

---

#### `components/readers/Reader.vue`

Added `textTint: 100` to `ereaderSettings` defaults:

```js
ereaderSettings: {
[...]

PR #1822, opened 2026-03-12
Text tint for ereader

abouthillier:
## Brief summary

Added a tint slider to the ereader settings to further augment the theme controls in dark and black themes. This is intended to increase viewing comfort at lower light levels, beyond what phone defaults are capable of achieving. Google Play Books uses a toggle for this, which is fine, but I thought a slider would be a more aesthetic UX choice.

## Which issue is fixed?

Addresses Issue #1821 

## Pull Request Type

I have only tested this on Android, but it should work for iOS as well. This is a front-end change, and can easily be implemented in both the current web app, as well as the upcoming React front-end.

## In-depth Description

I've added a input element to the reader component, and modified the themeRules() in the epubReader component to read and parse the new setting. The tint input data is added to ereaderSettings alongside theme, font, etc. and is save to localStorage appropriately. 

In themeRules(), I've added tintBaseColor and tintFullColor variables that are then lerped between based on the percentage passed from the input slider. 

Once the tinted color is decided upon from the lerp, it is used to define the fontColor, as long as theme is set to dark or black.

## How have you tested this?

I have built the project on Windows and tested this in Android Studio, as well as on my Samsung Galaxy S20 FE

## Screenshots

studio64_CSoiOXeiHr


jlkalberer (commented):
I want to switch over to this app for my ebooks but it needs some love on the reader side of the app (this PR starts to address it). In one of the books I tried to read, all the text was blue and was basically unusable..

Anyways, I'd love to see this PR get merged but I'd think that just having a set of colors to switch between would be more user-friendly than another slider. (The sliders are hard to use IMHO).  I have only used Play Books but the settings there are per light/dark/middle with a night light toggle. I think the colors are calibrated to be accessible + WCAG and that's what you'd want here.

As is, this PR would move things in the right direction though so thanks for doing it!

PR #1832, opened 2026-03-28
Fix save password

ckbaker10:
## Brief summary

Adds Android Credential Manager integration so the app can save and retrieve login credentials via the system password manager (e.g., Google Password Manager), and enables proper `autocomplete` attributes on the login form inputs for better autofill support.

## Pull Request Type

- **Platform:** Android only (iOS unchanged)
- **Scope:** Frontend + Android native plugin

## In-depth Description

Currently the app has `autocomplete="off"` hardcoded on all text inputs, and there is no integration with the Android Credential Manager API. This means users cannot save or autofill their server login credentials using their device's password manager.

This PR adds:

1. **`AbsCredentialManager` native plugin** (Kotlin) — A Capacitor plugin that wraps Android's `androidx.credentials` API with two methods:
   - `saveCredential({ username, password })` — Saves credentials via `CreatePasswordRequest`. Called after a successful password login so the system prompts the user to save.
   - `getCredential()` — Retrieves saved credentials via `GetPasswordOption`. Called when the auth form is shown to pre-fill username/password if available.
   - Errors are handled gracefully (e.g., no credential provider configured, no saved credentials) — they log warnings but don't block the login flow.

2. **`AbsCredentialManager` JS bridge** — Capacitor plugin registration with a web fallback (no-op on web).

3. **`TextInput.vue` changes** — The component now accepts `autocomplete`, `name`, and `inputId` props instead of hardcoding `autocomplete="off"`, enabling proper HTML autofill semantics.

4. **`ServerConnectForm.vue` changes** — The username and password fields now have `autocomplete="username"` / `autocomplete="current-password"` and proper `name`/`id` attributes. After successful login, `saveCredential` is called. When the auth form appears, `getCredential` is called to pre-fill saved credentials.

5. **`MainActivity.kt`** — Registers the new plugin and sets `importantForAutofill = YES` on the WebView for Android O+.

6. **`build.gradle`** — Adds `androidx.credentials`, `credentials-play-services-auth`, and `lifecycle-runtime-ktx` dependencies.

The save will fail gracefully with a warning if the device has no credential provider configured (e.g., `CreateCredentialNoCreateOptionException: No create options available`). This does not affect the login flow.

## How have you tested this?

1. Clean install (cleared app data), entered server address, username, and password
2. Verified `AbsCredentialManager.saveCredential` is called after successful login (visible in logcat as `pluginId: AbsCredentialManager, methodName: saveCredential`)
3. Verified the error is caught gracefully when no credential provider is configured on the device
4. Verified the login flow completes successfully regardless of credential manager success/failure
5. Verified `AbsCredentialManager.getCredential` is called when the auth form is displayed after entering a server address


nichwall:
I've only taken a brief glance through this, but does it handle multiple server connections and OIDC? It doesn't look like it does.

I am not familiar with the Credential Manager API, is there a reason it needs to be included instead of just having the autofill set up for the frontend?

There are also some unrelated example test files that are deleted as part of this PR, why were those deleted?

capacitor.config.json
```diff
@@ -12,6 +12,6 @@
     }
   },
   "server": {
-    "androidScheme": "http"
+    "androidScheme": "https"
```
advplyr, line 15: Are there any compatibility issues that could come up with this change?

ckbaker10: Not really a big issue, it affects the web storage. Changing this you'd loose your current cookies/local storage and may have CORS issues if you are not using https and still set CORS to be active

CORS is a setting that will usually only be activated when running with valid certificates.

So no impact for selfhosters without certificates.

Issue #1428, opened 2025-01-04
[Bug]: Download percentages are in the billions


LilTrublMakr:
### What was the Problem?

When downloading a audiobook, sometimes the percentages are reaching in to the literal billions. I mainly notice this if one has tried to download an audiobook, it fails or freezes, and they try to download it again. But I have also seen it on first download attempt as well.

![Screenshot_20250104-100739](https://github.com/user-attachments/assets/0bf503b5-f266-4e05-adf9-b6e8e35c30cb)

### Steps to Reproduce the Issue

1. Download an audiobook.

### What was Expected?

Download percentage to be 2 digits. e.g. 37% (Maybe 3-4 if doing a decimal e.g. 37.42%)

### Phone Model

Pixel 8, Pixel 8 Pro

### Phone OS

Android 15

### Audiobookshelf App Version

Android App - 0.9.77

### Installation Source

Google Play Store

### Additional Notes

I have tried to find where the files are being downloaded to to delete them but am unable to find the files. In the app, all I can find is "Internal App Storage" but does not tell me where that is.

An idea would be to see if a file exists and expected file size matches before trying to download it again. Being able to pause/resume downloads would also be appreciated.

nichwall:
Internal App Storage is internal to the app, you will not be able to access it from the phone file system or other apps.

Does this only happen with multi-file books, or does it also happen with single-file books?

Are you connecting over LAN or downloading remotely? Downloading remotely will usually be much slower due to most people having much slower upload speeds through their ISP, so I'm wondering if this only happens when the download is incredibly slow.

LilTrublMakr:
It seems to only be doing it to multi file audiobooks. But I am also now unable to recreate the issue. I did add an external folder but am still downloading to the internal app storage.

I am downloading over LAN.

I set up everything yesterday (happy wife, happy life...)  so I am still learning the ins and outs of how things work.

LilTrublMakr:
And this is a bit of an odd one... Downloading to the Internal storage (/Audiobooks), the audiobooks are downloading but showing a 0% completion rate. Once the files are downloaded, they change to the green check mark. This is on the Pixel 8. I will see if I can recreate on my 8 Pro.

jpeel:
I tried to reproduce for a few minutes on a Pixel 9 Pro using Android 16, but haven't succeeded. The percentages are all 2 digits for me. Is it still happening with the latest release?

Issue #1167, opened 2024-04-05
Problem Downloading big Audiobooks


Maxklos:
### Steps to reproduce

1. Download a big Audiobook
2. Wait for download
3.

### Expected behaviour
- Tell us what should happen

### Actual behaviour
I use ADB for a while now and most of the times i don't encounter this problem. Now i have a Audiobook (533 Chapters -> 533 m4b files, ~43h, 2,45gb) I can listen to it fine when using the "Stream" Option or the Webinterface. When i try to download it, it takes a looong time (around 1h) then the app just hangs at "Downloading... Processing" indefinitely. When i change apps and come back the download is complete gone. I tried manually putting the files in the correct folder on my Phone but the same problems occur.
It doesn't matter if i download to the internal App Storage or a custom folder on the device.
Changing the Energy saving settings to "unrestricted" didn't help.

I looked at a couple of Issues here on Github which are similar:
https://github.com/advplyr/audiobookshelf-app/issues/816 -> Same Problem, but i don't use trafik
https://github.com/advplyr/audiobookshelf-app/issues/956 -> Would probaly help the download side

### Environment data

Audiobookshelf Version: 0.9.73-beta
Server Version: v2.8.1

- [x] Android App?

#### Android Issue
Android version: 
13 / 14

Device model:
- Pixel 3A
- Pixel 5a
- Pixel 6

Stock or customized system:
Stock

advplyr:
Are you downloading using a local ip address or a reverse proxy?

Maxklos:
Ah sorry, i forgot.
ABS is running on a Ubuntu Server and accessed through Cloudflaretunnel. I tested it with another server and direct IP access, it worked once but not reliable.

Another update. While writing the Issue I tested it again but with the screentimeout of my Phone set to 30min. With that the download worked and also the processing finished (processing was around 15min). With the same Audiobook converted to mp3 (3.50GB) the process is faster.

I think its an Problem with m4b files and the Socket disconnect if the Phonescreen locks / changes

bwprog:
I think I'm experiencing the same issue.
Android 14
Pixel 8
0.9.73-beta
ABS 2.8.1 and now 2.9.0 on Docker Linux
NPM latest or Caddy latest on docker linux -- I've tried going back and forth with same issue but both worked with same image/config a month ago.

[...]

PR #1833, opened 2026-03-28
Android: improve download throughput, safety, and concurrency

ricetim:
## Brief summary

Improves Android download speed, correctness, and concurrent download behaviour. This PR is based on top of #1587 and is intended to complement that refactor — if #1587 merges first, the diff here collapses to just the two files changed in the final commit.

## Which issue is fixed?

Related to #1167, #1428, #1479 (same issues referenced in #1587).

## Pull Request Type

Android backend only. No frontend or iOS changes.

## In-depth Description

All changes are in `InternalDownloadManager.kt` and `DownloadItemManager.kt`.

### `InternalDownloadManager`

**Shared `OkHttpClient` singleton**
Previously a new `OkHttpClient` was constructed per `InternalDownloadManager` instance, so each file download paid a fresh TLS handshake cost even when downloading multiple files from the same server. Moving it to a `companion object` with a `ConnectionPool(5)` reuses TCP connections across files.

**`readTimeout(30s)`**
Without a read timeout, if a server stops sending data while keeping the TCP connection open (e.g. a hung reverse proxy), the OkHttp dispatcher thread blocks on `inputStream.read()` indefinitely — even after the stall detector in `DownloadItemManager` has already removed the part from the active list. 30 seconds is generous enough that even very slow connections never false-fire (since `read()` returns as soon as any bytes arrive, not when the full buffer is filled), while ensuring the thread is eventually released via `SocketTimeoutException`.

**Buffer 8 KiB → 256 KiB, remove `BufferedInputStream`**
OkHttp/Okio already provides internal buffering on the response body stream; wrapping it in a `BufferedInputStream` with an 8 KiB window adds a redundant copy. A 256 KiB read buffer reduces the number of `read()`/`write()` syscall pairs by 32× for large audio files, which is the dominant overhead on a fast LAN connection.

**Catch `IOException` inside `write()`**
The outer `try/catch(Exception)` in `download()` wraps the synchronous setup before `enqueue()`, not the async callback. If `outputStream.write()` throws (e.g. disk full mid-transfer), the exception propagates through the OkHttp dispatcher which catches and logs it silently — `progressCallback.onComplete()` is never called and the download slot is permanently leaked, preventing any further downloads from starting. Catching `IOException` inside `write()` ensures `onComplete(true)` always fires.

**Call `onProgress` every chunk — no throttling**
The Capacitor bridge is crossed in the watcher loop on its own 500ms schedule, not in `onProgress`. Throttling `onProgress` would starve `lastUpdateTime` updates and cause the stall detector (introduced in #1587) to false-fire mid-transfer at moderate download speeds.

### `DownloadItemManager`

**`CopyOnWriteArrayList` for both lists**
`mutableListOf()` is not thread-safe. The watcher coroutine iterates `currentDownloadItemParts` while OkHttp callback threads and `Dispatchers.Main` coroutines call `add()`/`remove()`, which can produce `ConcurrentModificationException` in practice when downloading many files simultaneously.

**`@Volatile isDownloading`**
Without `@Volatile`, the JIT can cache the value in a register. A coroutine that sets `isDownloading = false` at the end of the watcher loop may never have that write seen by the thread checking it in `startWatchingDownloads`, causing a second watcher to silently never launch after the first one exits.

**`@Synchronized` on `addDownloadItem` and `checkUpdateDownloadQueue`**
When a large series (e.g. 90 files) is queued, metadata callbacks for all 90 items can complete near-simultaneously, causing multiple threads to call `checkUpdateDownloadQueue` concurrently. Each thread sees `downloadId == null` for the same `DownloadItemPart` and starts a duplicate `InternalDownloadManager` for it — two OkHttp calls writing interleaved bytes to the same file, corrupting it, while progress oscillates as two independent byte counters advance.

**Round-robin slot distribution**
With sequential slot filling, all available download slots are assigned to the first queued book before any slot is given to the second. Round-robin distributes one slot per book per pass, so all queued books make progress simultaneously.

## How have you tested this?

Built the APK from this branch using GitHub Actions and sideloaded it onto an Android device. Downloaded a large multi-file audiobook series over LAN from a local Audiobookshelf instance and verified that:
- Multiple files download concurrently
- Downloads complete correctly without corruption
- Progress reporting is stable (no backwards jumps)
- The full series completes without hanging at the "Processing" step

## Screenshots

N/A — backend only changes.


ricetim:
Closing in favour of a PR directly to the android_download_rewrite branch in #1587.

PR #1835, opened 2026-03-30
Fixed a progress saving error when connection to internet is lost

ckbaker10:
## Brief summary

When streaming an audiobook from the server and the network connection drops, playback progress is lost and the app does not automatically switch to a downloaded local copy. This PR fixes both issues by saving server progress locally as a fallback and by automatically switching to a local copy when the network is lost.

## Pull Request Type

- Affects: Android only
- Changes: Backend (native Kotlin player/sync logic)

## In-depth Description

There are two problems when streaming from the server and the network drops mid-playback:

### Problem 1: Progress is lost

In `MediaProgressSyncer.sync()`, the server-item code path only syncs progress to the server. When the network is down, it logs "Not sending progress to server" and discards the progress entirely -- nothing is saved locally. This means the user loses their listening position and has to manually find their place again.

**Fix:** When a server item cannot sync (network down or server sync failure), the progress is now saved to the local database as a `LocalMediaProgress` entry with a `server-offline-` prefixed ID. This ensures the listening position is always persisted. These offline entries are cleaned up naturally by the existing DB cleanup cycle after the progress is synced back to the server.

### Problem 2: No automatic fallback to local copy

When the network drops during server streaming, the player stalls and eventually errors out. The only existing fallback is direct play to transcode (HLS), which also requires a server connection. There is no fallback to a downloaded local copy of the same audiobook.

**Fix:** A new `switchToLocalCopyIfAvailable()` method checks if a local copy of the currently streaming audiobook exists (via `getLocalLibraryItemByLId`). If found, it seamlessly switches playback to the local copy at the same position. This is triggered in two places:

1. **Proactively** in the `networkCallback.onLost()` -- when the system detects network loss, the app immediately switches to the local copy before the player buffers run out.
2. **Reactively** in `handlePlayerPlaybackError()` -- if a playback error occurs on a server item, the app tries switching to a local copy before falling back to transcode or failing.

For podcasts, the correct local episode is matched via `serverEpisodeId`. The playback position (`currentTime`) is preserved across the switch since both server and local sessions use the same absolute time coordinate system.

### Files changed

- `MediaProgressSyncer.kt` -- Added `saveServerProgressLocally()` method; called when network is down or server sync fails in the non-local branch
- `PlayerNotificationService.kt` -- Added `switchToLocalCopyIfAvailable()` method; added `onLost` to `networkCallback`; modified `handlePlayerPlaybackError()` to try local copy first

## How have you tested this?

1. Stream an audiobook from the server that also has a downloaded local copy
2. Disable network (airplane mode / WiFi off) during playback
3. Verify: playback switches seamlessly to the local copy at the same position
4. Verify: progress is saved locally and visible in the app
5. Re-enable network
6. Verify: progress syncs back to the server

Also tested:
- Streaming without a local copy available: player behaves as before (stalls/errors, no crash)
- Network loss while not playing: no unnecessary switch attempt
- Podcast episode streaming with local copy: correct episode matched and switched

## Screenshots

N/A -- backend changes only, no UI modifications.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/media/MediaProgressSyncer.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/media/MediaProgressSyncer.kt
@@ -313,10 +313,17 @@ class MediaProgressSyncer(
             failedSyncs = 0
           }
           AbsLogger.error("MediaProgressSyncer", "sync: Progress sync failed (count: $failedSyncs) (title: \"$currentDisplayTitle\") (currentTime: $currentTime) (session id: $currentSessionId) (${DeviceManager.serverConnectionConfigName})")
+
+          // Save progress locally as fallback when server sync fails
+          currentPlaybackSession?.let { saveServerProgressLocally(it) }
         }
         cb(SyncResult(true, syncSuccess, errorMsg))
       }
     } else {
+      // Save progress locally as fallback when network is down
+      currentPlaybackSession?.let { saveServerProgressLocally(it) }
+      lastSyncTime = System.currentTimeMillis()
+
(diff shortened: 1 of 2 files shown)


kindermann-dev:
+1

PR #1845, opened 2026-04-13
Fix Android chapter-relative notifications

### LINKED ISSUE #239: Notification progress bar does not respect "chapter track" setting

Kyrluckechuck:
## Brief summary

Fixes the Android notification and Android Auto progress bar showing full-book progress for multi-chapter files (m4b, etc) instead of per-chapter progress. Brings the existing iOS chapter title display to Android, and adds a small new toggle for whether the secondary notification line shows the book title or the author.

Also fixes a minor second bug related to permission grants that stopped downloaded playback from being able to correctly stream the art/metadata since it wasn't an https URI.

## Which issue is fixed?

Fixes #239

## Pull Request Type

Mostly Android, with symmetric iOS changes. Touches both the native sides (Kotlin/Swift) and the JS frontend (Vue dropdown menu and backing code)

## In-depth Description

iOS already handled this by populating `MPNowPlayingInfoCenter` directly with chapter-relative duration and current time. Android can't do the same trick because `PlayerNotificationManager` and `MediaSessionConnector` both pull position and duration straight from the `Player` interface, so the override has to live at the player boundary. The method wraps `mPlayer` (and the cast player) in a `ChapterAwareForwardingPlayer` that overrides `getCurrentPosition`, `getDuration`, and `getBufferedPosition` with chapter-relative values when chapter mode is on. The real player stays untouched so the rest of the service still operates in absolute book time.

A few other things in here:
- A custom `MediaMetadataProvider` on the connector emits chapter-aware `MediaMetadataCompat` (chapter title in `METADATA_KEY_TITLE`, chapter duration, "Chapter X of Y" track number).
- A 1-second ticker that mirrors iOS's `addPeriodicTimeObserver` pattern. It only runs while playing and invalidates the system UI when natural playback crosses a chapter boundary so the displayed chapter title and duration update.
- `AbMediaDescriptionAdapter` reads directly from the playback session for the in-shade notification, since some Android versions populate that from `Builder.mContentText` rather than from the session metadata.

For sync, JS is the source of truth. Native caches in `SharedPreferences` (Android) and the existing Realm-backed `PlayerSettings` (iOS) so cold-start scenarios like Android Auto resuming before the JS app boots get the right behaviour. JS pushes the value to native on every toggle and app load

Also bundles a small fix for a pre-existing bug where offline-downloaded book covers were blank in Android Auto. The cause was that local FileProvider cover URIs were only being granted read permission to `com.android.systemui` (the lock screen), not to Android Auto's projection process, so Android Auto silently failed to load them and showed a blank background. Streamed covers were unaffected since those are HTTPS URLs that don't need a grant. The fix adds a small helper that grants permission to both processes and is shared between the metadata builder and the queue navigator. Happy to split this out into a separate PR if reviewers prefer.

A few decisions worth flagging:
1. The new "Show author below chapter title" toggle defaults off, which means the secondary notification slot shows the book title by default. iOS previously always had the author in that slot (via the artist field) when chapter mode was on. So existing iOS users will see a small visible change after upgrading: the line under the chapter title becomes the book title. They can flip the toggle on to restore the old behaviour, or if it's preferred I can make this setting default to on.

2. Setting `METADATA_KEY_ARTIST` to the book title in chapter mode is technically a small lie since "artist" semantically means author. The reason is that `Notification.MediaStyle` on Android pulls the secondary line from `METADATA_KEY_ARTIST` off the session and overrides whatever was set via `Builder.mContentText`. Author still flows through `METADATA_KEY_ALBUM_ARTIST` and `METADATA_KEY_AUTHOR` for any consumer that reads those (Bluetooth carkits typically read `ALBUM_ARTIST` as the secondary attribution).

3. The Realm schema version on iOS had to be bumped from 20 to 21 with a migration block for the new `useAuthorAsChapterSubtitle` property. Without this, existing iOS users would crash on app upgrade. Followed the same pattern as the existing chapterTrack migration.

4. One small pre-existing JS bug fixed along the way: AudioPlayer.vue `loadPlayerSettings` never pushed the persisted chapter-track value back down to native, so on cold-start the native side could be out of sync with the JS-stored value until the user toggled it. The push on load fixes this so it's the same on both platforms

Tested in various scenarios mostly involving m4b since that's what I have available to me. Tests include using Android Auto.
Would appreciate someone with iOS (and ideally CarPlay as well) to confirm the changes on that side work as expected, as additionally I do not have a Mac to build this on.

## Screenshots
Existing behaviour, showing full-duration in notification (released, same in AA):
image

Android notification, chapter mode on, m4b playing, chapter-relative time
image

Android Auto now-playing screen with the same m4b playing, showing Chapter title, book title in the secondary line, and chapter-relative scrubber
image

The new toggle:
image

Android notification with the new toggle on (secondary line shows the author)
image

Offline media cover's not displaying (non-streaming), including chapter notification changes:
image
After fix, including chapter notification changes:
image

## Notes
I did notice an additional bug with how the chapter skip-ahead works, where the skip to end of chapter doesn't actually progress the bar to the next chapter and sits at the end due to how the `floor(timestamp)` works if the chapter isn't rounded to a clean integer, and will follow that up in a future PR to keep this more tightly scoped with just the one additional bug fix for the metadata image.
Added PR https://github.com/advplyr/audiobookshelf-app/pull/1848 to fix this


brandonfhall:
LGTM — seek translation and metadata invalidation look correct.

KristjanTammekivi:
Tested it out on my emulator, works.
bump @advplyr

nichwall:
I'm in the process of reviewing https://github.com/advplyr/audiobookshelf-app/pull/1747 to migrate to media3. It might still be a few weeks based on the current rate of things.

I haven't looked at the code at all, does depend on exoplayer or does it not matter if switching to media3?

Kyrluckechuck:
> I haven't looked at the code at all, does depend on exoplayer or does it not matter if switching to media3?

Based on what I'm seeing in that PR, this would still be applicable, but would need some additional updates after to re-apply the same semantics to how the media3 implementation is currently designed, and a few things would need to be tweaked around the plugin boundaries since the additional call I added (such as `AbsAudioPlayer.setChapterTrack()`) would need to be routed through the backend abstraction where it picks which backend, but would be relatively small.

If that PR would be likely to be merged in the near future, I can wait and rebase this PR (I've just been running it in my local Android Studio build anyways for the last few months), or if it's preferred this could be merged first and that new backend refactor could likely include the compatibility changes without too much extra work

Issue #1624, opened 2025-07-08
[Enhancement]: next chapter time


Silther:
### Describe the Feature/Enhancement

Currently next chapter only starts once the time is greater than the chapter beginning. 
For example:
Chapter 1: [0s - 100s]
Chapter 2: ]100s - 200s]
Chapter 3: ]200s - End]

So if you go to time 100s it will show chapter 1, which seems like no big deal, but if you use the media controls to go to the start of the currently chapter it shortly switches to the previous chapter and than back to the current one. [will add video later]

What I would prefer:
chapter-end time is exclusive, chapter-start time is inclusive.

Chapter 1: [0s - 100s[
Chapter 2: [100s - 200s[
Chapter 3: [200s - End]

So the time 100s would be chapter 2 and the back button would not go to the previous chapter

### Why would this be helpful?

Less flickering between chapter if using media controls

### Future Implementation (Screenshot)

cannot be shown in a screenshot

### Audiobookshelf App Version

Android App - 0.9.81

### Current Implementation (Screenshot)

will upload later

PR #1848, opened 2026-04-14
Seek to millisecond rather than flooring to second

Kyrluckechuck:
## Brief summary
Fixes the "next chapter" button in the in-app player not advancing when a chapter starts at a non-integer second (common with m4b files). The player would land just before the chapter boundary and subsequent presses wouldn't move at all.

## Which issue is fixed?

n/a, discovered while testing a separate fix. Happy to file one first if preferred

## Pull Request Type

Android native plugin plus JS frontend. iOS already accepted Double values through the seek path so no Swift changes were needed.

## In-depth Description

The seek plugin truncated to integer seconds: `AbsAudioPlayer.seek({ value: Math.floor(time) })` on the JS side, and `call.getInt("value", 0)` on the Android side. If a chapter started at 100.5s, `Math.floor(100.5) = 100` was sent to native, which seeked to 100000ms.
That's still inside the previous chapter, which ends at 100500ms. Pressing next again produced the same result because `nextChapter` is computed by comparing against `currentTime`, and `currentTime` was still less than `nextChapter.start`.

Traced the Math.floor back to a 2022 commit that switched the plugin contract from a millisecond string to an integer-seconds value. The floor was defensive to match the new Int required value from Android, not a deliberate precision decision. iOS already accepted Double, so it had never suffered the truncation.

The fix drops the Math.floor in both JS callers of `AbsAudioPlayer.seek` and switches the Android plugin to read the value as a Double and convert to ms with full precision. `seekForward` and `seekBackward` are unchanged since their values are always integer-second jump intervals from device settings.

One minor side effect is that drag-to-seek on the timeline now lands exactly where you drop rather than at the floor of the nearest second. That's an improvement IMO, but also a tiny behavior shift, so flagging it here and it could potentially be worked around to stay as it was if that's preferred.

## How have you tested this?
- Tested by toggling on chapters in the player and using next/previous buttons

## Screenshots
- N/A

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/plugins/AbsAudioPlayer.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/plugins/AbsAudioPlayer.kt
@@ -325,10 +325,10 @@ class AbsAudioPlayer : Plugin() {
 
   @PluginMethod
   fun seek(call: PluginCall) {
-    val time:Int = call.getInt("value", 0) ?: 0 // Value in seconds
+    val time: Double = call.getDouble("value", 0.0) ?: 0.0 // Value in seconds, fractional
     Log.d(tag, "seek action to $time")
     Handler(Looper.getMainLooper()).post {
-      playerNotificationService.seekPlayer(time * 1000L) // convert to ms
+      playerNotificationService.seekPlayer((time * 1000L).toLong())
       call.resolve()
     }
   }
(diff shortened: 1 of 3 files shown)


nichwall:
Reproduced original issue and confirmed working with a single-file book on Pixel 6a with Android 16.

This also fixes https://github.com/advplyr/audiobookshelf-app/issues/1624

advplyr:
Thanks!

Issue #857, opened 2023-09-03
Progress percentage not shown correctly / only updated on app restart


0verEngineer:
### Steps to reproduce

1. Listen to an audiobook on the app (i listened to a downloaded one)
2. Check the progress percentage
3. Jump an hour in the audiobook progress
4. Check the progress percentage again, same as before
5. Restart the app to see the new progress percentage

### Expected behaviour
- The progress percentage should always be up to date

### Actual behaviour
- The progress percentage shows a wrong number

### Environment data

Audiobookshelf Version:

- [ x] Android App?
- [ ] iOS App?

#### Android Issue
Android version: 13

Device model: Google Pixel 7 Pro

Stock or customized system: CalyxOS 4.12.0

nichwall:
Are you referring to the progress percentage here? (There's a few places you could be referring to)

![Screenshot_20230903-193200.png](https://github.com/advplyr/audiobookshelf-app/assets/5686638/3dc33428-7bd5-4a2b-8434-2d7ca0e6f308)

The percentage and time remaining both update in this view without requiring an app restart, but it may only update if you are listening to the book or refresh the page (such as from restarting the app)

Did you resume playing the book after seeking an hour, or only seek? Did the time remaining change but the percentage didn't, or did neither change?

0verEngineer:
Yea i mean this one, both time and percentage are not updating if i randomly select (and of course play) chapters in a book.

advplyr:
This is intentional as it updates when the progress sync happens

0verEngineer:
@advplyr i pressed play and pause multiple times and it did not update, this is intentional?!? :D lol

0verEngineer:
And even if intentional, wrong information is shown to the user, so this is a bug, why do you close this? It affects multiple users as well.

advplyr:
Have you pressed play and listened for at least 15 seconds?
If you are on a metered network you would need to listen for 60 seconds since it syncs less on metered connections.

0verEngineer:
I can't remember, but i think i listened long enough because i wanted to continue with the book.

But imho this is irrelevant, wrong data should never be shown.

How does this happen, is this the old timestamp that is saved locally on the phone bcs then it would make sense to me.

advplyr:
It's relevant because that progress represents where you will pick up at if you force closed the app before the next sync happens. We don't save progress every second to prevent making many requests to the server. Progress is saved every 15 seconds for unmetered connections.

[...]

PR #1853, opened 2026-04-21
fix: prefer freshest local or server progress on detail pages

anatolyra:
## Brief summary

When a downloaded audiobook is played and the user navigates back to the book detail page, the progress percentage was stale. This is because the pages only read server progress (updated via Socket.IO roundtrip) and ignored locally-available progress (updated instantly via the native `onLocalMediaProgressUpdate` event).

Now each page compares local and server progress timestamps and displays whichever is more recent.

## Which issue is fixed?

Addresses the stale progress display reported in https://github.com/advplyr/audiobookshelf-app/issues/857

## Pull Request Type

Both android and ios, frontend Vue code only

## In-depth Description

The fix adds a `localProgressForServerItem` computed property that looks up local progress by server item ID (using the existing `getLocalMediaProgressByServerItemId` getter), and modifies `userItemProgress` to prefer whichever source has the more recent `lastUpdate` timestamp.

Affected views:
- Book detail page (`pages/item/_id/index.vue`)
- Podcast episode detail page (`pages/item/_id/_episode/index.vue`)
- Item more menu modal (`components/modals/ItemMoreMenuModal.vue`)

This eliminates the display lag because local progress is updated instantly via `onLocalMediaProgressUpdate` events from the native layer, while server progress only updates after a Socket.IO roundtrip.

## How have you tested this?

1. Download an audiobook
2. Play it for a while
3. Pause playback and navigate back to the book detail page
4. Confirm the progress percentage updates immediately (no delay)
5. Tested with both streaming and local playback

## Screenshots

N/A (logic-only change)

### DIFF
--- a/mixins/progressHelpers.js
+++ b/mixins/progressHelpers.js
@@ -1,0 +1,16 @@
+export default {
+  computed: {
+    localProgressForServerItem() {
+      if (!this.serverLibraryItemId) return null
+      return this.$store.getters['globals/getLocalMediaProgressByServerItemId'](this.serverLibraryItemId, this.serverEpisodeId)
+    }
+  },
+  methods: {
+    getFreshestProgress(server, local) {
+      if (!server && !local) return null
+      if (!server) return local
+      if (!local) return server
+      return local.lastUpdate > server.lastUpdate ? local : server
+    }
+  }
+}
--- a/pages/item/_id/_episode/index.vue
+++ b/pages/item/_id/_episode/index.vue
@@ -228,7 +228,12 @@ export default {
     },
     userItemProgress() {
       if (this.isLocal) return this.localItemProgress
-      return this.serverItemProgress
+      const server = this.serverItemProgress
+      const local = this.localProgressForServerItem
+      if (!server && !local) return null
+      if (!server) return local
+      if (!local) return server
+      return local.lastUpdate > server.lastUpdate ? local : server
     },
     localItemProgress() {
       if (!this.localLibraryItemId || !this.localEpisodeId) return null
@@ -237,6 +242,10 @@ export default {
     serverItemProgress() {
       if (!this.serverLibraryItemId || !this.serverEpisodeId) return null
       return this.$store.getters['user/getUserMediaProgress'](this.serverLibraryItemId, this.serverEpisodeId)
+    },
+    localProgressForServerItem() {
+      if (!this.serverLibraryItemId) return null
+      return this.$store.getters['globals/getLocalMediaProgressByServerItemId'](this.serverLibraryItemId, this.serverEpisodeId)
     },
     progressPercent() {
       return this.userItemProgress?.progress || 0
--- a/pages/item/_id/_episode/index.vue
+++ b/pages/item/_id/_episode/index.vue
@@ -57,6 +57,7 @@ import { Capacitor } from '@capacitor/core'
 import { Dialog } from '@capacitor/dialog'
 import { AbsFileSystem, AbsDownloader } from '@/plugins/capacitor'
 import cellularPermissionHelpers from '@/mixins/cellularPermissionHelpers'
+import progressHelpers from '@/mixins/progressHelpers'
 
 export default {
   async asyncData({ store, params, redirect, app }) {
@@ -116,7 +117,7 @@ export default {
       startingDownload: false
     }
   },
-  mixins: [cellularPermissionHelpers],
+  mixins: [cellularPermissionHelpers, progressHelpers],
   computed: {
     transformedDescription() {
       return this.parseDescription(this.description)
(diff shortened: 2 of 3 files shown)


itayavra:
Solves my issue! Would love to see this merged!

anatolyra:
@advplyr can you please take a look?

advplyr:
im not able to reproduce the original bug. can you go more in depth on reproduction steps?

the socket events come in through default.vue which checks the lastUpdate timestamp

advplyr:
does this have to do with a really slow device since the sync logic in default.vue has to update the database on the device that stores local progress?

anatolyra:
> im not able to reproduce the original bug. can you go more in depth on reproduction steps?

To reproduce:

1. **Download** an audiobook to the device
2. **Play** the downloaded book for a while (enough time for the progress percentage to change)
3. **Stop/pause** playback while on the player screen
4. **Minimize or close** the player (e.g., swipe down or tap back to the book detail page)
5. The progress percentage and time remaining shown on the book detail page are **stale** — they reflect the server's last-known progress, not the local progress that was just updated during playback

The progress only catches up after the server roundtrip completes — either on app restart, or after enough time passes for the next sync cycle to push a `user_media_progress_updated` Socket.IO event back to the client.

The key detail: the downloaded item has a server ID (e.g. `li_xxx`), so `isLocal` is `false`. That means `userItemProgress` returns `this.serverItemProgress`, which reads from `user/mediaProgress` in the Vuex store — the **server-side progress**. The instantly-available local progress in `globals/localMediaProgress` (updated reactively by `onLocalMediaProgressUpdate`) is never consulted for non-local items.

> does this have to do with a really slow device since the sync logic in default.vue has to update the database on the device that stores local progress?

No — I'm using a Galaxy S25 Ultra. The issue isn't a slow device or slow DB writes. The native player emits `onLocalMediaProgressUpdate` events fine, and they update `globals/localMediaProgress` in Vuex reactively. The problem is that `userItemProgress` for server-linked items only reads from `user/mediaProgress` (server progress), so even though local progress is available instantly in the store, the detail page doesn't use it. It waits for the full roundtrip: native → HTTP sync to server → Socket.IO push back → `userMediaProgressUpdated` handler in `default.vue` → Vuex `user/mediaProgress` update.

advplyr:
it seems like it may not be the round trip that is the problem, it is that webview is not getting the socket events after a while if the app is in the background. the socket events are what update the server progress in memory that the book details is reading from.

for the reproduction steps, are you listening to the book while the app is minimize/in the background? or are you keeping the app open to the player the entire time you are listening?

anatolyra:
> for the reproduction steps, are you listening to the book while the app is minimize/in the background? or are you keeping the app open to the player the entire time you are listening?

Both. Most of the time the app is open with the player the entire time. Sometimes I minimize it to use some other app.

anatolyra:
> it seems like it may not be the round trip that is the problem, it is that webview is not getting the socket events after a while if the app is in the background. the socket events are what update the server progress in memory that the book details is reading from.

You raise a fair point, background event delivery is definitely an issue. On Android, `AbsAudioPlayer.kt` explicitly drops `onLocalMediaProgressUpdate` events when `isInForeground` is false, and the WebView suspends JS execution, so Socket.IO events are missed. The `handleOnResume` only sends the last snapshot, not missed updates. So I agree the background case is real.

But the foreground case is also affected, which is what I'm seeing most often. Here's why:

When I'm listening with the app open, `onLocalMediaProgressUpdate` fires every 15 seconds from the native layer and immediately updates `globals/localMediaProgress` in Vuex. But for a downloaded book with a server ID, `userItemProgress` on the detail page only reads from `serverItemProgress`, which comes from `user/mediaProgress`. That store only gets updated after the full round-trip: native player → HTTP POST to server → server processes it → Socket.IO push back to client → `userMediaProgressUpdated` handler in `default.vue` → `user/updateUserMediaProgress`.

The local progress is already sitting in Vuex, but the detail page never looks at it. So there's a visible window, sometimes several seconds, where the page shows stale server progress instead of the fresher local progress.

This PR addresses both cases: `getFreshestProgress()` compares `lastUpdate` timestamps and returns whichever is more recent. In the foreground, it eliminates the round-trip delay. After backgrounding, it picks up local progress from the SQLite reload that `visibilityChanged` already does when coming back.

advplyr:
if it's just the round trip delay that you are experiencing, does that mean the progress does get updated after a second or two? and this PR is eliminating that second or two delay from the round trip?

the linked issue doesn't clearly state the problem. some comments suggest it is not showing accurate times at any point, some comments suggest it is a few second delay they are concerned about.

anatolyra:
It's more than a second or two. It takes time (maybe the duration I mentioned in my previous comment was not accurate, it definetly takes a long time). I usually have to kill the app and open again. It bothers me because I like to update goodreads at the end of the day.

advplyr:
ah ok, so it isn't the round trip that is the issue. while this pr might address an issue we should still try to figure out what that actual issue is because it may affect other parts of the app. unless im misunderstanding what you outlined so far

Issue #1838, opened 2026-03-31
[Bug]: Download Timeout


GothNinja:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

When downloading books for offline listening the download stalls when the app isn't running in the foreground. I have unrestricted battery usage turned on for the app. 

If I am listening to something while downloading I do not have this issue.

### Steps to Reproduce the Issue

1. Initiate download
2. Switch apps on device for a few minutes 
3. Return to ABS app
4. Download is now frozen and doesn't resume unless the app is closed, reopened and the download is reinitiated.

### What was Expected?

The app to retain server connection in the background until download completion.

### Phone Model

Pixel 9 pro

### Phone OS

Android 16

### Audiobookshelf App Version

Android App - 0.12.0-beta

### Installation Source

Google Play Store

### Additional Notes

Could a download progress notification be added to Android to ensure that connection is retained in the background?

Zageron:
I've been meaning to report this for ages and ages. Thanks for reporting it. 😂 
I encounter this almost daily.

dedors:
I do have the same problem. I also noticed an issue where a download finishes, but the post-processing phase (which can take several minutes) is interrupted if the app is not in the foreground. This shortens the total book length, causing playback to end prematurely. To fix that, it must be deleted and redownloaded. Currently, keeping the app open in the foreground with the screen active seems to be the only reliable way to prevent this.

Also Pixel 9 Pro, A16, 0.12.0-beta

WhiteSockedDancer:
I've got the same since forever.
On Samsung Galaxy S21 FE

My workaround: Screen timeout to 10min for the download.
When forgotten and interrupted: Kill the app and restart download.

jcrmit:
I can confirm this is an issue for me too. Pixel 9. Downloads do not continue if the app is not open and in the foreground. Leaving the app, stalls the download and the app needs to be closed and reopened to restart the download.

jimm0thy:
Just noticed this issue as well with my S25

misacek007:
It is also worth noting that it is all or nothing so when you get interrupted you do not get to keep what you already downloaded. You also get interrupted when network changes (wifi/vpn).

This is very annoying bug that would be worth looking at.

TacticalLaptopBag:
[...]

PR #1858, opened 2026-04-29
[Android] Restore download queue on app kill/restore

eseverson:
## Brief summary

  Restore Android download queue across app suspends. Persisted downloads are now read back on launch, per-part progress is saved as parts complete, and resume picks up from the last known state instead of restarting from zero.

##  Which issue is fixed?

Partially addresses #1838 - after the user returns to the app, the queue is restored and in-flight downloads pick up where they left off, so the "close, reopen, reinitiate the download" workaround is no longer needed. Does not prevent the original stall while backgrounded; that requires a foreground service, which is out of scope here.

There are also several other download-queue related issues.

##  Pull Request Type

  Android only - frontend (one Vue component, gated $platform === 'android') + native Kotlin. iOS unaffected.

##  In-depth Description

  The bug had two layers:
  1. DbManager.getDownloadItems() had zero callers, so a killed app forgot every in-flight download.
  2. saveDownloadItem was only called once at queue-add time, so even if the queue had been restored, parts would all look fresh and re-download.

 ### Changes

  - DownloadItemPart - Uri fields don't survive serialization, so they're now computed getters over String paths. Added destinationPath so the SAF temp path survives restart. serverUrl is rebuilt from serverPath + DeviceManager.{serverAddress,token} on each access (also handles token rotation across 
  suspends).
  - DownloadItemManager —-New restoreQueue() re-hydrates the in-memory queue, re-attaches external parts to the system DownloadManager by their saved downloadId, and re-enqueues parts that can't be resumed (in-process OkHttp). Bails out if no server is connected. Self-heals from disk: internal-storage parts whose file already exists at the expected size are adopted as completed regardless of what the persisted record claims -this is what makes the fix work for records written by older builds. New persistParentOf() saves at every state transition: downloadId assigned, completion, move success/failure.                

 ### Out of scope

  - Foreground service / WorkManager migration - proper long-term fix for backgrounded downloads (and the full fix for #1838), but a much larger refactor.

 ## How have you tested this?

  - ./gradlew :app:testDebugUnitTest -8/8 pass.
  - ./gradlew assembleDebug - clean.
  - On emulator and on physical Pixel 9 pro: queue an audiobook, let several parts complete, force-stop, reopen + reconnect. Restore picks up correctly: completed parts skipped, queue count accurate, remaining parts continue.  Tested with internal app storage and sdcard.

##  Screenshots

  No visual changes.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -127,6 +127,8 @@ dependencies {
 
   // Jackson for JSON
   implementation 'com.fasterxml.jackson.module:jackson-module-kotlin:2.12.2'
+
+  testImplementation 'junit:junit:4.13.2'
 }
 
 apply from: 'capacitor.build.gradle'
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
@@ -75,6 +75,98 @@ class DownloadItemManager(
     checkUpdateDownloadQueue()
   }
 
+  /**
+   * Re-hydrate the in-memory queue from persisted state and resume any in-flight downloads.
+   * Idempotent: safe to call multiple times. Bails out if no server is connected yet, since any
+   * internal-storage part needs a fresh signed URL to re-enqueue.
+   */
+  fun restoreQueue() {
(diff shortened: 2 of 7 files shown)


advplyr:

Issue #656, opened 2023-04-01
Better display in landscape mode


Deanna4711:
Hi!
First of all: Thanks for this great piece of software!
I use the Android app on my Android car radio. So this is obviously in landscape mode. But in landscape mode there is not much space to scroll through the libary, because the start/forward/rewind buttons on the bottom and the top menu use most of the display space.
Is it possible to reduce or minimize the control menu and buttons to get a better navigation expierience?

Even if not, thanks a lot! :-)

hobesman:
At one point I had similar ideas for the playback screen in landscape mode.

![Screenshot_20221121-224321_Gallery](https://user-images.githubusercontent.com/29721765/229312829-d7d7625e-e5c1-421a-bdd8-724cad324710.jpg)

For browsing, the left side could be used for browsing and keep the right side as playback controls. Just an idea for a starting point.

mdeeter:
Even in the vertical view, I think the playbars are too close to the edges.  I agree that I like this side-by-side view... just 2-cents to make the playbars a little narrower.

Issue #1343, opened 2024-10-12
[Bug]: When going from portrait mode to landscape and back, the progress bar bugs out


ScuttleSE:
### What was the Problem?

When going from portrait mode to landscape and back, the progress bar bugs out

https://github.com/user-attachments/assets/eedf6758-e765-4337-8441-d9adcd4843b9

### Steps to Reproduce the Issue

1. Start playing a book or podcast
2. Tilt the phone to landscape
3. Tilt the phone to portrait

### What was Expected?

I expected the progress bar to not bug out.

### Phone Model

iPhone 14 Plus

### Phone OS

iOS 18

### Audiobookshelf App Version

iOS App - 0.9.76

### Installation Source

Testflight

mdeeter:
I too have seen this... on both my old iPhone 12 Pro (pre iOS 18)... and also on my new iPhone 16.

PR #1861, opened 2026-05-07
Player view in landscape mode

lkiesow:
Rotating your phone to landscape mode and opening the player, the view wasn't using the available space very efficiently. Especially the cover image is not very helpful being pushed to a minimal size.

This patch redesigned the view in landscape mode, pushing the cover image to the left half of the screen while the controls are moved to the right half of the screen.

This will affect both Android and iOS.

#### User Interface Preview

Here is how the view looks with and without this patch:

https://github.com/user-attachments/assets/9ac06a0d-bbc6-4d96-afce-3ad59ddc8858


hobesman:
I can't comment on the underlying code, but this is the exact solution I proposed back in 2022. https://discord.com/channels/942908292873723984/942908295579062335/1045749006829097011

This was my mock-up at the time.

Screenshot_20221121-224321_Gallery-1


My only proposal would be to have the book title somewhere (perhaps the to of the right side?) rather than rely on the cover art to convey the title.

Side note: is there a reason the chapter title and author text shrinks in landscape compared to portrait layout?

lkiesow:
Good catch with the font sizes. Fixed that. This is how the UI looks in landscape mode now:

Screenshot from 2026-05-10 00-55-19

I briefly looked into adding the title, but the text over the author already covers displaying the title, chapter,  or both and doing it differently from portrait mode is probably a bit confusing. We could just not hide the book title in fullscreen mode if there is a chapter. Then, the text over the author would be `<title> | <chapter>`. But I think that's a separate discussion. I can file a quick pull request for that, though.

EDIT: Filed #1865 to discuss adding the title in funnscreen player mode.

hobesman:
Some follow up questions from a non-coder.

How does this fare on a non-widescreen device, like a 4:3 tablet screen? 

How does this handle future changes to the UI? Does it mean all future changes have to be made twice? Or if an element is deleted or added in portrait it will automatically update landscape layout as well? 

If not automatic, could that be done? Like perhaps those buttons are added to a "second half" layout that gets placed either at the bottom or on the right depending on screen orientation. That way for future changes it's one layout with one update that applies in both orientations.

Just some thoughts.

lkiesow:
> How does this fare on a non-widescreen device, like a 4:3 tablet screen?

This is the UI on a 4:3 screen in portrait and landscape mode:

Screenshot from 2026-06-16 18-22-52
Screenshot from 2026-06-16 18-22-42


> How does this handle future changes to the UI? Does it mean all future changes have to be made twice? Or if an element is deleted or added in portrait it will automatically update landscape layout as well?

All of this is just CSS, meaning that it just changes the layout of elements. No new elements are added or duplicated. Can this still mean you may need some adjustment for one or the other mode to look pretty. Sure. But you definitely do not need to do everything twice. In fact, before, you would have also needed to check if a change would break in any orientation. So, testing is not much different than before.

Designing two entirely different layouts would have probably been easier and this is therefor a bit of a compromise, But I also didn't want to maintain things twice :grin: 

> If not automatic, could that be done? Like perhaps those buttons are added to a "second half" layout that gets placed either at the bottom or on the right depending on screen orientation. That way for future changes it's one layout with one update that applies in both orientations.

That's actually more or less how it works.

nichwall (commented):
Overall, this is looking good in my testing on Android 16 with a Pixel 6a.

I did notice an issue with the chapter track width not being recalculated in landscape mode so it sticks off the edge of the screen (related to https://github.com/advplyr/audiobookshelf-app/issues/1343). The "total book" track is resized correctly.

Image

nichwall:
For completeness, this should fix https://github.com/advplyr/audiobookshelf-app/issues/656

advplyr:
This definitely looks better, but my original hesitation on this design choice is that some users may be using landscape for the wider track bar.

We can try it out and see what kind of feedback it gets.

Thanks!

hobesman:
Did the progress bar issue raised by @nichwall get resolved? 

Also, to @advplyr's point, could a compromise be to put the progress bars below the buttons and that way it can still span the entire width?

advplyr:
Yes I fixed the bar overflow. If we put the bar underneath it needs to be able to also show the optional chapter track bar

PR #1864, opened 2026-05-09
Fix external progress sync for open sessions

### LINKED ISSUE #1161: [Bug]: Mobile app ignores external progress when playing

advplyr:
## Brief summary

This handles the case where a local media item is open and paused in the player and external media progress has been made that needs to get synced.

This has also been getting discussed and tested in https://github.com/advplyr/audiobookshelf-app/pull/1777

## Which issue is fixed?

Fixes #1161 and probably many others

## Pull Request Type

Android and iOS

## In-depth Description

Users frequently leave the app open and media player open but paused. When the app doesn't have focus for an extended period of time, most devices will kill the socket connection. External media progress events are not received if the socket is not connected.

This PR solves this by fetching the server media progress after 30 seconds or more where the device did not have focus OR the socket was disconnected.
For local media items, the local media progress is compared with the server media progress. If the server media progress has been updated more recently, then the current playback session time is updated and the local media progress is updated.

## How have you tested this?

Testing steps:
1. Start playing a book downloaded locally
2. Pause after ~30 seconds to ensure media progress update was made to the server
3. Leave the player open and turn off device network connection
4. In the web app start playing the same book
5. Pause after the server media progress is updated. It is best to jump way ahead so it is noticeable.
6. Reconnect the device network on mobile
7. Play button will show a loading state while it fetches the server progress

### DIFF
--- a/components/app/AudioPlayer.vue
+++ b/components/app/AudioPlayer.vue
@@ -69,22 +69,22 @@
 
       <div id="playerControls" class="absolute right-0 bottom-0 mx-auto" style="max-width: 414px">
         <div class="flex items-center max-w-full" :class="playerSettings.lockUi ? 'justify-center' : 'justify-between'">
-          <span v-show="showFullscreen && !playerSettings.lockUi" class="material-symbols next-icon text-fg cursor-pointer" :class="isLoading ? 'text-opacity-10' : 'text-opacity-75'" @click.stop="jumpChapterStart">first_page</span>
-          <div v-show="!playerSettings.lockUi" class="jump-icon text-fg cursor-pointer flex flex-col items-center" :class="isLoading ? 'text-opacity-10' : 'text-opacity-75'" @click.stop="jumpBackwards">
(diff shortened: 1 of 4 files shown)


advplyr:
This also handles when streaming a media item and keeping it paused. The server progress was getting properly updated in the track bar, but was not being updated in the native audio player session. So when pressing play, the time jumped back to the native audio player session time.

PR #1869, opened 2026-05-14
Send less data to server with progress update.

### LINKED ISSUE #238: Local playback generating a huge mobile data usage
### Steps to reproduce

1. Download a Audiobook 
2. Keep the audiobook playing for a few hours

### Expected behaviour
- With local downloaded audiobooks there should only be minimal bandwith usage. The bandwith should **not** be much higher than the actual local file size of the audiobook

### Actual behaviour
-  After playback and long usage the app generates huges amount of traffic (>2GB) even though its only playing a single previous downloaded audiobook

### Environment data

Audiobookshelf Version: 0.9.46-beta

- [ X ] Android App?
- [ ] iOS App?

#### Android Issue
On newest Samsung and Lineage OS

AndrolGenhald:
## Brief summary

Reduce the amount of data sent with each progress update.

## Which issue is fixed?

Fixes #238

## Pull Request Type

Android only, however I believe the same issue exists for iOS. Sadly I don't have a Mac or iOS device to test with, so someone else will have to fix that.

## In-depth Description

A bunch of unnecessary metadata is being sent on every progress update. The server doesn't even parse most of these fields, so sending them is completely unnecessary. All together, for the book I checked, the original request was 150kb, sent every minute. Listening for 10 hours updating the progress once a minute that's around 88MiB. Last month I had almost 1GiB of mobile data eaten by audiobookshelf despite exclusively listening to locally downloaded books.

By removing unneeded data I reduced the payload to around 1KiB, including HTTP headers. For most books I expect this will be 100-200x smaller, although just how large the previous payload was varied based on number of chapters, number of audio tracks, etc.

## How have you tested this?

Checked network requests in android studio, body size for /api/session/local POSTs has been reduce (for a particular book) from 150KiB to 556 bytes.
Deleted session in sqlite database, verified it was recreated correctly.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/server/ApiHandler.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/server/ApiHandler.kt
@@ -17,6 +17,7 @@ import com.audiobookshelf.app.plugins.AbsLogger
 import com.audiobookshelf.app.managers.SecureStorage
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import com.fasterxml.jackson.core.json.JsonReadFeature
+import com.fasterxml.jackson.databind.node.ObjectNode
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import com.getcapacitor.JSArray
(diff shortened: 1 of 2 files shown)


advplyr:
I implemented this on the iOS side as well. Thanks!

AndrolGenhald:
Thanks for getting back to this and fixing the iOS side, glad to see it merged! I was close to sideloading a fixed apk a few times but it was never _quite_ bad enough for me to get around to it :smile:

Issue #1781, opened 2026-01-15
[Bug]: Search results is limited to 5 results


timosturm:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

The search is limited to 5 results; no indicator is present that some results are not shown.

### Steps to Reproduce the Issue

1. Open the search
2. Enter some keyword that should yield more than 5 books
3. Observe that the results are limited

I think the relevant line is this: 

https://github.com/advplyr/audiobookshelf-app/blob/b2c2b625e3cf586562397f4c02e38059f2b8ef5c/pages/search.vue#L114

### What was Expected?

I expect to be able to find all matches. Each category seems to be limited to max. 5 entries.
The sub-entries should be scrollabel, or maybe have a button "show all results".

### Phone Model

Samsung S9e

### Phone OS

Android 14

### Audiobookshelf App Version

Android App - 0.11.0

### Installation Source

Google Play Store

### Additional Notes

How  it is implemented now is very confusing, because at first it seems like the server is not recognizing / missing the other results when. However the search works fine and as expected in the web version.

InsaneOP1990:
Me and a friend noticed the same issue 

Dungeon Crawler Carl wont even show as an option when searching "Dungeon" on books or Series

but when you search Carl you get see a Authors section and Narrators section

<img width="947" height="938" alt="Image" src="https://github.com/user-attachments/assets/71d57fa5-bfab-4664-8191-d2043e8e3989" />

PR #1879, opened 2026-05-25
Remove hardcoded search result limit of 5

### LINKED ISSUE #1872: [Enhancement]: Show more than 5 results in search results

https://github.com/advplyr/audiobookshelf/issues/5245#issuecomment-4471736949

Currently excludes obvious search results due to arbitrary and unnecessary limit.

app version is is 0.12.0-beta

same as current but with search showing all results

Android App - 0.12.0

Image
Image

OguzIronCode:
## Brief summary

Removed the hardcoded search result limit parameter (`limit=5`) from the library search API request in the app.

## Which issue is fixed?

Fixes #1872

## Pull Request Type

This change affects both Android and iOS apps as it is a shared frontend code modification in `pages/search.vue`.

## In-depth Description

In `pages/search.vue`, the search API call was sent with a hardcoded `limit=5` parameter. This restricted the app's search results to only 5 items per category (books, podcasts, authors, etc.) even when more matches were available on the server. By removing this hardcoded limit, the API request will now allow the server's default pagination or full search results to be returned and rendered in the app.

## How have you tested this?

1. Built and ran the app locally.
2. Performed searches that are known to return more than 5 results (e.g. searching for a common letter or tag).
3. Verified that more than 5 results are now displayed correctly in the search results page.

### DIFF
--- a/pages/search.vue
+++ b/pages/search.vue
@@ -122,7 +122,7 @@ export default {
         return
       }
       this.isFetching = true
-      const results = await this.$nativeHttp.get(`/api/libraries/${this.currentLibraryId}/search?q=${value}&limit=5`).catch((error) => {
+      const results = await this.$nativeHttp.get(`/api/libraries/${this.currentLibraryId}/search?q=${value}`).catch((error) => {
         console.error('Search error', error)
         return null
       })


nichwall:
This is working in my testing on Android 16 with a Pixel 6a.
As of server v2.36.0, the API uses a default search limit of 12 which is working and renders correctly in my testing.

nichwall:
This will also resolve https://github.com/advplyr/audiobookshelf-app/issues/1781

advplyr:
Thanks!

PR #1880, opened 2026-05-25
[Enhancement]: Add autoplay next podcast episode setting

### LINKED ISSUE #416: Provide ability to queue/playlist audiobooks
The kids sometimes listen to audiobooks when going to sleep. They can choose one each. At the moment, after the first book is finished I have to manually start the second book.

It would be handy to have a feature where I could enqueue audiobooks, so that the next book automatically starts after the one before it in the queue, until the queue is empty, or until the sleep timer has finished.

pjanyboy-oss:
## Summary

When a podcast episode ends, the app currently stops — unless Android Auto is active. This PR extends the existing Android Auto auto-advance logic to regular (non-Auto) playback via a new user setting.

- Adds `autoPlayNextPodcastEpisode` toggle to Device Settings (Android only)
- When enabled: after an episode ends, the app syncs progress, finds the next unfinished episode, and starts it automatically
- Reuses the existing `handlePlaybackEnded()` / `getNextUnfinishedEpisode()` code path that already works for Android Auto
- Default: **off** (no behavior change for existing users)

## Files changed

| File | Change |
|------|--------|
| `DeviceClasses.kt` | Add `autoPlayNextPodcastEpisode: Boolean = false` to `DeviceSettings` |
| `PlayerNotificationService.kt` | Extend `handlePlaybackEnded()` guard: `isAndroidAuto \|\| autoPlayNextPodcastEpisode` |
| `settings.vue` | Toggle switch in Settings (Android only, `v-if="!isiOS"`) |
| `strings/en-us.json` | Label string `LabelAutoPlayNextPodcastEpisode` |

## Test plan

- [ ] Enable setting → play a podcast episode to completion → next episode starts automatically
- [ ] Disable setting → play a podcast episode to completion → playback stops (existing behavior)
- [ ] Android Auto with setting off → auto-advance still works (unchanged)
- [ ] Last episode in feed → no crash, playback stops normally

Closes #416


th3coop:
@pjanyboy-oss  Probably a silly question, I haven't pulled down the code yet; This will have zero effect for the "play next in series" idea for audiobooks, correct? ie. for this PR to actually close 416 more work would need to be done?  Not saying YOU should do more work btw, I'm looking for something to work on :P

PR #1883, opened 2026-05-27
cascade in-progress

rcketscientist:
## Brief summary

Android auto often loads directly into the library which reduces the safety of the app.

## Which issue is fixed?
Endpoints now cascade to ensure we load in-progress.

## Pull Request Type

android
## In-depth Description

Cacade queries
## How have you tested this?
Wasn't familar with capacitor, but managed to deploy.  This will be my daily driver for the next few days, I'll clear cache each morning and update if in-progress load reliably.


rcketscientist:
Been usiing this for over a week.  Works exactly as expected.  Both the full app and hte compact views prioritize the most recent audiobook, and auto-resume works perfectly.

PR #1843, opened 2026-04-07
feat: add Android TV support with D-pad navigation

bilbospocketses:
**Closes #606**

## Summary

Adds full Android TV support to the audiobookshelf app: leanback launcher integration, spatial D-pad navigation, focus ring styling, focus memory with fingerprint-based restore, overlay trapping, audio player TV controls, author detail page, and a TV user guide. All TV behavior is gated behind device detection — mobile/tablet behavior is unchanged.

Combines what was originally two PRs (basic TV scaffolding + D-pad navigation) into one since the scaffolding has no standalone value.

## For testers

User guide: [`docs/TV_USER_GUIDE.md`](https://github.com/bilbospocketses/audiobookshelf-app/blob/android-tv-dpad-navigation/docs/TV_USER_GUIDE.md). Sideload APK: [latest release](https://github.com/bilbospocketses/audiobookshelf-app/releases/latest).

## Commits

1. `feat: add Android TV support with D-pad navigation` — all feature code
2. `chore: normalize line endings to LF` — cosmetic-only pass across the repo since `core.autocrlf` differed between contributor machines. Kept separate from the feature commit for review clarity.

## Note to maintainer

Shared components (cards, toolbars, modals, player) have `tabindex` + `keydown.enter.prevent` added to enable D-pad activation. These attributes are inert on touch devices — they don't affect tap, layout, or mobile behavior.

PR #1893, opened 2026-06-06
Android TV (2/10): keyboard hygiene + foundational TV DOM markup

bilbospocketses:

### Scope
`tabindex` + `@keydown.enter.prevent` across ~25 shared components/pages so the existing UI is keyboard/D-pad navigable, plus inert TV DOM hooks (`data-tv-target` / `data-tv-overlay` + scroll-container `id`s) that the navigation engine reads in a later PR.

### Safety
Every change is inert on touch/pointer devices; phone/tablet behavior is byte-identical.

### Testing
Phone: tap regression across modified elements (unchanged). TV: Enter activates without double-firing.

Independent and reviewable on its own. The inert DOM hooks here are what the later navigation-engine PR consumes; nothing reads them yet.

### DIFF
--- a/components/app/Appbar.vue
+++ b/components/app/Appbar.vue
@@ -1,10 +1,10 @@
 <template>
   <div class="w-full h-16 bg-primary relative z-20">
     <div id="appbar" class="absolute top-0 left-0 w-full h-full flex items-center px-2">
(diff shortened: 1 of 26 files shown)


bilbospocketses:

PR #1892, opened 2026-06-06
Android TV (1/10): foundation + TV detection

bilbospocketses:
Part of the Android TV support work, replacing #1843 — that one was closed in favor of splitting it into smaller, focused PRs per the review-volume feedback.

### Scope
Detection + launcher entry only — **no navigation behavior yet**. Adds the leanback manifest entries + TV banner, `DeviceManager.isAndroidTV()`, an `isAndroidTv` Vuex flag, the `android-tv` `<html>` class injection, TV `onStop` termination, and a cast-init guard for devices without full Play Services.

### Safety
Phone/tablet behavior is byte-identical — changes are TV-gated or purely additive.

### Testing
Sideloaded on a Google TV Streamer 4K and an Android phone: launcher entry + banner present, `isAndroidTv` correct per device, no phone-side changes.

Independent and reviewable on its own. It lays the `isAndroidTv` foundation the later TV PRs build on — the drawer gate, the navigation engine, and the audio behavior all read this flag.

PR #1894, opened 2026-06-06
Android TV (3/10): hide "Go to Web Client" on Android TV

bilbospocketses:

### Scope
Hide the "Go to Web Client" side-drawer item on Android TV (a TV has no browser, so the action dead-ends). ~19 LOC: an `isAndroidTv` computed plus gating the existing `items.push(...)`.

### Safety
Gated on `isAndroidTv`; phone/tablet unchanged (item still shows). Inert until the `isAndroidTv` flag from #1892 merges.

### Testing
GTV Streamer 4K: item hidden. Phone: item present and opens the web client as before.

Depends on the `isAndroidTv` flag from #1892. Also edits `components/app/SideDrawer.vue`, which #1893 touches in different regions — whichever merges second may need a trivial rebase.

### DIFF
--- a/components/app/SideDrawer.vue
+++ b/components/app/SideDrawer.vue
@@ -81,6 +81,9 @@ export default {
     userIsAdminOrUp() {
       return this.$store.getters['user/getIsAdminOrUp']
     },
+    isAndroidTv() {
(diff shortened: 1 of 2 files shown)


bilbospocketses:

PR #1898, opened 2026-06-12
[Android] Skip already-downloaded files and resume interrupted downloads on restart

### LINKED ISSUE #215: Downloads to temp folder continue after app crashes; orphaned downloads persist after restart
### Steps to reproduce

1. Press download on a book in android app
2. Cause app to crash (not sure how to consistently replicate)
3. Watch downloads continue to temp directory (e.g. /Android/data/com.audiobokshelf.app/files/Download)
4. Restart app and temporary files remain in temporary directory. They don't get deleted, nor moved to final directory for listening.

### Expected behaviour
- Files would be recognized as the triggered downloads, or else would be deleted.

### Actual behaviour
- Files linger unused and unremoved.

### Environment data
Android 9
Server 2.0.18
Downloading to internal storage
Audiobookshelf Version:

- [ X] Android App? 0.9.46-beta
- [ ] iOS App?

#### Android Issue
Android version: 9

Device model: LG G7 Thinq

#### iOS Issue
iOS Version:

iPhone model:

### LINKED ISSUE #1722: [Enhancement]: Resume audiobook download after Connection issues

Dear Dev,
I know you can't reproduce https://github.com/advplyr/audiobookshelf-app/issues/1024

but I have a nice Idea for a useful feature that would be also a workaround for users like me that are facing this issue too.

Please implement the ability to resume downloads after the connection dropped. 
Right now any download starts at 0% percent even If you already have transfered 50% of the files. If it's not a sigle file audiobook it is not necessary to start the download at the beginning.

Thanks for your work!

-Quality of life improvement 
https://github.com/advplyr/audiobookshelf-app/issues/1024
nice Idea for a useful feature that would be also a workaround for users like me that are facing this issue too.

Just a resume button. At the download page.
Check for existing files
only download missing files

Android App - 0.10.1

_No response_

ifoundthetao:
## Problem

When a multi-track audiobook download is interrupted (force-close, crash, or OS kill), the app has no way to resume. On restart the user must tap Download again, which re-downloads every track from scratch — including ones that already completed.

Fixes #215, fixes #1722.

## What this PR does

Two cooperating changes, both Android-only:

**1. Skip already-downloaded files (`DownloadItemManager`)**

`processDownloadItemParts()` now calls `isFileAlreadyDownloaded()` before starting each part. If the file already exists at its final destination with the expected size, the part is marked complete/moved and the queue moves on. This fires on both the restore path and on any normal download batch, so a file that finished before the interruption is never re-fetched.

**2. Restore the download queue on startup (`DownloadItemManager` + `AbsDownloader`)**

`restoreDownloadQueue()` (called once from `AbsDownloader.load()`) reads persisted `DownloadItem`s from Paper DB via the existing-but-previously-uncalled `getDownloadItems()`, resets transient per-part state, and re-queues everything. The skip check then detects already-finished parts automatically.

**3. Fix `android.net.Uri` fields not surviving Kryo round-trips (`DownloadItemPart`)**

Paper DB uses Kryo for serialization. `android.net.Uri` fields deserialize as `null` after a round-trip (Kryo bypasses constructors via `sun.misc.Unsafe`, so `@JsonIgnore` and non-null declarations are not respected). Three Uri fields were affected:

- `finalDestinationUri` — converted to a computed property derived from the persisted `finalDestinationPath` string
- `serverUrl` — was computed from `uri` (null on restore); now computed from new `serverAddress`/`serverToken` String fields stored at creation time
- `uri` / `destinationUri` — still null on restore, but only used in `getDownloadRequest()` for external-storage downloads, which is out of scope for this PR

**4. Re-emit download items when JS listener registers (`AbsDownloader`)**

`restoreDownloadQueue()` runs during `load()`, before the WebView has loaded and JS listeners have registered. The initial `onDownloadItem` events are therefore dropped. The `addListener` override re-emits all in-progress queue items when the JS layer registers for `onDownloadItem`, so the download UI reflects restored downloads correctly.

## Scope

- **In scope:** Android internal-storage downloads (the common case), multi-track books.
- **Out of scope:** iOS — iOS uses `URLSessionConfiguration.background`, which survives app termination at the OS level and does not have the same restart problem. External-storage (SAF) restore is also not addressed; `uri`/`destinationUri` remain null on restore for that path.

## Testing

Tested on a physical Android device against a real ABS server (13-track audiobook):

1. Started download, force-closed after 3 tracks completed.
2. On restart: `Restoring 1 persisted download item(s) from disk` in logcat, then `Skipping already-downloaded file` for each completed track.
3. Remaining 10 tracks resumed and completed.
4. Book appeared as fully downloaded in the library.
5. Tapping Download on an already-downloading item correctly shows "Download already started" rather than re-queuing.

Logcat snippet (restart run):
```
I DownloadItemManager: Restoring 1 persisted download item(s) from disk
I DownloadItemManager: Skipping already-downloaded file: Arkady & Boris Strugatsky - Roadside Picnic 01-13.mp3
I DownloadItemManager: Skipping already-downloaded file: Arkady & Boris Strugatsky - Roadside Picnic 02-13.mp3
I DownloadItemManager: Skipping already-downloaded file: Arkady & Boris Strugatsky - Roadside Picnic 03-13.mp3
I DownloadItemManager: Resuming 1 restored download item(s)
D DownloadItemManager: Start internal download to destination path .../Roadside Picnic 04-13.mp3
D DownloadItemManager: Start internal download to destination path .../Roadside Picnic 05-13.mp3
D DownloadItemManager: Start internal download to destination path .../Roadside Picnic 06-13.mp3
```

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
@@ -75,6 +75,60 @@ class DownloadItemManager(
     checkUpdateDownloadQueue()
   }
 
+  /**
+   * Restores any download items that were persisted to disk and resumes them. Called once on app
+   * startup. Parts whose files already exist at their final destination are skipped by the queue's
+   * existing file-existence check; the rest are re-downloaded.
+   */
+  fun restoreDownloadQueue() {
+    val savedItems = DeviceManager.dbManager.getDownloadItems()
+    if (savedItems.isEmpty()) return
+
+    Log.i(tag, "Restoring ${savedItems.size} persisted download item(s) from disk")
+
+    var restoredCount = 0
+    savedItems.forEach { downloadItem ->
+      // Skip anything already in the in-memory queue
+      if (downloadItemQueue.any { it.id == downloadItem.id }) return@forEach
+
+      // Reset transient per-part state so the queue re-evaluates every part. getNextDownloadItemParts()
+      // only picks parts with a null downloadId, so a part that was mid-download when the app died must
+      // be reset or it would never resume. Already-finished parts are re-detected by the file-existence
+      // check in processDownloadItemParts().
+      downloadItem.downloadItemParts.forEach { part ->
+        part.downloadId = null
+        part.completed = false
+        part.moved = false
+        part.isMoving = false
+        part.failed = false
+        part.progress = 0
+        part.bytesDownloaded = 0
+
+        // Remove any stale partial temp file from an interrupted external download. The normal build
+        // path deletes these before queueing; the restore path bypasses it, and DownloadManager fails
+        // if the destination already exists.
+        if (!part.isInternalStorage) {
+          part.destinationUri.path?.let { tempPath ->
+            val tempFile = File(tempPath)
+            if (tempFile.exists()) {
+              Log.d(tag, "Removing stale temp file on restore: $tempPath")
+              tempFile.delete()
+            }
+          }
+        }
+      }
+
+      downloadItemQueue.add(downloadItem)
+      clientEventEmitter.onDownloadItem(downloadItem)
+      restoredCount++
+    }
+
+    if (restoredCount > 0) {
+      Log.i(tag, "Resuming $restoredCount restored download item(s)")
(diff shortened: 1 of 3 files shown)


advplyr:

PR #1907, opened 2026-06-26
fix(android): optimize internal download speeds and prevent silent stalling

jbcpollak:
## Description
Fixes an issue where Android internal storage downloads could stall indefinitely without throwing an error. This is done by adding proper OkHttp timeouts and try/catch exception handling to the async callback. 

Also improves download speeds on Android by:

1. Increasing the internal `FileOutputStream` write chunk size from 8KB to 64KB for optimal flash storage I/O.
2. Throttling the progress update events fired to the JavaScript UI layer to a maximum of once every 250ms, preventing the Capacitor bridge and UI thread from being flooded by thousands of updates per second.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/InternalDownloadManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/InternalDownloadManager.kt
@@ -18,7 +18,10 @@ class InternalDownloadManager(
 
   private val tag = "InternalDownloadManager"
   private val client: OkHttpClient =
-          OkHttpClient.Builder().connectTimeout(30, TimeUnit.SECONDS).build()
+          OkHttpClient.Builder()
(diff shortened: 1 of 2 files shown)


android/app/src/main/java/com/audiobookshelf/app/managers/InternalDownloadManager.kt
```diff
@@ -109,6 +127,6 @@ class BinaryFileWriter(
   }
 
   companion object {
-    private const val CHUNK_SIZE = 8192 // Increased chunk size for better performance
+    private const val CHUNK_SIZE = 65536 // Increased to 64KB for much faster I/O
```
jbcpollak:
```suggestion
    private const val CHUNK_SIZE = 65536 // Increased chunk size for better performance
```

PR #1909, opened 2026-06-26
fix: only discard credentials on genuine token rejection (401), not transient errors (404/403)

### LINKED ISSUE #1908: [Bug]: Auth token discarded if server reboots / is temporarily unavailable

Yes

### What was the Problem?

I have audiobookshelf behind traefik.
If I restart the audiobookshelf container, my android app logs me out and discards its auth token.
Once the server is back up I need to log-in again. 
This issue does not appear if I do not use the app while the server is restarting.

1. Reverse proxy returns 404 while the docker container is restarting
2. App discards auth token
3. User is forced to re-auth manually via password

### What was Expected?

Do not discard the auth token.
Inform the user that the server is unreachable and maybe display the http error code returned by the server.
Once the server is back, continue as if nothing happened.

### Phone Model

Pixel 10

### Phone OS

Android 16

Android App - 0.13.0

### Installation Source

Google Play Store

### Additional Notes

_No response_

Valentin-Metz:
### Brief summary
Differentiate a genuine token rejection (HTTP 401) from transient server
unavailability (404/403/5xx/network) during token refresh, so the app no
longer discards valid credentials and forces re-login when the server
reboots or a reverse proxy briefly returns a non-401 error.

### Which issue is fixed?
Fixes #1908

### Pull Request Type
- **Platforms:** Android and iOS, plus the WebView JS bridge (Capacitor/Nuxt).
- **Area:** Auth layer / token-refresh logic. No Nuxt page, iOS UI, or
  Android layout changes.

### In-depth Description
When the Audiobookshelf server reboots (or sits behind a reverse proxy
while the container restarts), API requests fail with 401 due to a stale
access token. The app then calls `/auth/refresh`, but if the server (or
proxy) isn't fully ready, that endpoint can return 403, 404, a 5xx, or a
network error.

Before this PR, **all four** refresh code paths treated **any non-200**
response from `/auth/refresh` as a permanent authentication failure and
deleted the stored refresh token / cleared the session. Users were
therefore forced to re-login on every server reboot even though their
credentials were still valid.

**The fix:** an HTTP **401** (the server explicitly rejecting the refresh
token) is the *only* status that triggers logout. All other outcomes —
403, 404, 5xx, and transport/network errors — are treated as transient:
the error is propagated to the caller, but the stored refresh token and
server config are left intact in secure storage, so the next request
succeeds once the server is back up.

This also repairs two pre-existing latent bugs in the genuine-401 logout
path: on both Android and iOS, `handleRefreshFailure` read the config id
*after* it had already nulled the in-memory server config, so the
`removeRefreshToken(...)` call was guarded by an id that was always `""`
/ `nil` and never actually removed the token. The config id is now
captured *before* the config is cleared. On iOS, `.validate(statusCode:
200..<300)` was added to the refresh request so non-2xx responses
reliably route into the `.failure` branch where the status is checked.

#### Files changed
- `android/.../server/ApiHandler.kt` — network error and non-2xx refresh
  responses now return a `{ transient: true }` error to the caller
  instead of calling `handleRefreshFailure`; only a 401 triggers logout.
  `handleRefreshFailure` captures the config id before nulling the config.
- `ios/App/Shared/util/ApiClient.swift` — added `.validate(statusCode:
  200..<300)`; in the `.failure` branch, only a 401 calls
  `handleRefreshFailure()`, others are logged as transient.
  `handleRefreshFailure()` removes the refresh token before clearing
  `Store.serverConfig`.
- `plugins/nativeHttp.js` — `refreshAccessToken` returns structured
  sentinels (`_permanentFailure` / `_transientFailure`) instead of bare
  `null`; thrown transport exceptions become `_transientFailure`.
  `handleTokenRefresh` only calls `handleRefreshFailure` when the error
  is not transient.
- `store/user.js` — `refreshToken` returns `null` for permanent failure
  (401 / missing token / malformed 200) and
  `{ _transientFailure: true, status }` for transient non-2xx.
- `plugins/axios.js` — distinguishes `null` (→ logout), `_transientFailure`
  (→ keep credentials), and a valid token (→ retry); in the `catch`
  block, only logs out when the refresh error is a genuine 401.

### How have you tested this?
1. **JS syntax/style:** `node --check` passes on `axios.js`,
   `nativeHttp.js`, and `store/user.js`; `npx prettier --check` passes
   on all three.
2. **WebView build:** `npx nuxt generate` succeeds with no errors
   introduced by these changes.
3. **Kotlin/Swift:** brace/import sanity checked; full Gradle/Xcode
   compile not run locally.
4. **Logic review:** traced the 401 → refresh chain across all four
   paths and confirmed that 401 is the only status that triggers logout
   on every platform, and that the transient branches preserve the
   stored refresh token and server config.
5. **Manual / reproducible steps:**
   - Run the ABS server in Docker (optionally behind a reverse proxy).
   - Log in on the app.
   - Restart the container (`docker restart abs-server`).
   - Immediately try to play an audiobook / load a library.
   - Before fix: app throws an error, deletes the refresh token, and
     redirects to the login screen.
   - After fix: the request fails transiently, the app stays logged in,
     and the next request succeeds once the server is back online.

### Screenshots
N/A — no UI changes.

### DIFF
--- a/.github/workflows/build-apk.yml
+++ b/.github/workflows/build-apk.yml
@@ -30,6 +30,9 @@ jobs:
       - name: install dependencies
         run: npm ci
 
+      - name: run unit tests
+        run: npm run test:unit
+
       - name: build Nuxt project
         run: npm run generate
 
--- a/android/app/src/main/java/com/audiobookshelf/app/server/ApiHandler.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/server/ApiHandler.kt
@@ -213,51 +213,62 @@ class ApiHandler(var ctx:Context) {
       client.newCall(refreshRequest).enqueue(object : Callback {
         override fun onFailure(call: Call, e: IOException) {
           Log.e(tag, "handleTokenRefresh: Failed to connect to refresh endpoint", e)
-          AbsLogger.error(tag, "handleTokenRefresh: Failed to connect to refresh endpoint for server ${DeviceManager.serverConnectionConfigString} (error: ${e.message})")
-          handleRefreshFailure(callback)
+          AbsLogger.error(tag, "handleTokenRefresh: Failed to connect to refresh endpoint for server ${DeviceManager.serverConnectionConfigString} (transient: ${e.message})")
+          // Network error is transient — keep credentials, return error to caller
+          val errorObj = JSObject()
+          errorObj.put("error", "Server temporarily unavailable (network error)")
+          errorObj.put("transient", true)
+          callback(errorObj)
         }
 
         override fun onResponse(call: Call, response: Response) {
           response.use {
-            if (!it.isSuccessful) {
-              AbsLogger.error(tag, "handleTokenRefresh: Refresh request failed with status ${it.code} for server ${DeviceManager.serverConnectionConfigString}")
+            if (it.code == 401) {
+              // Refresh token is genuinely invalid — log out
+              AbsLogger.error(tag, "handleTokenRefresh: Refresh request rejected (401) for server ${DeviceManager.serverConnectionConfigString}")
               handleRefreshFailure(callback)
-              return
-            }
+            } else if (!it.isSuccessful) {
+              // Any other non-2xx (403/404/5xx) is transient — server may be rebooting
+              AbsLogger.error(tag, "handleTokenRefresh: Refresh request failed with status ${it.code} for server ${DeviceManager.serverConnectionConfigString} (transient, keeping credentials)")
+              val errorObj = JSObject()
+              errorObj.put("error", "Server temporarily unavailable (status ${it.code})")
+              errorObj.put("transient", true)
+              callback(errorObj)
+            } else {
+              val bodyString = it.body!!.string()
+              try {
+                val responseJson = JSONObject(bodyString)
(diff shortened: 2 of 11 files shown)


Valentin-Metz:
I added a few tests, but since we currently have no test-suit, those are a bit blunt.

* **Native Android (`ApiHandler`) & iOS (`ApiClient`):** These are hardcoded against global singletons (`DeviceManager`, Alamofire's `AF`, `Store`, etc.). Testing them properly would require a refactor to support dependency injection first.
* **The `axios.js` `onError` orchestrator:** I focused on the leaf functions (`refreshToken` action and `refreshAccessToken`) that make the actual return decision. Fully testing the axios retry loop would require mocking Nuxt's interceptor lifecycle and `window.location`, which adds a lot of boilerplate.

I mocked the Capacitor bridge and made sure the two JS functions return the correct shapes (`null` for permanent rejection, `{ _transientFailure }` for network/proxy errors, and the token object on success). This required adding `vitest` to the project and wiring it into the `build-apk` workflow so tests run automatically on CI.

If the test toolchain and CI changes add too much scope creep, just merge the first commit and drop the rest. The main goal is getting the auth fix out, but the tests act as a quick regression guard for it.

Issue #1849, opened 2026-04-16
[Bug]: ABS Android App Stops Playing After Bluetooth Disconnect


NoxRaijin:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

ABS on Android seems to be breaking background usage as soon as I disconnect from a bluetooth device. Originally I thought it was an issue with media just not continuing after a certain period of time, however, I recently found an easy way for me to reproduce the issue.

Every time that I connect to my Android Auto head unit in the morning, I have to manually start the audio even though I have it set to auto-resume audiobookshelf when connected. Since I have a Samsung device, I utilize the "Modes and Routines" feature to auto-play audiobookshelf when I connect to my Galaxy Buds. I tested this with YouTube Music on both my Android Auto head unit and buds which seemed to work flawlessly. 

I attempted to troubleshoot this by clearing all the app data and cache for the app, turning on "unrestricted" in background usage, signing out of the app, uninstalling the app completely and performing a fresh reboot before reinstalling the app. 

### Steps to Reproduce the Issue

1. Connect to a bluetooth device
2. Audio does not resume from last played media

### What was Expected?

Connecting to bluetooth should autoplay on Android Auto. Connecting to bluetooth earbuds should either autoplay when connected (using a routine) or I should at least be able to tap my bud to play the last known audio (which would be audiobook in this case).

### Phone Model

Samsung Galaxy S26+

### Phone OS

Android 16 (OneUI 8.5)

### Audiobookshelf App Version

Android App - 0.11.0

### Installation Source

Google Play Store

### Additional Notes

One thing that I can say at least is that I am hosting my ABS with Docker. I have it set up with Traefik (custom domain) and I am using Authentik with OpenID Connect Authentication. I'm not certain if this could be causing an issue but I can share parts of my configuration with Traefik if that might help.

I should also note that this happens whether I am streaming the audio or have it downloaded. I almost always download my audio before playing.

NoxRaijin:
This is still not working properly.

epipenRX:
Hi -- the root cause here is that some car head units (and certain BT disconnect events) send `KEYCODE_MEDIA_STOP` to the Android media session, and the current handler calls `closePlayback()` which destroys the session entirely rather than pausing. That is why auto-resume routines see no active session to resume.

[...]

Issue #1476, opened 2025-02-11
[Bug]: Player closes on disconnect from car


Protarios:
### What was the Problem?

Since switching to an Samsung Galaxy S25 Ultra I habe the problem, that everytime I lock the car I am using the app, it does not only pause playback, but closes the player.
The annoying part is that to start playing again the phone needs to be onlocked and the play button needs to be pushed on screen, what is a mess while driving.
Is there any way to stop this behaviour and just keep the player open. Ideally also have it reopened with the last audiobook when the audiobookshelf app is started.

### Steps to Reproduce the Issue

1. Play Audiobook connected to car (BMW) by bluetooth
2. Lock the car via key (Bluetooth disconnecting)
3. The app stays active but the player inside the app is closed.

### What was Expected?

The app pauses playback, but the player stays active.

### Phone Model

Samsung Galaxy S25 Ultra

### Phone OS

Android 15, One UI 7.0

### Audiobookshelf App Version

Android App - 0.9.79

### Installation Source

Google Play Store

Protarios:
Duplicate of already exisiting issue 801

epipenRX:
Hi -- I hit this exact issue with my setup (Toyota Entune, not Android Auto). The root cause is that some car head units send `KEYCODE_MEDIA_STOP` when powering off or disconnecting, and the current code handles that by calling `closePlayback()`, which tears the session down entirely rather than just pausing.

I submitted a fix in PR #1913 that changes the handler to `pause()` instead. With the fix, locking/powering off the car pauses cleanly and leaves the book in the notification shade so it can resume from there. Hopefully this addresses what you were seeing too.

Issue #801, opened 2023-07-22
Android: Open audiobook player session doesn't stay open after disconnecting from car bluetooth


tzw9373:
### Steps to reproduce

1. Play audiobook (either local or stream)
2. Connect to 2016 Toyota Corolla via Bluetooth and resume audiobook 
3. Pause audiobook and turn off vehicle 
4. Return to vehicle and start, reconnect via Bluetooth

### Expected behaviour
- I would expect the open player session to still be there, app battery usage setting is set to Unrestricted 

### Actual behaviour
- Open player session is closed and must be restarted

### Environment data

Audiobookshelf Version: 2.3.2

- [X] Android App: 0.9.65
- [ ] iOS App?

#### Android Issue
Android version: 13

Device model: Google Pixel 6

Stock or customized system: stock

#### iOS Issue
iOS Version:

iPhone model:

advplyr:
I also have a Pixel 6 using Android 13. When I listen via bluebooth in my car it will automatically pause when I turn off the car. Then when I turn the car back on and bluetooth connects the audiobook automatically starts playing again.

I think this has been discussed before and may have to do with the vehicle/device you are connecting to, I'm not sure. If you get a chance to test it out in another vehicle let me know.

Also, you could try making sure there are no other media apps open in the background for testing.

tzw9373:
I tested your recommendation and the issue persists. It appears to be limited to this vehicle (doesn't occur with any other Bluetooth devices), but only for this app. Other apps exhibit the expected behavior.

awilisch:
I have a 2020 Toyota Highlander and I believe I'm experiencing the same behavior. 
I get in the car and pull the app up in Android Auto to play. When I return it will pull up Audiobookshelf as what I was listening to, but I have to scroll through the list to pick my audiobook to start playing again. 
I am trying to test what causes this to happen. I believe this might be when I continue listening on my bluetooth headset while I'm out of the car. I'll update with more details when I have more data.

imcompiling:
I have this exact same issue with a 2015 Toyota Corolla. I know that this is an app issue as Spotify and Audible do not have this issue.

Protarios:
Have the same problem. Opened a bug report that is a duplicate of this issue unfortunately having not seen this topic.

tzw9373:
I know it's been a while, but I'm still seeing this behavior (now also with a 2016 Toyota Sienna). I'm happy to provide whatever info possible to help get this one squashed!

[...]

Issue #612, opened 2023-02-27
[Android] Pause on Bluetooth Disconnect from car


i-am-at0m:
### Is your feature request related to a problem? Please describe.
I often get in and out of my car, and in other apps playback pauses when Bluetooth disconnects. It does not do so on Audiobookshelf so I skipped an hour in my book while my phone was in my pocket playing on minimum volume because it kept playing after disconnecting. 

### Describe the solution you'd like
Pause on disconnect. Possibly play on connect if it's the active audio app but that's secondary IMO. 

### Describe alternatives you've considered
N/A

### Additional context
I am on a Pixel 7 Pro and using a Pixel Watch, so if this feature exists but only pauses when NO Bluetooth devices are connected, that could be a cause. But my watch isn't a playback device so that doesn't make much sense. Server version is v2.2.15, android app 0.9.62-beta.

i-am-at0m:
Maybe this should be labeled a bug. I don't know what the intended behavior is.

hobesman:
Mine pauses when Bluetooth disconnects, but I usually only have one Bluetooth device connected so I can't address that scenario with your watch.

advplyr:
I just re-opened a similar issue #156 but I'm not sure if this is the same situation.

i-am-at0m:
Looks like it, I'm just experiencing it with a car instead of headphones. I searched but was only looking for open issues and didn't see that one.

advplyr:
Yeah I just re-opened it. That issue is also having to do with car bluetooth.
This may be specific to certain cars because it does pause in mine.

i-am-at0m:
Yeah this is really killing usability for me. I caught it twice today but didn't catch it and lost another hour of my book, which isn't chaptered, so trying to find where I left off was rough.

Is there an option for me to turn on debug logs and do some controlled testing? Or use Android developer mode to do an info dump or something?

advplyr:
Can you test this while only connected to one bluetooth device (your car)?

advplyr:
Interesting the two reports of this issue are from Volkswagen bluetooth

i-am-at0m:
> Can you test this while only connected to one bluetooth device (your car)?

Just tried it, put the watch in airplane mode, verified the phone only had one Bluetooth device connected, turned the car off and opened the door to power down the radio, and it didn't miss a beat and continued playing on my phone. 

> Interesting the two reports of this issue are from Volkswagen bluetooth

[...]

PR #1913, opened 2026-06-30
Treat KEYCODE_MEDIA_STOP as pause instead of closing the playback session

epipenRX:
## Summary

- When a car head unit powers off, it sends `KEYCODE_MEDIA_STOP` to the Android media session. The current handler calls `closePlayback()`, which tears down the session entirely, removing the media notification from the shade and requiring the user to manually reopen the app and restart the book.
- Standard media apps (YouTube, Plex, Jellyfin) treat `KEYCODE_MEDIA_STOP` as a pause, keeping the session and notification alive so playback can resume from the notification shade or when the car reconnects.
- This fix changes the handler in `MediaSessionCallback.kt` to call `pause()` instead of `closePlayback()`.

## Testing

- Tested on Toyota Entune (legacy Bluetooth audio system, not Android Auto).
- Toyota Entune sends `KEYCODE_MEDIA_STOP` on power-off and `KEYCODE_MEDIA_PLAY` on power-on. With this fix, powering off pauses and leaves the book in the notification shade; powering the car back on resumes automatically.
- Not tested with Android Auto. Android Auto should not be negatively affected by this change. If the same issue exists on Android Auto, this fix would likely address it there as well, though we have no way to confirm at this time.

## Related Issues

This fix addresses a root cause reported across several open issues:

- #204 (open since 2022) -- Playback closes when connecting to car Bluetooth (Jabra + Samsung Note 20 Ultra)
- #612 (open since 2022) -- Pause on Bluetooth disconnect from car (Pixel 7 Pro + Pixel Watch)
- #801 (open since 2023) -- Open audiobook player session doesn't stay open after disconnecting from car Bluetooth (Pixel 6 + Toyota Corolla 2016)
- #1476 (closed as duplicate, 2024) -- Player closes on disconnect from car (Galaxy S25 Ultra + BMW)
- #1849 (open since 2024) -- ABS Android app stops playing after Bluetooth disconnect (Samsung + Android Auto)

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/player/MediaSessionCallback.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/player/MediaSessionCallback.kt
@@ -234,7 +234,7 @@ class MediaSessionCallback(var playerNotificationService:PlayerNotificationServi
             playerNotificationService.jumpBackward()
           }
           KeyEvent.KEYCODE_MEDIA_STOP -> {
-            playerNotificationService.closePlayback()
+            playerNotificationService.pause()
           }
           else -> {
             Log.d(tag, "KeyCode:${keyEvent.keyCode}")


advplyr:
Thanks!

PR #1914, opened 2026-06-30
feat(android): auto-resume on BT headset after car disconnect

epipenRX:
The goal here was straightforward: automatically resume playback on the headset once it reconnects after the car powers off. Building that surfaced a prerequisite bug, though: without a live foreground service there's nothing to resume into. PlayerNotificationListener wasn't handling the non-ongoing notification state correctly, so Android 12+ would eventually kill the foreground service and drop the session entirely after a Bluetooth disconnect. That had to be fixed first, so it's included in this PR.

This builds on #1913 (merged), which handled the other half of the same shutdown problem: KEYCODE_MEDIA_STOP was calling closePlayback() instead of pause(). Between the two, a car power-off no longer kills the session, and now it can resume automatically once the original headset reconnects.

How it works:
- When the car sends KEYCODE_MEDIA_STOP (power-off), the app records which BT device was active and pauses instead of ending the session
- A 15-second window stays open for the BT device to reconnect
- If the same device reconnects within the window, playback resumes automatically
- If the window expires the state clears, so reconnecting later does not trigger an unexpected resume
- Controlled by a toggle in device settings, on by default

Notes:
- Only fires for the exact device that was active when the car powered off, checked by product name
- previousBtDevice is marked @Volatile so the background timer thread and the AudioDeviceCallback thread see a consistent value
- No new permissions required

Tested: start playback on Bluetooth headphones, get in the car (audio transfers to the car's Bluetooth as expected), turn the car off. Playback resumes automatically on the headphones with no phone interaction, as long as they're still powered on and connected.


epipenRX:
Update: rebased this onto current master to resolve the merge conflicts from the branch drifting behind. Also extended the Bluetooth reconnect window from 3 to 15 seconds, to give headphones more time to kick back on after the car disconnects. Description above is updated to match.

Issue #407, opened 2022-10-08
audio equalizer in app


yegods:
### Is your feature request related to a problem? Please describe.
Audio levels on all books are not consistent.  It is a nice feature on some other book players to have an equalizer.  I use it to reduce the high end and increase the low end in loud environments.  

### Describe the solution you'd like
A simple 5 band equalizer.  I will attach an example pic from Smart Audiobook Player on android.

### Describe alternatives you've considered
Don't really have any, but raising the volume, which can hurt the ears especially in the high end.

### Additional context
![equalizer](https://user-images.githubusercontent.com/114703516/194728235-3fd7a668-4386-4384-afe4-926abcf4b0da.jpg)

iconoclasthero:
+1  I've also found this to be a useful feature of SABP because I use very small .opus files (17 kb/s) with an additional ffmpeg filter that strips out things that are supposed to be inaudible.  As a result the files tend to be a bit flat and I've been using the eq in SABP to make it sound a bit better.  I don't use the loudness feature.  

EDIT: Removed redundant screenshot.

robdennehy:
+1 for this as well. Sometimes a narrator's voice is just too bassy and I need to play with my car's EQ or the headphone's to stop it booming so much but then I have to fix both when I want to listen to anything else. Would be great to have this apply just to the audiobooks, or even better, a specific book.

linuxcommander:
Equalizer. Remove silence. At least five band equalizer. Bonus features, each audiobook can have its own EQ settings.

iconoclasthero:
Hi.  I was trying to listen to _All the Pretty Horses_ and the narration has too much low end and I needed to play it through SABP to use the eq so it would be comprehensible with my headphones.  The offline copy into a shared folder makes this easy, but it's a poor work around.  This isn't something I do often and it wasn't a feature I used _a lot_ in SABP, but it is sometimes _necessary._

Alternatives include system-wide DSP like JamesDSP for Android and ViperDSP.  

https://play.google.com/store/apps/details?id=me.timschneeberger.rootlessjamesdsp&hl=en_US
https://xdaforums.com/t/jamesdsp-a-working-alternative-to-viper-on-android-11.4179957/

I don't know what it would take to get something like this integrated into the android app, but I imagine there should be some off-the-shelf android equalizers that could be put into ABS/Android.  I'm interested in finding a like-minded person or two to work on this aspect of the mobile app.

KiwiHour:
[...]

PR #1915, opened 2026-07-01
Android: add independent pitch adjustment control

epipenRX:
Adds a pitch adjustment control to the playback speed modal on Android.

Users can now raise or lower the narrator's pitch independently of playback speed. Speed correction is already applied by default in ExoPlayer, so this control allows fine-tuning pitch for personal preference or hearing needs without affecting the pace of playback.

Details:
- Range: 0.5x to 2.0x in 0.05 steps
- Persists locally between sessions (stored in device settings, not synced to server, pitch is a per-device hearing preference)
- Displayed in the existing playback speed modal below the speed control, using the same compact layout
- Input is validated and clamped on the native side in addition to the UI controls
- Android only; Cast playback is unaffected
- No new permissions required

Relates to #407 (audio equalizer request). Pitch control is a simpler focused complement to a full EQ for users who primarily need to shift the narrator's register.

Addresses the use case raised in #5189. Users with hearing sensitivity in certain frequency ranges can now shift pitch up or down to improve clarity without changing playback speed.


epipenRX:
Update: rebased this onto current master to resolve the merge conflicts from the branch drifting behind.

PR #1920, opened 2026-07-05
fix(android-auto): run server connection check off the main thread to prevent ANR

### LINKED ISSUE #1919: [Bug]: Android Auto browsing causes ANR/crash when webview server connection isn't established (main-thread network in checkSetValidServerConnectionConfig)
### I have verified that the bug is not already awaiting release

Yes

### What was the Problem?

When the phone is connected to Android Auto, the app repeatedly crashes: on the phone it simply "closes", and on the Android Auto head unit a generic "failed to connect" error appears. The standalone app (playing over Bluetooth, driven by the webview) works flawlessly — only the Android Auto browse path is affected.

**Root cause (traced through the source):**

Android Auto browsing enters `PlayerNotificationService.onLoadChildren`, which runs on the media browser service **main thread**. For the `AUTO_MEDIA_ROOT` node it calls `MediaManager.loadAndroidAutoItems` → `checkSetValidServerConnectionConfig`. That function used `runBlocking { ... }` and, inside it, **pinged every saved server config and ran `authorize` synchronously** (`checkServerConnection`/`authorize` are suspend network calls). Running that under `runBlocking` on the main thread blocks the main thread on network I/O.

This is made much worse by a reload storm:
- Two clients poll the browse root every ~2s: Android Auto (`com.google.android.projection.gearhead`) and Google Assistant (`com.google.android.googlequicksearchbox`).
- While the Capacitor webview hasn't yet established the server connection, `DeviceManager.serverConnectionConfig` is null, so every `onGetRoot` sees "No server connection" and flags `forceReloadingAndroidAuto = true`, which forces `loadAndroidAutoItems` to re-run on the next `onLoadChildren`.
- So the blocking network work is triggered repeatedly and re-entrantly → **ANR** → the process is killed (phone app "closes", Auto shows the generic connection error).

Log excerpt (server address redacted) showing the ~2s reload storm with the connection state flapping between the server and "No server connection", before the webview has even mounted:

```
10:48:37.163 onGetRoot: clientPackageName: com.google.android.projection.gearhead
10:48:37.165 onGetRoot: Reset Android Auto server items cache (https://[REDACTED])
10:48:39.437 onGetRoot: clientPackageName: com.google.android.projection.gearhead
10:48:39.439 onGetRoot: Reset Android Auto server items cache (No server connection)
10:48:41.763 onGetRoot: ... Reset Android Auto server items cache (No server connection)
10:48:44.024 onGetRoot: ... Reset Android Auto server items cache (No server connection)
...
10:52:24.423 mounted: initializing first load (android v0.13.0-beta)   <-- webview only mounts here
```

A secondary latent bug: in `MediaManager.checkResetServerItems`, the server config id is compared with `!==` (reference identity) instead of `!=` (value), which can cause spurious cache resets that feed the reload storm.

1. Have the app configured to connect to a remote ABS server.
2. Force-stop the app so the Capacitor webview is not running.
3. Connect the phone to Android Auto (or launch Android Auto so it cold-binds the media browser service before the webview establishes the server connection).
4. Android Auto attempts to browse the root repeatedly (~2s), the app hangs on main-thread network I/O and the process is killed. Auto shows a generic connection error; the phone app "closes".

Most reliable when the server ping/authorize is even slightly slow (e.g. around a token refresh), which pushes the blocked main thread over the ANR threshold.

### What was Expected?

Android Auto should be able to browse without the app hanging/crashing. The server connectivity check must not run network I/O on the media browser service main thread, and repeated browse-root requests should not spawn overlapping blocking loads.

### Phone Model

Pixel 9 Pro

### Phone OS

Android 17 (stock)

Android App - 0.13.0

### Installation Source

Google Play Store

### Additional Notes

Workaround that confirms the diagnosis: disabling the Android Auto app, restarting Android Auto, and playing the audiobook through the standalone app over Bluetooth avoids the crash entirely — because playback driven by the webview never enters the `checkSetValidServerConnectionConfig` main-thread path.

I have a fix ready and will open a PR referencing this issue: move `checkSetValidServerConnectionConfig` off the main thread onto `Dispatchers.IO`, coalesce overlapping `loadAndroidAutoItems` calls behind a single in-flight guard, and fix the `!==` → `!=` comparison. Verified on device: builds and resolves the crash 100%.

databoy2k:
## Brief summary

Fixes Android Auto browsing repeatedly crashing the app (phone app "closes", Android Auto shows a generic "failed to connect" error). The server connectivity check ran network I/O on the media browser service main thread via `runBlocking`, causing an ANR that was compounded by the ~2s Android Auto browse-root polling.

## Which issue is fixed?

Fixes #1919

## Pull Request Type

Android only, native (Kotlin) backend of the app. No frontend changes.

## In-depth Description

`PlayerNotificationService.onLoadChildren` runs on the media browser service **main thread**. For the Android Auto root it calls `MediaManager.loadAndroidAutoItems` → `checkSetValidServerConnectionConfig`, which used `runBlocking { ... }` to ping every saved server config (`checkServerConnection`) and run `authorize` — all suspend network calls. Under `runBlocking` on the main thread, this blocks the main thread on network I/O.

It is made re-entrant by a reload storm: Android Auto (`gearhead`) and Assistant (`googlequicksearchbox`) each poll the browse root every ~2s. While the Capacitor webview hasn't established the server connection, `DeviceManager.serverConnectionConfig` is null, so every `onGetRoot` flags `forceReloadingAndroidAuto = true`, forcing another `loadAndroidAutoItems` on the next `onLoadChildren`. Repeated blocking loads → ANR → process killed.

Changes (all in `android/app/src/main/java/com/audiobookshelf/app/media/MediaManager.kt`):

1. **`checkSetValidServerConnectionConfig` now runs on a dedicated `Dispatchers.IO` scope** instead of `runBlocking`, so the pings/authorize never block the caller. The `(Boolean) -> Unit` callback contract is unchanged; downstream Android Auto callbacks already ran off the main thread (delivered on the OkHttp callback thread), so ordering/threading of the result is preserved.

2. **`loadAndroidAutoItems` coalesces overlapping calls** behind a single in-flight guard and fires all queued callbacks when the one load completes. This prevents the browse-root polling storm from spawning concurrent loads that race on the shared server state (`serverLibraries`, `serverUserMediaProgress`, etc.). The existing 5-second ping cache already bounds network cost; this bounds concurrency.

3. **Fix a latent bug in `checkResetServerItems`**: the server config id was compared with `!==` (reference identity) instead of `!=` (value), which could trigger spurious cache resets that fed the reload storm.

No public API or behavior changes outside the Android Auto browse path.

## How have you tested this?

Built from source in Android Studio and installed on a physical device (Pixel 9 Pro, Android 17, app 0.13.0). Before the change, connecting to Android Auto reliably crashed the app within seconds of browsing; after the change the crash is fully resolved and Android Auto browsing works normally, including the previously-failing cold-start case where Auto binds the media browser service before the webview establishes the server connection.

Repro used: force-stop the app (webview not running), then launch Android Auto so it cold-binds the browse service. Pre-fix: ANR/crash. Post-fix: browses normally.

## Screenshots

N/A — no front-end changes.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/media/MediaManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/media/MediaManager.kt
@@ -9,8 +9,11 @@ import com.audiobookshelf.app.device.DeviceManager
 import com.audiobookshelf.app.server.ApiHandler
 import com.getcapacitor.JSObject
 import java.util.*
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.coroutineScope
-import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.launch
 import org.json.JSONException
 import org.json.JSONObject
 import kotlin.coroutines.resume
@@ -45,6 +48,16 @@ class MediaManager(private var apiHandler: ApiHandler, var ctx: Context) {
   var serverLibraries = listOf<Library>()
 
   var userSettingsPlaybackRate:Float? = null
+
+  // Android Auto browses on the media browser service main thread, so the server pings
+  // and authorize call must not run there or they block it and trigger an ANR
+  private val androidAutoIOScope = CoroutineScope(Dispatchers.IO + SupervisorJob())
+
+  // Android Auto re-requests the browse root every ~2s and multiple clients do so at once,
+  // so overlapping loads are coalesced into one to avoid racing on the shared server state
+  private var isLoadingAndroidAutoItems = false
+  private val pendingAndroidAutoLoadCallbacks = mutableListOf<() -> Unit>()
+  private val androidAutoLoadLock = Any()
 
   fun getIsLibrary(id:String) : Boolean {
     return serverLibraries.find { it.id == id } != null
@@ -136,7 +149,7 @@ class MediaManager(private var apiHandler: ApiHandler, var ctx: Context) {
     //   and reset any server data already set
     val serverConnConfig = if (DeviceManager.isConnectedToServer) DeviceManager.serverConnectionConfig else DeviceManager.deviceData.getLastServerConnectionConfig()
 
-    if (!DeviceManager.isConnectedToServer || !DeviceManager.checkConnectivity(ctx) || serverConnConfig == null || serverConnConfig.id !== serverConfigIdUsed) {
+    if (!DeviceManager.isConnectedToServer || !DeviceManager.checkConnectivity(ctx) || serverConnConfig == null || serverConnConfig.id != serverConfigIdUsed) {
       podcastEpisodeLibraryItemMap = mutableMapOf()
       serverLibraries = listOf()
       serverLibraryItems = mutableListOf()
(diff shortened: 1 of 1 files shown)


advplyr:
Makes sense, thanks!

PR #1810, opened 2026-02-22
feat: auto-play next episode when playing podcast playlists

Pittermaennchen:
## Summary

- When playback ends while a playlist queue is active, automatically start the next episode instead of stopping
- Add `playlistQueue` state to Vuex store to track current playlist context (`playlistId`, `items`, `currentIndex`)
- Set queue context in `playlist/_id.vue` when the play button is pressed
- Pass playable items down to `ItemTableRow` so individual episode clicks also set up the queue for sequential playback
- Emit `playback-ended` event from `AudioPlayer` on `STATE_ENDED`
- Handle `playback-ended` in `AudioPlayerContainer` to advance the queue and trigger the next `play-item`; clear queue when playing outside a playlist

## Test plan

- [x] Open the Podcasts library and navigate to a playlist
- [x] Press the playlist play button and verify the first unfinished episode starts playing
- [x] Wait for the episode to end and verify the next episode in the playlist starts automatically
- [x] Verify playback stops after the last episode in the playlist
- [x] Click an individual episode in the playlist table and verify auto-play continues to the next episode
- [x] Play an item outside of a playlist and verify the queue is cleared (no unintended auto-play)
- [x] Test with locally downloaded episodes as well as streamed episodes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1923, opened 2026-07-07
feat: auto-play next episode when playing podcast playlists (rebased #1810, tested)

### LINKED ISSUE #416: Provide ability to queue/playlist audiobooks

bobtheskull-flameeyes:
This PR is a rebased, cleaned-up version of #1810 by @Pittermaennchen, which implements auto-play for podcast playlists (the feature requested in #416).

## Changes from #1810

- **Rebased onto current master** (`185cba1`) — no conflicts
- **Dropped the debug icon commit** (`e243477`) — the upside-down app icon in debug builds is an intentional visual distinction between debug and release builds, not a bug to fix
- **All other commits preserved** with original authorship intact

## What it does

When playing a podcast episode that's part of a playlist, the app now automatically advances to the next episode when the current one finishes. This works even when the app is backgrounded or the device enters Doze mode, using Kotlin-native WakeLock and network wake handling.

## Commits (5 of the original 6)

1. `feat: auto-play next episode when playing podcast playlists` — core JS-side playlist advancement
2. `fix: forward ENDED metadata event to JS when app is backgrounded` — ensures the event reaches the webview
3. `fix: advance podcast playlist natively to survive Android Doze mode` — Kotlin-side advancement for background/Doze
4. `fix: keep CPU/network awake during podcast playlist advancement` — WakeLock handling
5. `fix: robust background playlist advancement for podcast playlists` — edge-case hardening

## Testing

Built a debug APK from this branch and tested on a physical Android device — podcast playlists advance correctly through episodes, including when the app is backgrounded.

Closes #416

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -72,9 +72,13 @@ The bridge between JS and native is five custom Capacitor plugins. Each has a JS
 
 ### Playlist Queue
 
-Podcast playlist playback is coordinated entirely in the JS layer. `store/index.js` holds `playlistQueue: { playlistId, items, currentIndex }`. When an episode ends, `AudioPlayer.vue` receives the `ENDED` `onMetadata` event and emits `playback-ended` on `$eventBus`. `AudioPlayerContainer.vue` handles this by advancing `currentIndex` and emitting `play-item` for the next episode.
+Playlist playback uses a two-layer approach so advancement works even when the screen is off or Android's Doze mode throttles the WebView's JS engine:
 
-**Background playback gotcha**: `AbsAudioPlayer.kt` suppresses `onMetadata` events while the app is backgrounded (`isInForeground == false`) to avoid flooding the WebView. The `ENDED` state is explicitly exempted from this suppression so that playlist advancement works when the screen is off or the app is minimized. Do not remove this exemption.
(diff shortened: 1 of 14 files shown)


bobtheskull-flameeyes:
Friendly ping on this one — it's been open since July. I've just pushed one more commit to the branch: `fix: ensure Bluetooth disconnect pauses playback with playlist queue active`.

While testing the playlist-queue feature on a Samsung device I found that `PlayerNotificationListener.onNotificationPosted` re-asserted `startForeground()` whenever `playlistQueue` was non-empty — including right after `ACTION_AUDIO_BECOMING_NOISY` had paused the player on Bluetooth disconnect. On some devices (Samsung One UI, Android 12+) this fights the service-lifecycle demotion that should accompany the audio-route change, so disconnecting headphones didn't reliably pause.

The fix guards the foreground re-assertion (and the notification-dismissal / `onTaskRemoved` keep-alive) on `currentPlayer.playWhenReady` in addition to the queue being non-empty, so the service stays alive between episodes (`STATE_ENDED` with `playWhenReady=true`) but properly yields on pause. It also registers an explicit `ACTION_AUDIO_BECOMING_NOISY` receiver in `PlayerNotificationService` as defense-in-depth, so the pause sticks even if ExoPlayer's internal handler races the notification lifecycle.

Tested on-device with a debug build. Happy to split the fix into a separate PR if you'd rather review the feature and the fix independently.

Issue #1784, opened 2026-01-27
[Bug]: iOS - 0.11.0-beta - Downloads are very slow on iOS app


sallan2000:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

Downloading a 30 minute podcast file can take upwards of 10 minutes to download in 0.11.0-beta

Earlier versions download files very quickly

When doing so, the phone gets very hot.

### Steps to Reproduce the Issue

Go to library page
select any book or podcast episode
click download
go to the downloads page
wait

### What was Expected?

Files to download in a short period of time

### Phone Model

iPhone 16

### Phone OS

iOS 26.3 public beta

### Audiobookshelf App Version

iOS App - 0.11.0

### Installation Source

Testflight

nichwall:
Related/duplicate of https://github.com/advplyr/audiobookshelf-app/issues/1341

sallan2000:
> Related/duplicate of [#1341](https://github.com/advplyr/audiobookshelf-app/issues/1341)

I think it might be but the terrible speed of downloading to local storage is a decidedly different symptom.

WillGunn:
I have been experiencing the slow downloads issue for the past 2 months or so, there weren't any problems before that.  I wasn't sure if it was an app problem or my server until I downloaded a different ios app that supports connecting to the audiobookshelf server (called 'Still') and it's download speed was back to where this app was a few months ago.  For instance it was able to download a 1.3GB audiobook in about a minute, whereas the ABS app took 2 hours for the same book.

iPhone 16 Pro Max
iOS 18.7.2

ABS-app 0.11.0-beta
ABS Server v2.32.1

sallan2000:
> I have been experiencing the slow downloads issue for the past 2 months or so, there weren't any problems before that. I wasn't sure if it was an app problem or my server until I downloaded a different ios app that supports connecting to the audiobookshelf server (called 'Still') and it's download speed was back to where this app was a few months ago. For instance it was able to download a 1.3GB audiobook in about a minute, whereas the ABS app took 2 hours for the same book.
> 
> iPhone 16 Pro Max iOS 18.7.2
> 
> ABS-app 0.11.0-beta ABS Server v2.32.1

Hi Will,

Does your iPhone get very hot during downloading too?

Thanks!

WillGunn:
[...]

Issue #1569, opened 2025-05-07
[Bug]: ios Downloads bug


mirhgs:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

![Image](https://github.com/user-attachments/assets/608d6363-7e51-4ac1-b4a2-a1ee01d6f54b)

The problem I am having is that I can not download my audiobook using the audiobookshelf app, but it does work in third-party apps like Plappa. 

### Steps to Reproduce the Issue

Click on the download button. 

### What was Expected?

I should be able to the download the audiobook. 

### Phone Model

iphone 16 pro max 

### Phone OS

18.4.1

### Audiobookshelf App Version

iOS App - 0.9.81

### Installation Source

SideStore

advplyr:
Can you share the file format of the audiobook? Are you connected over a local ip address or through a reverse proxy? Have you tested other audiobooks that you have that are a different format or from a different source to eliminate the possibility that this is file specific?

mirhgs:
My whole library is m4b. I am connected over my local IP address. Tried many different books. I deleted the app and reinstalled it.

ucrbuffalo:
@advplyr I have reproduced the bug (unintentionally) by starting a download while on mobile data, then switching to WiFi while the download was happening. 

After the download “failed” and I reattempted, the download progress bar flashes back and forth from 0% to whatever percentage the download is at.

mirhgs:
I finally found the solution to my problem. You have to use AltStore to sideload the app.

p4xel:
> [@advplyr](https://github.com/advplyr) I have reproduced the bug (unintentionally) by starting a download while on mobile data, then switching to WiFi while the download was happening.
> 
> After the download “failed” and I reattempted, the download progress bar flashes back and forth from 0% to whatever percentage the download is at.
> 
> The book is not listed in my “downloads” and there’s isn’t a file available for me to delete in the Files app.

I think i'm facing the same issue. I started download on Mobile, and didn't finish it, and when I tried again on WiFi I am getting this same error. Did you find a solution to this? Reinstalling the app did not fix it

Choozey:
I'm having this issue on all of my devices. iPhone 13 Pro Max on iOS 18 and iPhone 14 Pro on iOS 26, v0.12.0 app version. I use Feather to sign the IPA with my certificate.
[...]

Issue #961, opened 2023-11-27
[Bug]: download complete but cant create library item (ios)


xanthos84:
### Describe the issue

downloading items does not work on last version
ios 0.9.67 / 0.9.68

### Steps to reproduce the issue

1. open app
2. download item for offline listening
3. items download progresses until 100%
4. error msg „download complete but cant create library item“

### Audiobookshelf version

2.5 server 0.9.67/0.9.68 ios

### How are you running audiobookshelf?

Docker

xanthos84:
> Are you able to open the browser and download successfully from the website?

yes that works fine for single files and for complete packages (zip).

xanthos84:
this is still an issue with the newest app on iOS + newest server version!

Jeor:
Having same issue

Server 2.8.1
ios (self signed) 0.9.73-beta

soberhofer:
Having the same issue. App is signed and installed using a valid dev certificate. How can I further troubleshoot this?

jerry00020001:
Also getting this error, only happens on my wife's phone, any book. Tried hostname, then IP. Tried deleting app and re-installing. Tried creating a second account, issue still persists. Tried restarting iPhone. re-scanning book. works fine on my phone.

clincher61:
Also getting this error. Client version 0.9.74.  Server version 2.9.0.  Every attempt to download a book yields "Item download complete but failed to create library item" error message on the client.  App is personally signed and sideloaded (not part of the test flight beta).  iOS 17.4.1.  

I suspect this may have something to do with personally signed apps as it sounds like everyone reporting above is not in the Test Flight beta.  ?

cofifield:
**TLDR: Use AltStore to sideload the app.**

I have found the root cause of this problem. It is occurring due to a missing sandbox entitlement which seems to be coming from how users are choosing to sideload their app. It is **not** an application bug/build/configuration issue. At this time I have tested Sideloadly and AltStore with sideloading via Altstore being the solution. Other sideloading methods may work but I have not tested them.

When building the application from source and deploying to a simulator or physical device, we can review the console logs while attempting to download a file from the server. A proper download looks like this:
[...]

PR #1924, opened 2026-07-09
Fix offline downloads: unreachable track URL, background-session throttling, retry/stall handling (iOS + Android)

### LINKED ISSUE #344: iOS downloads do not handle poor network conditions or partial failures gracefully
### Steps to reproduce

1. Download on a poor network connect
2. Have the download fail
3. Be forced to redownload the whole book

### Expected behaviour
- There should be some automatic retrying and partial recovery before out right failure

### Actual behaviour
- Complete failure if any parts fail to download

### Environment data

Audiobookshelf Version:

- [ ] Android App?
- [x] iOS App?

#### Android Issue
Android version:

Device model:

Stock or customized system:

#### iOS Issue
iOS Version:

iPhone model:

DrNgo:
## Brief summary

Makes offline downloads on iOS actually finish, and finish quickly. What began as hardening
against silent wedges and stalls turned up two much larger problems once the fixes were tested
on a real device against a large (multi-GB, 10-part) audiobook:

1. **Every track part persisted a download URL pointing at a host that does not exist**, so any
   download restarted from the database could never transfer a byte. Pre-existing, and the real
   reason large books never completed (§1).
2. **iOS throttles background `URLSession` transfers to ~0.6 MB/s aggregate**, so even a working
   download crawled. Downloads now run on an unthrottled session while the app is in the
   foreground — measured **~0.6 MB/s → 39-48 MB/s** on device (§2).

Also fixes the original "Processing…" hang, retry/stall handling that gave up too early, duplicate
concurrent transfers, and the Android issues raised in review.

## Which issue is fixed?

Fixes #344 (no retry / partial-failure recovery on poor networks), #1784 (iOS downloads very
slow / phone runs hot), and #1479 (download sometimes gets stuck on a single file — see the
backgrounding caveat below).

_Caveat on #1479:_ the core symptoms are fixed on both platforms — a stalled part no longer
holds a concurrency slot (stall watchdog cancels + retries), and progress is no longer lost on
restart (launch reconciliation). It does **not** address the separate Android complaint that a
download stalls specifically when the app *loses focus / is backgrounded* — that needs a
foreground download service, which this PR does not add.

**Partially addresses #961** (download complete but can't create library item) — deliberately
*not* auto-closing. This fixes the **network-dependent** completion failure mode (offline /
HTTP 429 / expired token — see §4 below). But many reports in #961 are a separate,
**environmental** iOS issue: when the app is sideloaded without the correct entitlements, the
sandbox denies the file move out of the `nsurlsessiond` cache (`deny(1) file-write-unlink` on
`CFNetworkDownload_*.tmp`). That's a signing/OS problem no app-code change can fix — install
via TestFlight / AltStore / LiveContainer works; Sideloadly / e-sign / Scarlet / Feather have
been reported to fail. (A full disk produces the same message too.) With this PR, those users
now get a clear failure toast instead of a silent hang, but the download still won't succeed
until the app is installed through a method that grants the entitlements. Details in
[this comment](https://github.com/advplyr/audiobookshelf-app/issues/961#issuecomment-4930204816).

**Partially addresses #1569** (iOS downloads bug) — also *not* auto-closing. The reports there
of a download failing after a mid-transfer network change (mobile → Wi-Fi) should be helped by
the stall watchdog + retry + `waitsForConnectivity`. However, the OP's own case was resolved by
switching to AltStore, i.e. the same **sideloading-entitlements** issue as #961 above (as was at
least one other report signing with Feather) — that part is not fixable in app code.

May also resolve (related; worth a maintainer's confirmation): #673 (download indicators vanish
after an app refresh — reconciliation re-emits them), and #215 (orphaned downloads persist after
a restart — DB-tracked orphans are now reconciled and resumed/cleaned on launch).

## Pull Request Type

Affects **both Android and iOS**, plus a small shared **frontend** change (failed-part
toast). Native download managers + one Vue widget.

### 1. Track parts stored an unreachable download URL (iOS) — the root cause of "large books never finish"

`startLibraryItemTrackDownload` passed `Store.serverConfig!.address` as a part's `serverPath`,
while ebook and cover parts passed a real API path. `DownloadItemPart`'s initializer builds the
stored URI as `"\(address)\(serverPath)?token=…"`, so for **every track part** that became
`address + address`:

```
https://abs.example.comhttps://abs.example.com?token=…   ->   host = "abs.example.comhttps"
```

A hostname that does not resolve. And because the download session sets
`waitsForConnectivity = true`, URLSession does not fail fast on an unreachable host — it *waits*.
So the transfer sat at **exactly zero bytes** until the stall watchdog cancelled it and retried
against the same dead URL, forever.

The first attempt at a track never used that stored URI (it used `urlForTrack`), which is why
fresh downloads worked and this went unnoticed. Only a restart **from the database** used it —
reconcile after a relaunch or a kill. A small book finishes on the first attempt and never touches
that path; a large one is inevitably interrupted, reconciled, and then permanently dead. This is
pre-existing (it predates this PR — `git log -L` traces it to the iOS offline-ebooks change) and
it explains the long tail of "download gets stuck on one file" reports.

Fixes: track parts now persist the real API path; download URLs are **rebuilt from the current
server config** every time a task is created from a persisted part, rather than trusting the
stored URI. That second part also fixes a latent bug in the same line — the URI baked in the
access token at queue time, and those are short-lived JWTs, so anything reconciled hours later
presented an expired one.

### 2. Background-session throttling (iOS) — downloads were ~60x slower than the link

Measured on device, same network, same moment:

| | throughput |
|---|---|
| background `URLSession` | ~0.6 MB/s **aggregate** (3 transfers ≈ 0.20 MB/s each; 6 transfers ≈ 0.10 MB/s each) |
| default `URLSession` (8 MB probe) | 21.58 MB/s |
| desktop browser, same server | 30-40 MB/s |

Two details make this conclusive rather than circumstantial: transfers advanced to **byte-identical
totals** at every sample (independent TCP connections never track each other — that is a scheduler
handing out fixed chunks), and doubling the concurrency exactly halved the per-transfer rate while
the aggregate stayed pinned. So the throttle is **per session, not per task**, and no amount of
concurrency tuning helps.

Fix: transfers run on an in-process default session while the app is in the foreground and are
migrated to the background session when it leaves, then back on return. Migration cancels for
resume data and reissues on the other session — which only preserves progress because the server
answers Range requests with `206`, verified before building it. Both sessions share one serial
delegate queue, so the `didFinishDownloadingTo`-before-`didCompleteWithError` ordering the
completion logic depends on still holds across a migration. A migration cancel consumes no retry
attempt, there is a 5 s cooldown so app-switching can't become a cancel/reissue storm, and
backgrounding takes a `UIBackgroundTask` assertion so the round trip isn't cut off.

Measured after the change: **39-48 MB/s aggregate**, with 300-412 MB files finishing in 20-28 s each.

This supersedes the earlier per-callback-overhead explanation for #1784 — that overhead was real and
is still fixed (progress persistence is coalesced to ~1/s and the 0.2 s completion poll is gone), but
it was not the reason downloads were slow.

### 3. Transfers could silently stall, and retries gave up too early

The iOS background `URLSession` used the default 7-day resource timeout with no stall detection, so
a stalled task never errored and its concurrency slot was never freed. Android's dual-path manager
(streamed OkHttp + system `DownloadManager`, driven by a 500 ms poller) had latent wedge bugs and
thread-safety gaps.

Both platforms: a stall watchdog, guaranteed concurrency-slot release on every terminal path, launch
reconciliation of in-flight/persisted downloads, and throttled progress writes. Android additionally
puts all queue/bookkeeping state behind one lock, funnels terminal failures through a single path,
guards against divide-by-zero when the server omits `Content-Length`, claims the watcher coroutine
atomically (it could double-launch), synchronizes the duplicate-download check, and resolves parts
whose `enqueue()` / file-open fails instead of stranding them.

Device testing then exposed several ways the iOS retry logic still gave up on downloads that were
fine:

- **Error classification was an allowlist of nine URL error codes**, so everything it forgot was
  treated as fatal. A transient `NSURLErrorCannotParseResponse` (-1017) permanently failed five parts
  of one book within four seconds, retry budgets untouched. It is now a denylist — retry everything
  except what retrying cannot fix.
- **`NSURLErrorCancelled` (-999) was treated as terminal.** Force-quitting makes iOS tear down the
  session with -999, so reopening the app permanently failed every adopted part "after 0 retries".
  Every deliberate cancel is accounted for before the classifier is consulted, so a -999 reaching it
  is the system, and the transfer should resume.
- **The stall watchdog fired on app *resume*, not on real stalls.** It is a main-run-loop `Timer`, so
  it stops while suspended, but `lastProgressAt` is wall-clock — every healthy transfer looked ">30 s
  idle" the moment the user came back and was cancelled, burning retries and (without resume data)
  restarting large files from zero. A long gap between ticks now means "we were suspended" and
  re-baselines instead of cancelling; adopted tasks get a grace period.
- **Retry counts lived only in memory**, so every relaunch reset them and a part could cycle
  "attempt 1/3" forever, never completing and never failing. They are now persisted, reset when a
  transfer proves the connection recovered, and retries back off exponentially rather than
  re-queueing instantly.
- **Duplicate concurrent tasks for the same part.** Download-item and part ids are both deterministic,
  so a re-requested or reconciled item produced tasks for parts already in flight, and
  `activeDownloads[partId] = …` overwrote a live entry — orphaning a running task that kept writing
  to the same destination, untracked. Guarded at both the queue and item level.
- **A wedged download could not be restarted.** Tapping download on an item that already had a record
  returned early, so the only escape was reinstalling the app. An explicit request now replaces the
  record: in-flight tasks cancelled, stored state cleared, fresh parts queued.

### 4. Completed downloads could hang at "Processing…" (iOS)

A fully-downloaded item (files on disk) could hang forever. Two root causes:
- The delegate queue was concurrent, so a part's `didFinishDownloadingTo` (which commits
  `completed = true`) could overlap its `didCompleteWithError` on another thread. The new one-shot
  completion check then read the part's own `completed` as `false`, and — with the old main-thread
  polling loop removed — never re-ran. **Fix:** serialize the delegate queue and `refresh()` Realm
  before the check; a claim set makes finalization run once.
- Completion (`onItemDownloadComplete`) was emitted only from inside a `getLibraryItemWithProgress`
  server callback, so an offline / HTTP 429 / expired-token response silently dropped it. **Fix:**
  finalize immediately from the `DownloadItem`'s own captured metadata; server-side progress syncs
  normally afterward. Completion no longer depends on the network at all.

### 5. An HTTP error body could be saved as the audio file (iOS)

To URLSession any HTTP response is a successful transfer — an error page or JSON body is simply the
payload — and nothing checked the status code, so `didFinishDownloadingTo` moved whatever arrived
into place **as the track**. A refused request produced a "completed" download holding garbage that
only surfaces later as a file that won't play. The likely trigger is an access token expiring
mid-transfer, which a multi-gigabyte book can easily outlive.

Non-2xx responses are now rejected and the body discarded; `401/403/408/425/429` and `5xx` retry
(the retry rebuilds the URL, which is what picks up a refreshed token), `404/410/416` fail immediately.

### 6. Resumed downloads were invisible in the UI (iOS)

`load()` runs at plugin-init, ~125 ms *before* the web layer mounts and registers its listeners, so
`notifyListeners("onDownloadItem")` was dropped and a resumed download ran natively while never
appearing in the downloads list. Now emitted with `retainUntilConsumed`.

### 7. Android review feedback

All five points raised by @areidyOTH in review (thank you): internal transfers now stream to a
`.part` file renamed into place only on success (and deleted on terminal failure) instead of writing
straight to the final filename; a partially failed item is no longer handed to `FolderScanner`
(`isDownloadFinished` only means nothing is still running — a failed part is "completed" too, so
`didDownloadSuccessfully` now answers separately); signed `?token=` URLs are redacted before logging;
retries back off exponentially instead of being eligible on the next 500 ms tick; and an
`InternalDownloadManager.close()` failure after a good copy is promoted to a failure, since `close()`
is what flushes the final bytes.

## How have you tested this?

- **Real device (iPhone 15 Pro), against a 10-part ~3.5 GB audiobook** — this is what surfaced §1, §2
  and most of §3. Verified after the fixes: the whole book downloads at 39-48 MB/s aggregate; every
  session migration reports `resume=true`; backgrounding, foregrounding, force-quitting and relaunching
  mid-download all resume rather than restart; no spurious stall bursts on resume; duplicate tasks and
  superseded completions are dropped.
- **iOS Simulator (iPhone 17, iOS 26.5):** repeated downloads and re-downloads of large (~3 GB) and
  small titles all finalize; reproduced the original "Processing…" wedge and confirmed the fix
  (finalizes even when the server 429s / the token has expired).
- **Stall watchdog + failed-part path reproduced live (iOS Simulator):** cut the simulator's network
  mid-transfer and confirmed the full chain — stall watchdog → bounded retries → `markPartFailed` →
  `onDownloadItemPartUpdate {failed:true}` → failure toast naming the file, download control reverts
  instead of leaving a frozen progress bar. Screenshot below.
- **Android emulator (Android 15, arm64):** two internal-storage downloads complete and finalize with
  no wedge; single watcher start/finish cycle; clean `Item download complete` emission.
- **41 unit tests** covering the extracted logic — retry classification (including the -1017 and -999
  regressions), stall evaluation across a suspension gap, cross-launch retry/resume state, URL
  construction (including a test that a doubled address can never be produced again), and throughput
  maths.
- Both native targets build clean (`xcodebuild` for the `App` scheme; `./gradlew compileDebugKotlin`).
- Not yet exercised on device: §5 (hard to trigger a 401 deliberately) and the Android changes in §7,
  which are compile-verified only.

## Screenshots

No layout or markup changes in this PR. The only frontend change is script-only: it surfaces
a permanently-failed download part via the app's existing `$toast.error`. The screenshot below
shows that failure toast (top) firing on a real network-cut failure — the file is named,
and the download button has reverted (no more frozen progress bar).

download-failure-toast

> Note: the second toast in the capture ("Item download complete but failed to create library
> item") is pre-existing and not introduced by this change; the top toast is the new behavior.

---

### Notes for the maintainer
- **The dual-session change (§2) is the biggest behavioural change here** and deserves the closest
  look. It depends on the server answering Range requests with `206`; if a deployment did not, every
  migration would restart the file from zero, which would be worse than the throttle. Happy to gate it
  behind a capability check or a setting if you'd prefer.
- Trade-off: dropping the completion-time server refresh means local listening progress comes from the
  download-time snapshot (seconds-to-minutes old) and reconciles on the next normal sync — chosen so
  completion never blocks on the network.
- Cross-launch resume relies on resume data captured while the app was alive; a force-quit at exactly
  the wrong moment can still lose a partial transfer, because URLSession owns the temp file. At 40 MB/s
  that costs seconds rather than the previous half-hour.
- `statusNotification["libraryItemId"]` carries the *download-item* id (`libraryItemId-episodeId` for
  episodes) — pre-existing behavior the frontend already keys on, called out for clarity.


DrNgo:
Tested on-device: I installed this build on my own iPhone via my personal TestFlight, and downloads now complete quickly and correctly — files download reliably and play back without getting stuck. 👍

areidyOTH:
Thanks for this. I encountered the Android internal-storage freeze on a Pixel 6 running Android 16. I attempted to have GPT-5.6 and Claude Opus find and fix the problem but they found this PR fixes most of it. They found a few ways this PR could be improved as well. Begin LLM slop...

Before merging, four Android safety gaps should be addressed:

1. Internal transfers still write directly to `finalDestinationPath`. After the final retry fails, a truncated or zero-byte media file remains at its final filename.
2. `DownloadItem.isDownloadFinished` treats failed parts as finished, and `checkDownloadItemFinished()` still invokes `FolderScanner`. This can index the truncated file and report a partially failed item as locally downloaded.
3. Signed download URLs containing `?token=...` are still logged by both `DownloadItemManager` and `InternalDownloadManager`.

I recommend downloading internal parts to `<filename>.part`, atomically renaming only after successful completion, deleting `.part` files after terminal failure or cancellation, and bypassing `FolderScanner` whenever any required part failed.

I also suggest making retry timing explicit. Current retries appear eligible on the next 500 ms queue tick, so a sleeping device could exhaust all attempts rapidly before the user wakes it. Even though foreground execution is correctly outside this PR's scope, retry attempts should not form a tight loop.

One smaller edge case: if `InternalDownloadManager.close()` fails after a successful body copy, `finishWith()` currently logs the exception but still reports success. A close failure should promote the terminal result to failure.

DrNgo:
@areidyOTH Thank you — this was a genuinely useful review, and all five points are now addressed in `35defbc8`. Taking them in order:

**1. Internal transfers writing straight to `finalDestinationPath`.** Fixed as you suggested: they now stream to a sibling `<filename>.part`, which is renamed into place only on success and deleted on terminal failure, so nothing usable is left behind. You were right that this is the dangerous one — a truncated file sitting at the real filename is indistinguishable, to the scanner and to the user, from a good download.

**2. `isDownloadFinished` treating failed parts as finished, and `FolderScanner` still running.** Fixed. "Finished" and "succeeded" are now separate questions — `isDownloadFinished` only means nothing is still running, and a permanently failed part is `completed` too. Added `DownloadItem.didDownloadSuccessfully`; when any part failed, the item is resolved in the UI and the scanner is skipped entirely rather than indexing whatever arrived and reporting the book as available offline.

**3. `?token=` URLs in the logs.** Fixed — both managers now log through a `redactUrl()` helper that keeps the endpoint and drops the query. Worth flagging that this one has a wider blast radius than it looks: those logs end up in anything a user exports or attaches to a bug report.

**4. Retry timing / tight loop.** Fixed, and your reasoning was exactly right — the queue ticks every 500 ms, so a part that failed fast on a sleeping device burned every attempt within a couple of seconds, long before the user could do anything. Retries now back off exponentially (capped at a minute) and a part waiting out its backoff is no longer selectable (`getNextDownloadItemParts` takes an eligibility predicate). The same policy is applied on iOS.

**5. `close()` failing after a successful body copy.** Fixed — promoted to a failure. `close()` is what flushes the final buffered bytes, so failing there means the file is incomplete however well the transfer itself went; reporting success would hand the scanner a truncated track.

Two notes beyond your list, in case they're relevant to what you hit on the Pixel 6:

- Device testing on iOS turned up a **separate, larger root cause** that this PR originally missed: every track part persisted a download URL built as `address + address`, resolving to a host that doesn't exist. Only downloads restarted *from the database* used it, which is why fresh downloads worked and large books — the ones that inevitably get interrupted and reconciled — never finished. That is now fixed, along with iOS retry logic that was giving up on transfers that were fine (an error allowlist treating `-1017` as fatal, and a stall watchdog that cancelled healthy transfers every time the app was resumed).
- iOS downloads also turned out to be throttled to ~0.6 MB/s aggregate by the background `URLSession` itself; they now run on an unthrottled session while the app is in the foreground, which took a real device from ~0.6 MB/s to 39-48 MB/s. If Android downloads also feel slow to you relative to your connection, it may be worth measuring — I haven't checked whether Android's `DownloadManager` path has an equivalent ceiling.

The PR description has been rewritten to cover all of it. Thanks again for taking the trouble — points 1, 2 and 5 in particular were real data-integrity bugs, not just tidiness.

Issue #998, opened 2023-12-22
[iOS] UI treating short dragging as a tap/open when it shouldn't be


johnpyp:
### Steps to reproduce

1. Go to the main screen
2. With your finger on a book or any clickable element, drag your finger quickly and release. 
3. It will treat it as if you've clicked, but typical well-behaving iOS apps would not treat this as a press.

### Expected behaviour
Short drags should not result in a click

### Actual behaviour
Short drags over clickable elements cause a tap/press to happen. It does not seem to happen for prolonged drags though, so I imagine the bug is something where the UI is doing some press-timing math when it shouldn't be.

### Environment data

Audiobookshelf Version: 0.9.70-beta, server v.2.6.0

- [ ] Android App?
- [x] iOS App?

#### iOS Issue
iOS Version: 17.1.2

iPhone model: iPhone 15 Pro

johnpyp:
Maybe this could be loosely related to https://github.com/ionic-team/capacitor/issues/6826, though it's not exactly the same.

zacaronii:
This may be related to a similar problem I have. After testing a lot of different swipes, my issue seems to be more related to swiping down to scroll and then placing my finger on the screen to stop scrolling. If I land on an audiobook, when trying to stop scrolling, and lift my finger, it will load into that book. If I land and slide my finger a little bit in any direction, before lifting it up, it will usually prevent opening the book unintentionally.

For lack of a better way of putting it, the book elements seem to be a little "sensitive" to what is determined as a press/click. I am not a dev, so I'm sure there's a better explanation. I am happy to test further, and I understand this is likely not a high priority problem. Just wanted to provide my experiences related to this issue.

PR #1934, opened 2026-07-16
fix(ios): don't open a card when a quick flick is used to scroll

DrNgo:
## Brief summary

On iOS, starting a quick flick on a book cover often opens that item's details instead of just scrolling the shelf. This adds a small shared guard so a gesture used to scroll is no longer treated as a tap.

## Which issue is fixed?

No existing issue found — reporting and fixing it in one go. Happy to link one if a matching report exists.

## Pull Request Type

Frontend only. The symptom is iOS-specific (it stems from WKWebView click synthesis), but the change is plain Vue/JS in the shared web layer — no native or platform-specific code is touched. It is inert on Android and desktop, where the gesture is either already handled correctly or there are no touch events at all.

## In-depth Description

**Cause.** WKWebView does not reliably suppress the `click` it synthesizes for a short, fast flick. Instrumenting real gestures on a physical iPhone 15 Pro (iOS 26.5.2) showed two cleanly separable populations:

| gesture | travel | duration | scrolled | click fired? |
|---|---|---|---|---|
| deliberate tap | **0px** | 40–96ms | no | yes — correct |
| flick to scroll | **17–44px** | 50–130ms | **yes** | **yes — the bug** |

A flick that short still carries enough velocity to scroll the list with momentum, yet stays under WebKit's own suppression threshold — so the list scrolls *and* the cover under the finger is activated. WebKit's own behaviour is erratic here: in one session a 384px flick and a 61px flick were both correctly suppressed, while a 25px flick fired a click.

The cards navigate on any click the WebView hands them, with no check that the gesture was actually a tap. That is why every card surface is affected (Home shelves, library grid, list view) and why it is long-standing rather than a recent regression — all five card components share a bare `@click="clickCard"`.

**Solution.** `utils/tapGuard.js` installs one set of passive, capture-phase listeners for the whole app and exposes `wasDragGesture()`. Each `clickCard` bails early when it reports a drag.

It combines two signals, because each catches what the other misses:

- **Did anything scroll under the finger?** A flick shorter than the slop still scrolls with momentum — the observed clicks at 17px and 19px are caught only by this.
- **Did the finger travel past a 20px slop?** A drag on a shelf already at its end scrolls nothing and fires no scroll event — caught only by this.

The scroll signal is gated on a touch actually being down, so a mouse-wheel scroll cannot latch it on non-touch devices. Resetting on `mousedown` instead would not work: iOS synthesizes `mousedown` after `touchend` but before `click`, which would clear the flag immediately before the guard reads it.

**Why this approach.** One shared module rather than per-card handlers means a single set of listeners for the app instead of a pair per card, and it fixes all five card types at once. The listeners are passive and capture-phase so they cannot interfere with scrolling or with WebKit's own gesture handling. The 20px slop sits in the measured gap: deliberate taps travelled 0px (worst observed roll on any control was 16px), accidental flick-clicks 17–44px.

**One behaviour change worth flagging:** a tap landing while the list is still coasting is now treated as a scroll and ignored. This is deliberate and matches UIKit, where the first tap on a decelerating scroll view stops it without selecting. If you would rather keep this purely distance-based, that half is a one-line removal.

This affects any iOS user who scrolls their library, not an edge case of one setup.

## How have you tested this?

Tested on a physical iPhone 15 Pro (iOS 26.5.2) against a live server — the simulator was not used, as synthesized drags do not reproduce real thumb timing.

Reproducing before the fix:
1. Open the app on an iOS device with a library of several books.
2. Put a finger directly on a book cover and flick quickly to scroll (a short, fast motion — roughly 25–45px in under ~130ms; a slow drag will not do it).
3. Repeat a few times: the shelf scrolls but a book's details page opens unintentionally.

After the fix, the same flicks scroll without opening anything, while deliberate taps still open books normally. Verified on both the Home shelves and the library grid.

Measurements in the table above were gathered by temporarily logging touch displacement, duration and scroll activity per gesture; that instrumentation was removed before this PR.

## Screenshots

Not applicable — this is a touch-handling fix with no visual change. There is nothing to show in a still image: the difference is that a flick no longer navigates. The before/after is captured in the table and repro steps above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/components/cards/LazyBookCard.vue
+++ b/components/cards/LazyBookCard.vue
@@ -102,6 +102,7 @@
 
 <script>
 import { Capacitor } from '@capacitor/core'
+import { wasDragGesture } from '@/utils/tapGuard'
 
 export default {
   props: {
@@ -479,6 +480,9 @@ export default {
       eventBus.$emit('play-item', { libraryItemId: this.libraryItemId, episodeId: this.recentEpisode.id })
     },
     async clickCard(e) {
+      // Ignore the click iOS synthesizes for a quick flick used to scroll
+      if (wasDragGesture()) return
+
       if (this.isSelectionMode) {
         e.stopPropagation()
         e.preventDefault()
--- a/components/cards/LazyCollectionCard.vue
+++ b/components/cards/LazyCollectionCard.vue
@@ -14,6 +14,8 @@
 </template>
 
 <script>
+import { wasDragGesture } from '@/utils/tapGuard'
+
 export default {
   props: {
     index: Number,
@@ -59,6 +61,8 @@ export default {
       this.isSelectionMode = val
     },
     clickCard() {
+      // Ignore the click iOS synthesizes for a quick flick used to scroll
+      if (wasDragGesture()) return
       if (!this.collection) return
       var router = this.$router || this.$nuxt.$router
       router.push(`/collection/${this.collection.id}`)
--- a/components/cards/LazyListBookCard.vue
+++ b/components/cards/LazyListBookCard.vue
@@ -49,6 +49,7 @@
 
 <script>
 import { Capacitor } from '@capacitor/core'
+import { wasDragGesture } from '@/utils/tapGuard'
 
 export default {
   props: {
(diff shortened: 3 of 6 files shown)


advplyr:
hey did you build this and test on your device?

DrNgo:
Yes I did. I'm actively using this on my fork

nichwall:
Would this fix https://github.com/advplyr/audiobookshelf-app/issues/998 ?

I am a bit unclear whether this fully fixes that issue if this is only for book covers or all scroll events misinterpreted as a click.

PR #1935, opened 2026-07-17
Fix custom folder downloads stuck at 0% on Android 17

sorantis:
## Summary
- Fixes Android custom-folder downloads that could start but remain stuck at 0% on Android 17 devices.
- Routes custom-folder downloads through the app-managed OkHttp downloader instead of Android DownloadManager.
- Keeps the existing SAF move into the selected folder after each temp download completes.
- Hardens download stream handling by closing output streams reliably, failing non-success HTTP responses, and avoiding divide-by-zero when content length is unknown.

## Context
On Android, internal/default downloads already used the app-managed downloader. Custom-folder downloads used a different path: Android DownloadManager fetched into an app temp file, then the app moved the completed file into the selected SAF folder. On Android 17 this path could appear to start but never report progress or produce files in the selected folder.

This change removes DownloadManager from the custom-folder fetch path so both internal and custom-folder downloads use the same app-managed download behavior. Custom-folder downloads still use SAF for the final move into the user-selected folder.

## Testing
- Built debug APK with ./gradlew assembleDebug.
- Installed the debug APK locally on an Android 17 Pixel.
- Verified a book download to a custom folder progresses beyond 0% and completes successfully.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
@@ -63,6 +63,7 @@ class DownloadItemManager(
 
   companion object {
     var isDownloading: Boolean = false
+    private const val APP_MANAGED_DOWNLOAD_ID = -1L
   }
 
   /** Adds a download item to the queue and starts processing the queue. */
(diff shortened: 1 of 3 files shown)


advplyr:

PR #1936, opened 2026-07-18
Fix Android downloads to custom folders stuck at 0%

### LINKED ISSUE #1273: [Bug]: Download complete but failed to create library item
### What was the Problem?

I am unable to download library items to a custom folder location on my phone's SD card. The download completes successfully, but I get the error 'Download complete but failed to create library item'. It's worth noting that the folder hierarchy for the author/books is created on the file system, and the cover image for the book is also added successfully. However, none of the audio files are added to the filesystem. There is more than enough space on the SD card (almost 100GB free).

1. Create new custom download location
2. Download book to selected location
3. See error described above

### What was Expected?

Audio files should download without error

### Phone Model

Samsung Galaxy A01

### Phone OS

Android 12

Android App - 0.9.74

### Installation Source

Google Play Store

### Additional Notes

Some other issues that seemed similar or possibly related, but not the same (or closed already):
- https://github.com/advplyr/audiobookshelf-app/issues/422
- https://github.com/advplyr/audiobookshelf-app/issues/775
- https://github.com/advplyr/audiobookshelf-app/issues/961

gordienoye:
## Brief summary

Fixes downloads to a custom local folder getting stuck at "Downloading... (0%)" indefinitely and never completing. Downloads to Internal Storage were unaffected.

## Which issue is fixed?

Fixes #1273

## Pull Request Type

Android only. Native/backend change (Kotlin) — no frontend, UI, or iOS changes.

## In-depth Description

Downloads to a custom SAF-picked folder were routed through the system `DownloadManager` via `DownloadManager.Request.setDestinationUri()` with a raw `file://` URI. On API 29+, `DownloadManager` only permits destinations that are either app-owned or the top-level Downloads directory — a user-picked custom folder is neither, so the enqueued download silently sat at `STATUS_PAUSED` with 0 bytes downloaded, permission or no permission. This is a hard platform restriction, not a fixable permission gap (see [Android's own documentation](https://developer.android.com/privacy-and-security/risks/unsafe-download-manager)), and this code path had been unchanged since January 2023.

A recent Android/Pixel OS update appears to have tightened enforcement of this restriction — verified via a controlled test across two physical devices (a Pixel 7a and a Pixel 9a), where the same production app version downloaded successfully to a custom folder on the 9a until it received the same OS update the 7a had already gotten, at which point both devices exhibited identical failures. So this isn't limited to old OS versions or specific hardware — it will likely affect more users as their devices take routine OS updates.

The fix routes custom-folder downloads through the same OkHttp/`FileOutputStream` mechanism already used (and already working) for Internal Storage downloads, caching into the app's own external files directory before moving the finished file into the user's chosen folder via the existing, unmodified Storage Access Framework move step (`moveDownloadedFile()` / `moveFileTo()`). This removes the system `DownloadManager` dependency from downloads entirely, along with its dead status-polling code, and unifies what were previously two divergent code paths into one.

Trade-off: downloads no longer get automatic retry/resume on a dropped network connection, or an OS-level download notification — both were only ever provided by the system `DownloadManager`, and Internal Storage downloads already lived without them, so this isn't a new class of risk, just extending an existing, working approach to the second path.

## How have you tested this?

Manually, on a physical Pixel 7a (Android 15) connected via ADB:
1. Built a debug APK (`./gradlew assembleDebug`) and installed it via `adb install -r`.
2. Reproduced the original failure first: downloading to a custom folder named "Audiobooks" got stuck at 0% indefinitely (confirmed via live `logcat`, tracks stuck at `STATUS_PAUSED`, `bytesDownloaded=0`).
3. Applied the fix, rebuilt, reinstalled.
4. Retried the same download to the same "Audiobooks" custom folder while tailing `logcat`. Confirmed all chapter files and cover art downloaded, moved into the custom folder, and the app logged `Download Item finished` followed by successful local library item creation.
5. Confirmed Internal Storage downloads are unaffected (unchanged code path).
6. Full debug build passes (`./gradlew assembleDebug` — BUILD SUCCESSFUL, no new warnings introduced).

No automated test coverage exists for the downloader in this repo (`app/src/test` / `app/src/androidTest` contain only unmodified Capacitor boilerplate), so this PR relies on the manual on-device verification above.

## Screenshots

Not applicable — this PR contains no frontend/UI changes.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
@@ -1,6 +1,5 @@
 package com.audiobookshelf.app.managers
 
-import android.app.DownloadManager
 import android.net.Uri
 import android.util.Log
 import androidx.documentfile.provider.DocumentFile
@@ -28,7 +27,6 @@ import kotlinx.coroutines.launch
 
 /** Manages download items and their parts. */
 class DownloadItemManager(
-        var downloadManager: DownloadManager,
         private var folderScanner: FolderScanner,
         var mainActivity: MainActivity,
         private var clientEventEmitter: DownloadEventEmitter
@@ -38,12 +36,6 @@ class DownloadItemManager(
   private var jacksonMapper =
           jacksonObjectMapper()
                   .enable(JsonReadFeature.ALLOW_UNESCAPED_CONTROL_CHARS.mappedFeature())
-
-  enum class DownloadCheckStatus {
-    InProgress,
-    Successful,
-    Failed
-  }
 
   var downloadItemQueue: MutableList<DownloadItem> =
           mutableListOf() // All pending and downloading items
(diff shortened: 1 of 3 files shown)


advplyr:
was this covered already in https://github.com/advplyr/audiobookshelf-app/pull/1937 ?

gordienoye:
Closing this — the root cause it fixed (system `DownloadManager.setDestinationUri()` silently failing on custom SAF folders) has since been independently resolved by the download-subsystem rewrite in #1587 (`InternalDownloadManager` + OkHttp-based streaming for all downloads, internal and external, with SAF `moveDownloadedFile()` handling the final move). That rewrite supersedes this branch entirely and this PR no longer applies cleanly against current `master`.

Thanks for reviewing what came in — no action needed here.

Issue #1253, opened 2024-07-05
[Bug]: Can't download audiobooks


Cossty:
### What was the Problem?

Recently, I finished setting up audiobookshelf for the first time. I decided to download 30+ audiobooks. Some of them have multiple tracks, so the android app was downloading around 400 files. After it downloaded around 100 it was stuck and didn't continue. I had to restart the app, disconnect and reconnect to the server, just to be able to download again. Tried again to download a lot of files at once. Same result, stalled after a while. 

Then I tried to find some answers here. The only thing I found was that when you start downloading a book, you should wait a little on it's page, to let the download set up. So I did that, I waited on each book 5-10s, before leaving for another one. It looked like it helped a little, because a lot more files got downloaded this time, but it stalled anyway. My internet is downloading books 5-6 m/s.

After that, I was downloading books one by one, which worked until it didn't. I have 42 local books. Audiobookshelf app uses 25.61 gb of user data. Now the download stalls before finishing one book. 
Connecting on local network. Battery optimization is set to unrestricted.

### Steps to Reproduce the Issue

1. Have no local books
2. Que a lot of books/files for download. 400+

1. Have a 42+ local books/25.61gb of user data
2. try to download another book

### What was Expected?

To be able to download books without stalling.

### Phone Model

Fairphone 4

### Phone OS

Android 13, stock, but stock for fairphone is /e/OS

### Audiobookshelf App Version

Android App - 0.9.74

### Installation Source

Google Play Store, Other (List in "Additional Notes")

### Additional Notes

_**App downloaded with Aurora Store, but if I understand it correctly, that's just google play version.
I don't have logs for when I was downloading a lot of books at once.
There are a lot of ping timeouts in my logs, is this intended?**_

2024-07-05 07:03:21.735  INFO  [SocketAuthority] Socket 00Pu9GzjwYI4xz8fAABl disconnected from client "cossty" after 426859ms (Reason: ping timeout)

2024-07-05 07:03:58.566  INFO  [SocketAuthority] Socket Connected HoT8oJ3gsiCfn9bYAABn

2024-07-05 07:11:51.829  INFO  [SocketAuthority] Socket HoT8oJ3gsiCfn9bYAABn disconnected from client "cossty" after 473263ms (Reason: ping timeout)

2024-07-05 07:33:51.669  INFO  [SocketAuthority] Socket Connected zbqvUMnkO3__VytrAABp

[...]

PR #1937, opened 2026-07-21
Android download rewrite, try 2

### LINKED ISSUE #34: partial downloads break downloading of an audiobook

### LINKED ISSUE #862: Download can't be restarted and would fill up internal app storage

nichwall:
## Brief summary

This PR fixes various Android download issues including:
- Continuing interrupted download
- Moving all downloads to Internal App Storage for performance (moves to Shared Storage after download completed)
- Check device has enough storage before downloading
- Download as foreground service to continue downloading after app closed
- Clean up incomplete downloads after 7 days if never resumed or retried

This PR supersedes https://github.com/advplyr/audiobookshelf-app/pull/1587

## Which issue is fixed?

Fixes #862 
Fixes #34 

Should also fix (not sure whether to mark them as fixed or not, but they are all similar issues). Previous testing from #1587 showed the main issue reported was background downloading was not supported, but this is now supported:
- #1479 
- #1273 
- #1167 
- #1253 
- #215 

## Pull Request Type

Android backend

## In-depth Description

This PR is a replacement for #1587 that I created over a year ago, but never finished and wanted to clean it up some more. I have copied bullets from before below, with some modifications:
- `LocalFile` and `LocalFolder` class updates:
  - Remove `simplePath` because it is one of several slow properties (SAF) and is unused anywhere in the codebase.
- Build the `LocalLibraryItem` object from the `DownloadItem` object instead of scanning folder in Shared Storage.
  - Much faster than waiting on SAF. Previous testing showed this takes about 150 ms per file, did not gather new statistics for this PR.
- Use temporary files to download and allow resuming partial downloads
- Retry download from last position in case of network interruption
  - If downloading to Shared Storage and the file already exists and is the correct size, mark as complete and continue. I am not performing this same behavior for Internal App Storage because if that has gotten corrupted, there is probably something else going on and would need to be downloaded again anyway.
- Fix some download requests inserting the user Access token into the URL
- Continues downloading in the background using a new `DownloadService` and `DownloadServiceHost`
  - Notification to cancel download (I was unable to test this part on my Pixel 6a on Android 16)
- Check available storage before downloading file to ensure storage does not become completely full
- Failed downloads are cleaned up after 7 days
  - Check at app startup and with background task in case app is not reopened (have not tested the background task)
- Added translation strings for new download prompts and download notification

Codex was used as part of cleaning up my old PR. I avoided autoformatting with Prettier (except for a few cases where I forgot to disable it) to help make the review process simpler.

## How have you tested this?

I tested the following using a Pixel 6a on Android 16, and an Android 10 emulator and confirmed changes work as expected. 

Conditions tested:
- Keep app in foreground
- Put app in background
- Force close app (swipe app away from background) during download
- LAN WiFi
- 5G Cell service (limited testing here due to data usage, mostly used friend's network in place of Cell)
- Friend's house for slower download speed

Tests performed:
- Download single file book to Internal App/Shared Storage
- Download book with both audiobook and ebook to both Internal App/Shared Storage
- Download podcast episode to both Internal App/Shared Storage
- Download multi-file book to Internal/Shared Storage
- Download series to Internal/Shared Storage
- Download already-existing book to Shared Storage

## Screenshots

No frontend changes, only Android backend


damccull:
All seems to be working great for me! Better than the previous version. Only problem is it wouldn't upgrade and I had to remove and reinstall. Merge ASAP!

nichwall:
> All seems to be working great for me! Better than the previous version. Only problem is it wouldn't upgrade and I had to remove and reinstall. Merge ASAP!

Thanks for the feedback! Can you clarify by what you mean "it wouldn't upgrade and I had to remove and reinstall"? Do you mean just installing the side loaded APK, or was there data corruption?

damccull:
> > All seems to be working great for me! Better than the previous version. Only problem is it wouldn't upgrade and I had to remove and reinstall. Merge ASAP!
> 
> Thanks for the feedback! Can you clarify by what you mean "it wouldn't upgrade and I had to remove and reinstall"? Do you mean just installing the side loaded APK, or was there data corruption?

I had the APK from the previous pr installed and tried to update this one over it. Didn't work. Had to uninstall and reinstall. Betting it's due to Dev signing certificate changed though.

jordanread:
What good timing! I just started playing with the app after using the web ui for a year to download and play in a different audiobook program. The initial downloads worked pretty well, and I immediately fell in love with the actual syncing with the server functionality. Had some issues yesterday, so started digging through the logs to see what I could see. And then checked here. Installed this apk today, and the downloads were a thing of beauty. Suddenly seeing the Downloaded file messages appear in the server logs was a highlight of my day.

OneHoopyFrood:
Looks like this creates a lot of linting noise. Makes it hard to review the actual change. I'm not a reviewer for the app or anything, obviously. I just was reading through and thought I'd mention it as I'd LOVE to see this PR get merged.

advplyr:
I'm going through this now. Thanks @nichwall and thanks everyone for testing

advplyr:
I'm on a pixel 9a with Android 16 and not seeing a download notification for internal or shared storage downloads. Is everyone here seeing a notification? What are you testing with?

nichwall:
> I'm on a pixel 9a with Android 16 and not seeing a download notification for internal or shared storage downloads. Is everyone here seeing a notification? What are you testing with?

I do not get a notification on a Pixel 6a with Android 16, but did get the notification on an Android 10 emulator.

android/app/src/main/java/com/audiobookshelf/app/managers/IncompleteDownloadCleanup.kt
```diff
@@ -0,0 +1,94 @@
...
+import androidx.work.WorkerParameters
+import com.audiobookshelf.app.device.DeviceManager
+import com.audiobookshelf.app.models.DownloadItem
+import java.io.File
+import java.util.concurrent.TimeUnit
+
+/** Removes terminally failed downloads after their retention window elapses. */
+object IncompleteDownloadCleanup {
+  private const val tag = "IncompleteDownloadCleanup"
+  private const val RETENTION_MS = 24L * 60L * 60L * 1000L
```
advplyr, line 20:
```suggestion
  private const val RETENTION_MS = 7L * 60L * 60L * 1000L
```

Should this be 7 days?

nichwall: Oh, yeah that is only 24 hours, not 7 days. To be 7 days it would be an additional multiply by 7.

I picked 7 days arbitrarily, 24 hours would also work for an incomplete download is deleted.

advplyr:
We needed to add POST_NOTIFICATION to the manifest for this to work. So users will need to approve that permission when opening the app. The reason it worked before is because those notifications came from the built-in downloader service. Using our own notifications require that permission on android 14+ I think it is.

Please give it a try

nichwall:
Awesome, thanks for reviewing. I can test it on my Pixel when I'm home in a few hours.

nichwall:
> We needed to add POST_NOTIFICATION to the manifest for this to work. So users will need to approve that permission when opening the app. The reason it worked before is because those notifications came from the built-in downloader service. Using our own notifications require that permission on android 14+ I think it is.
> 
> Please give it a try

The download notification is working for me on Android 16 on my Pixel 6a, thanks for fixing that. I confirmed the download continues and the notification is persistent with the app closed in the background as well.

jordanread:
Just tested and confirmed with Pixel 8 Pro Android 17. Permissions requests all as expected. Download notification does work when the app is in the background, as it does when downloading existing media (had to pick a long book cause it's pretty fast). I doubt that this is surprising, but it does respect the existing rules for notifications of this type. I have download notifications silent, so they only show up when the notifications are expanded, and this respects that.

android/app/src/main/java/com/audiobookshelf/app/services/DownloadService.kt
```diff
@@ -0,0 +1,102 @@
...
+                    CHANNEL_ID,
+                    DownloadServiceHost.notificationStrings(this).downloads,
+                    NotificationManager.IMPORTANCE_LOW))
+  }
+
+  private fun pendingIntentFlags(): Int = PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
+
+  companion object {
+    private const val CHANNEL_ID = "downloads"
+    private const val NOTIFICATION_ID = 4102
```
advplyr: I'm curious why you chose 4102 here

nichwall: No specific reason. Codex did most of this file and I did not change the notification identifier.

advplyr: Cool, just updated to 11 because the player notification ID is 10

advplyr:
I updated the notification so that the cancel button is reflected in the UI. I wasn't sure about how we should handle dismissing the notification. I was going to try to see how this is normally handled. Let me know what you think about that.

This is a major improvement to downloads from what we had previously. I will be doing a lot more testing in the next few days to get ready for the next release. Thanks for all of your effort on this!

nichwall:
Awesome, thanks for taking a look and finishing up the notification edge cases! I'll also try and get some more testing of the additional changes over the next few days.

Keavon:
I'm really glad to see work towards these major issues is headed toward the next release (whenever that might be, I'll be looking forward to it 😄). Thank you!

I also feel obliged to link #1371 as an on-topic continuation of this area of focus, as what I believe should become the optimal end-goal for this functionality. Perhaps the current momentum from this PR could be continued with another that resolves that too. I'd be grateful, at least! Since for me, that continues to be the single major defect with the whole app (anything else that's wrong is very minor by comparison).

PR #1938, opened 2026-07-21
Remember sort & filter options for podcast episodes

### LINKED ISSUE #304: [Enhancement] Remember sorting / make sorting persist for podcast episodes.
Feature Request

I'd like for the sorting to persist regarding Podcasts. Currently, the default sorting method is "most recent podcast episode" and while you can change it, once you leave the specific podcast it resets.

However, especially for audio drama podcasts or something that splits its topic over multiple episodes, sorting by most recent is inadequate. 
So please let the sorting that we set when we go on to a podcast persist once we leave it - both in the webinterface and the app.

cosandr:
## Brief summary

Fixes a pet-peeve bug/missing feature I noticed, namely that podcast episode sorting/filtering options weren't saved and always defaulted to "incomplete".

## Which issue is fixed?

Fixes #304

## Pull Request Type

Affects both Android and iOS, frontend only.

## In-depth Description

I will disclaimer that I am not a Node/JS/mobile developer, so the coding was done using Claude. It took several rounds as I immediately noticed the solutions were totally different than the rest of the codebase.

I do want to bring up something Claude noticed, but I refused to change as I'm not familiar enough with the code base, snippet (it's talking about [Fix ascending bug](https://github.com/advplyr/audiobookshelf-app/commit/9562e8f25271693116060a77dc1d309c081a5a0e)):

> Root cause: getUserSetting uses ||, collapsing falsy-but-valid values (false, 0, '') to null. Broke podcastEpisodesOrderDesc: false (ascending).
> Fix, store/user.js:51:
return state.settings?.[key] ?? null

Reasoning is that function is called many other places and since this isn't my field of expertise, I'd rather not change it blindly.

If you don't want AI assisted code merged, I totally understand. I'll keep using this build for myself until a fix is merged upstream some other way.

## How have you tested this?

In the Android emulator and on a Pixel 9 Pro. I have not tested the iOS app but I do have a Mac so I could perhaps try it on the emulator if need be.

It's pretty simple to test, in the master branch, any changes to sorting/filtering for podcasts are lost when navigating away, either going back or tapping on a podcast episode. The scroll position is maintained.

With this branch, this doesn't happen. Both sorting and filtering are preserved when going into an episode or backing out.

## Screenshots

I don't think this is necessary, if you insist I can add a video later.

### DIFF
--- a/components/tables/podcast/EpisodesTable.vue
+++ b/components/tables/podcast/EpisodesTable.vue
@@ -234,8 +234,8 @@ export default {
     },
     sortDesc: {
       get() {
-        const desc = this.$store.state.user.settings.podcastEpisodesOrderDesc
-        if (desc === null || desc === undefined) return this.mediaMetadata.type === 'episodic'
+        const desc = this.$store.getters['user/getUserSetting']('podcastEpisodesOrderDesc')
+        if (desc == null) return this.mediaMetadata.type === 'episodic'
         return desc
       },
       set(val) {
--- a/components/tables/podcast/EpisodesTable.vue
+++ b/components/tables/podcast/EpisodesTable.vue
@@ -50,7 +50,7 @@
 
     <modals-podcast-episodes-feed-modal v-model="showPodcastEpisodeFeed" :library-item="libraryItem" :episodes="podcastFeedEpisodes" />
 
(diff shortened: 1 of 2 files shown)


advplyr:
Thanks!

PR #1951, opened 2026-08-08
Add resumable server uploads

keeper1978:
## What changed

- add an upload page for users with the server upload permission
- allow selecting all files belonging to one audiobook with the Android system file picker
- upload files in 8 MiB chunks and display aggregate progress
- reinitialize an interrupted upload and continue from the offsets reported by the server
- preserve the selected library, folder, and book metadata locally
- add English and Russian UI strings

## Why

Uploading a large audiobook in one request means any connection interruption forces the user to restart from byte zero. Chunked uploads reduce repeated data transfer and make uploads over unstable mobile connections practical.

## Dependency

This draft depends on the proposed server API in advplyr/audiobookshelf#5443. It is intentionally opened as a draft so the client and server API shape can be reviewed together.

Because browsers do not allow a site to persist access to arbitrary selected files, the user must select the same files again after the app is terminated. The stable upload fingerprint lets the server return the saved offsets and continue instead of restarting.

## Validation

- Nuxt production generation succeeds
- generated route `/upload` is present
- Capacitor Android sync succeeds
- Android `assembleDebug` succeeds


nichwall:
The app is intended for consuming media from the server, not administration or adding media to the server. This would be done through the web client, not through the app.

keeper1978:
Thank you for clarifying the intended scope of the mobile app. This PR proposed the wrong workflow for the issue we were trying to solve, so I am closing it. The actual problem is recovery of interrupted server-to-device audiobook downloads; that fix is now proposed separately in #1953.

PR #1953, opened 2026-08-09
Recover interrupted audiobook downloads

keeper1978:
## What changed

- recover interrupted terminal downloads when their staging files are still present
- retry an existing unfinished download instead of only reporting that it was already started
- handle HTTP 416 by accepting a complete local file or restarting a stale partial file from byte zero
- finalize downloads correctly when Android document providers normalize or hide filename extensions

## Why

Interrupted multi-file audiobook downloads could leave a small number of tracks at 100% without marking them complete. Retrying the book then returned `Download already started for this media entity`, leaving no way to recover without clearing application data and downloading the entire book again.

The server files and advertised sizes can be valid in this situation. The failures occur while restoring a terminal queue entry, recovering from an invalid byte range, or reopening a file after Android's Storage Access Framework has normalized its name.

## User impact

Users can retry an interrupted book download. Completed files are reused, stale partial files restart safely, and files copied successfully into shared storage are recognized and finalized.

## Validation

- `./gradlew :app:compileDebugKotlin`
- `./gradlew :app:assembleDebug`
- verified a debug APK is produced successfully

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
@@ -78,6 +78,16 @@ class DownloadItemManager(
         checkDownloadItemFinished(item)
         return@forEach
       }
+      val recoverTerminalFailure =
+              item.terminalFailureAt != null &&
+                      item.downloadItemParts.any { part ->
+                        part.failed && File(part.destinationPath).exists()
+                      }
+      if (recoverTerminalFailure) {
+        Log.i(tag, "Retrying interrupted download ${item.id}")
+        item.terminalFailureAt = null
+        IncompleteDownloadCleanup.cancel(context, item.id)
+      }
       item.downloadItemParts.forEach { part ->
         if (part.moved) return@forEach
         if (item.terminalFailureAt != null && part.failed) return@forEach
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
@@ -138,6 +138,29 @@ class DownloadItemManager(
   }
 
   @Synchronized
+  fun retryDownloadItem(downloadItemId: String): Boolean {
+    val item = downloadItemQueue.find { it.id == downloadItemId } ?: return false
+    if (item.downloadItemParts.any { it in currentDownloadItemParts }) return false
+    if (item.isDownloadFinished) return false
+
+    item.terminalFailureAt = null
+    IncompleteDownloadCleanup.cancel(context, item.id)
+    item.downloadItemParts.filter { !it.moved }.forEach { part ->
+      part.failed = false
+      part.completed = false
+      part.isMoving = false
+      part.waitingForSpace = false
+      part.downloadId = null
+      part.retryCount = 0
+      part.lastUpdateTime = System.currentTimeMillis()
+    }
+    persist(item, force = true)
+    checkUpdateDownloadQueue()
+    notifyQueueChanged()
+    return true
+  }
+
+  @Synchronized
   fun cancelAll() {
     activeCalls.values.forEach(Call::cancel)
     activeCalls.clear()
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
@@ -5,6 +5,7 @@ import android.net.Uri
 import android.os.StatFs
 import android.util.Log
 import androidx.documentfile.provider.DocumentFile
+import com.anggrayudi.storage.file.fullName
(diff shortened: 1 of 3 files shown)


nichwall:
A large Android download PR was merged last week which supports background downloads and resuming incomplete downloads.

Does this include new functionality or fix a bug from the PR, or is this a PR against 0.13.0-beta without testing the download updates?

keeper1978:
This is a bug fix for the new download implementation that was recently merged, not a PR against 0.13.0-beta.

The branch is based on `master` at commit `12025ab` (`Version bump 0.14.0-beta`). I reproduced the issue using a debug APK built from that code.

In the affected multi-file audiobook, three files repeatedly reached 100% but were never finalized. Retrying returned `Download already started for this media entity`. The server responded with HTTP 416 when the client attempted to resume using an invalid saved offset.

I verified that the original server files were intact and that their actual sizes matched the sizes stored in the Audiobookshelf database:

- `0005.mp3`: 15,030,241 bytes
- `0008.mp3`: 14,492,833 bytes
- `0011.mp3`: 14,723,953 bytes

The PR addresses several recovery paths in the new downloader:

- if a partial file already matches the expected size, treat it as complete;
- if the saved offset is invalid and the server returns 416, remove the stale staging file and retry from byte zero;
- allow a terminal interrupted queue entry to be retried instead of only returning “Download already started”;
- recognize a finalized SAF file when Android’s document provider normalizes or hides its extension.

I rebuilt and tested the debug APK after these changes. The intention is to fix recovery/finalization edge cases in the newly merged background downloader, not to introduce a separate download system.

keeper1978:
что у нас с ответом по аудиобокселф

ljlongwing:
This recovers 416 / interrupted transfers well, but the download path still has no handling for a **401** on the file request — a separate way a download wedges into the same "stuck at 100% / Download already started" state.

`InternalDownloadManager.download()` sends `Authorization: Bearer <token>` using the access token read once in `DownloadItemManager.startDownload()`. Every other network path refreshes on 401 (`ApiHandler.makeRequest` → `handleTokenRefresh`, and the JS `nativeHttp`), but the file downloader doesn't. If the access token expires while items sit in the queue, or mid-transfer on a large book, each part fails and burns its retries against the dead token, then latches `terminalFailureAt`. It only clears once something else in the app refreshes the token — which is why several reports in #1970 / #1963 that aren't the Range bug say "kill the app and reopen, then it works."

I have a small change layered on this idea: `InternalDownloadManager` reports a 401 distinctly, `DownloadItemManager` parks the part (no retry consumed), refreshes via a reusable `ApiHandler.refreshAuthTokens()` extracted from the existing `handleTokenRefresh`, then re-queues so the part restarts with the fresh token — bounded so a genuinely dead refresh token still fails out. Same three files, sits alongside the 416 branch here.

Happy to open it as a follow-up once this merges, or fold it in here if you'd prefer.

keeper1978:
Thanks for identifying this. The 401/token-refresh path is a separate failure mode from the 416 and SAF finalization fixes in this PR, so keeping it in a focused follow-up PR would be preferable.

Please open your change as a follow-up and link it here. That will keep this PR reviewable while still addressing the other cause of downloads becoming stuck. The bounded refresh behavior you described sounds appropriate, especially not consuming normal transfer retries while authentication is being refreshed.

ljlongwing:
Opened as #2005.

nichwall:
Thanks for helping hunt this down and providing a PR. :)

I believe this functionality (and #2005) is fixed in #2002. This is a follow-up PR I made includes several different fixes in a single PR so it is easier to test all of the changes together in a single APK (and reduce the chance of new issues arising when merging multiple PRs related to the same thing).

If y'all want to review what I did to make sure it still fixes the issue, that would be great.

keeper1978:
Thanks for incorporating these fixes into #2002 and testing them together. Since #2002 has been merged, this PR is now superseded and can be closed.

ljlongwing:
Reviewed the merged code on master (#2002 plus `e06ce4a8` / `71bd354b`) against the original 401 failure mode and it holds up.

- `InternalDownloadManager` returns `onAuthError()` on a 401 instead of a generic failure, so the 401 no longer burns the transfer retries against a dead token.
- `DownloadItemManager.handleAuthError` tracks a separate `authRetryCount` bound at `MAX_AUTH_RETRIES`, refreshes via `refreshAuthTokens(serverConnectionConfigId, ...)` so a queued download targeting a non-active server still works, and dedupes concurrent parts with `refreshingServerIds`.
- The `RefreshResult` split is the key improvement over what I had in #2005: `Transient` failures (network blips, parse errors, non-401/403 codes) leave credentials untouched and retry, and only an explicit `Rejected` clears the session. My version would have logged the user out on any refresh failure, including a transient one.
- Persistent failure ends at `markTerminalFailure` + `IncompleteDownloadCleanup` rather than spinning forever.

The #1953 recovery path (queue restore, 416 = already-complete handling, stale-range restart from byte zero) is intact as well. Looks good to me superseding both #1953 and #2005.

Issue #1945, opened 2026-08-02
[Bug]: Progress overwritten / marked as finished from old browser session


0verEngineer:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

I listened to a audiobook via the app and everytime i paused (since yesterday) it was overwritten with 100% progress / marked as finished. I checked the log and it says that the 100%-progress comes from MacOS so i checked my always running MacMini and saw that it is playing on the MacMini but it was at 106%:

<img width="130" height="59" alt="Image" src="https://github.com/user-attachments/assets/91768ee0-566d-4233-af04-a0a3dbb6b2fa" />

I have no idea why the browser tab on MacOS started playing. All devices were disconnected from the MacMini.

It is eventually the same as this: https://github.com/advplyr/audiobookshelf-app/issues/1940

Here is a part of the logs:

audiobookshelf-1  | [2026-08-02 17:19:24.185] INFO: [MediaProgress] Marking media progress as finished because time remaining (-6534.170196000065) is less than 10 seconds (media item 38421ea8-14db-4c97-90b2-1f5cf3a1acb1)
audiobookshelf-1  | [2026-08-02 17:19:24.218] DEBUG: [ApiCacheManager] mediaProgress.afterUpdate: cleared user-progress cache slices (personalized=0, recentEpisodes=0, me=0)
audiobookshelf-1  | [2026-08-02 17:19:24.254] DEBUG: [ApiCacheManager] playbackSession.afterBulkUpdate: cleared user-progress cache slices (personalized=0, recentEpisodes=0, me=0)
audiobookshelf-1  | [2026-08-02 17:19:34.321] DEBUG: [PlaybackSessionManager] syncSession "f5f1fefb-956c-4bfa-a1f3-8a79b2913cae" (Device: Mac OS 10.15 / Firefox) | Total Time Listened: 61198
audiobookshelf-1  | [2026-08-02 17:19:34.327] INFO: [MediaProgress] Marking media progress as finished because time remaining (-6544.354187000077) is less than 10 seconds (media item 38421ea8-14db-4c97-90b2-1f5cf3a1acb1)
[...]

Issue #1940, opened 2026-07-25
[Bug]: Stale/duplicate session overwrites correct progress — server log shows two conflicting updates 12 seconds apart


Vikingjay95:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

While listening via the Android app (Android Auto in use), progress for one book was overwritten by what appears to be a stale/duplicate session reporting an old position. Server-side logs show two PlaybackSessionManager.syncLocalSession calls for the same media item 12 seconds apart:

2026-07-24 08:04:18.674 — Updating progress for "Uncapped (Unabridged)" with current time 27599.053 (previously 26704.964)
2026-07-24 08:04:30.299 — Updating progress for "Uncapped (Unabridged)" with current time 1.588 (previously 27599.053)

The first update continues correctly from a prior session (~7.4 hrs in, consistent with a large batch of accumulated progress from an offline/local session finally syncing). The second, arriving 12 seconds later, resets progress to ~1.6 seconds — effectively the start of the book — and this became the final stored value since it landed last.

No prior activity for this book appears in the several minutes before 08:04:18, and immediately afterward (08:04:39) a new Socket.IO connection is established (Socket Connected to /socket.io ...), consistent with the phone reconnecting to the home network at the end of a drive. This suggests two local sessions existed for the same book on the device — one continuing from ~26700s, one starting fresh — and both flushed to the server in the same reconnect window, with the fresh one winning the race.

Meanwhile, a concurrent session for a different book synced cleanly every ~10 seconds throughout the same window with no gaps, confirming the server/network/reverse proxy were functioning normally — this is app-side.

Checked subsequent days' logs — this appears to be a one-off tied to a specific reconnect event, not a nightly recurrence.

Server version: 2.35.1
App: Audiobookshelf Android app, used with Android Auto
Expected behavior: Only the active playback session's progress should be reported; a stale session for the same book should not be able to overwrite more recent progress.
Actual behavior: A stale/fresh session reporting an old position overwrote correct, in-progress listening data.

Possibly related to #1161 ("Mobile app ignores external progress when playing").

### Steps to Reproduce the Issue

1. Listen to an audiobook via the Android app while using Android Auto
2. Drive for an extended period, then finish or leave a book paused
3. End the drive and let the phone reconnect to home WiFi
[...]

Issue #1901, opened 2026-06-19
[Bug]: Certain app functions fail over IPv6


MikeBishop:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

When the app is connected to the server over IPv6 and not IPv4, the app is able to log in and browse the library normally. However, media playback and download fail. If content is already downloaded, progress sync fails.

This appears to be a problem exclusive to the mobile app, since the web app works fine in this situation — playing media and syncing progress function normally. 

### Steps to Reproduce the Issue

Make ABS accessible on an IPv6 address. Access the server by DNS name on a network without IPv4. Attempt to play back media. 

GitHub won't upload the video, so see https://1drv.ms/v/c/8adb072d266550ab/IQCGUv02-u0vTY1PQ-mp8ZBYAVrZ-U-02PoO1Te94M1dJVA

### What was Expected?

When connecting to a server by DNS name, the app should be IP-version agnostic. Playback, download, and media sync should work when the server is accessible, regardless of IP version. 

### Phone Model

OnePlus 11

### Phone OS

Android 16

### Audiobookshelf App Version

Android App - 0.13.0

### Installation Source

Google Play Store

### Additional Notes

Reported on Discord first.

MikeBishop:
This turns out to be a buried misconfiguration. The DNAT rule only applied to native IPv4 or IPv6 traffic, but missed NAT64-translated addresses. The two client paths in ABS result in different IPs being used -- one went native IPv6, the other selected IPv4 over NAT64, and the issue was obscured by #1954.

PR #1957, opened 2026-08-11
Add Android host-JVM test suite (no production changes)

ckbaker10:
## Brief summary

Adds a host-JVM unit test foundation for the Android app. Before this work, the
only tests on `master` were the Capacitor template's `2 + 2 == 4` check and a
stale package-name assertion; CI only assembled the APK.

The suite now contains **560 tests**. It covers the domain models, progress
reconciliation, downloads, authentication and API contracts, connectivity,
local-media cleanup, playback helpers, database persistence, Android Auto browse
logic, and the reachable Capacitor plugin surface without requiring an emulator.

The branch contains no production Kotlin changes. Its seven enabled failures are
regression specifications for three production defects found during the final
test audit. Those fixes remain separate by design; the companion
`fix-model-guards-from-suite-audit` branch turns all seven green without changing
their assertions. The resulting `master` baseline is **560 tests, 0 failures**.

## Which issue is fixed?

None. This PR adds tests and test infrastructure; it does not change app
behaviour.

It provides regression coverage for 29 reported issues:

- Progress conflict: #1945, #1940, #1852, #1516, #1510, #1442, #1416, #1338
- Download integrity and resume: #1838, #1827, #1709, #1479, #1428
- Auth and transient HTTP: #1908, #1900, #1901
- Pause and external control: #1847, #1828, #1491
- Connectivity state: #1702, #1560, #1802
- Local media lifecycle: #1680, #1630, #1392
- Large and invalid media: #1684, #1650, #1731
- Cover-image crash: #1817

The original suite carried 50 enabled regression failures. All 50 are now green
on `master` after their production fixes landed on separate branches, with the
test assertions left untouched. That gives each fix a reproducible target and
leaves a regression check behind.

The final audit exposed three additional defects, represented by seven enabled
failures at the tip of this test branch:

- Sleep-timer values such as `"0600"` parse as hour `600` instead of falling back
  to a valid configured hour.
- `BookMetadata.getAuthorDisplayName` returns `"Unknown"` for the server's plain
  book response because that shape supplies `authors` but not `authorName`.
- A playback position before the first track selects the last track instead of
  the first.

Their production fixes are intentionally not included here. Tests and fixes on
the same branch can move together and produce a false proof.

## Pull Request Type

Android only. Tests, test infrastructure, documentation, Gradle test
configuration, and CI only. No frontend, iOS, or production Kotlin behaviour is
changed.

The test-only Gradle configuration adds:

- JUnit 4, MockK, MockWebServer, and `kotlinx-coroutines-test` dependencies
- `org.json:json` on the test classpath because AGP's mockable `android.jar`
  removes the method bodies used by Capacitor's `JSObject`
- A Java 21 launcher scoped to unit tests because Capacitor classes are compiled
  for Java 21 while the app toolchain remains Java 17
- The three `--add-opens` required by PaperDb's Kryo serializer on JDK 21+
- A JaCoCo report task that depends on a fresh test run, preventing stale
  execution data from being reported as current coverage

`assembleDebug` is unchanged. Android JVM tests now have their own
`android-tests.yml` workflow and run for every `android/**` change instead of
being coupled to APK generation. Test reports are uploaded when the job fails so
the full assertion messages remain available for diagnosis.

### Test design

This is a host-JVM suite only: no Robolectric, emulator, Espresso, or
`src/androidTest` dependency.

- `AbsTestEnvironment` supplies the shared PaperDb sandbox and wrappers for gaps
  in AGP's mockable `android.jar`, including `Base64`, `Environment`,
  `MimeTypeMap`, `Uri`, `Build.*`, and `SDK_INT`.
- `AbsSingletonRule` resets PaperDb and the `DeviceManager`,
  `MediaEventManager`, and `AbsLogger` singletons before and after each test
  class. This replaces hand-copied lifecycle code that four suites had applied
  incorrectly.
- `MockServerRule` and `withMockServer` provide class-scoped and test-scoped
  MockWebServer lifecycles.
- `RecordingDownloadCallback` uses atomics and latches to verify completion
  cardinality without `Thread.sleep` race detectors.
- Known defects remain enabled failures rather than `@Ignore`d tests. Each defect
  spec records exact inputs, the expected and observed result, and the production
  code path in KDoc.
- Characterization tests are labelled explicitly so pinned behaviour is not
  mistaken for a desired contract.

`android/app/src/test/java/com/audiobookshelf/app/TESTING.md` is the working
guide for conventions, harness usage, known host-JVM limitations, the remediation
queue, and the issue-to-suite map. `android/TESTING.md` is the shorter environment
setup and command reference.

### Final audit and hardening

Every one of the previous 503 tests was read against the production path it
names, and the documented failure count was checked against a real run. That
audit corrected the stale 497/46 baseline to the actual 503/50 baseline, then
added 57 tests and strengthened the infrastructure.

The audit fixed several tests that were weaker than their names:

- A 14-row playback-speed table reused one mock without clearing its call
  history, allowing later rows to pass because an earlier row made the same call.
- Three external-pause assertions verified through a nullable safe call, so they
  could pass vacuously if the event emitter was absent.
- A negative-position test accepted either of two track indices while its name
  required the first; tightening it exposed a real production defect.
- A malformed sleep-timer test expected the same value as its fallback default,
  so it could not prove parsing had failed; tightening it exposed another defect.
- No-throw-only, misleading MIME-type, shallow-copy, and stale-path assertions
  were tightened or relabelled.

New coverage added by the audit includes:

- Golden server 2.36.0 response fixtures derived from the server's serializers,
  covering libraries, books, podcasts, users, and media progress. These found the
  missing author fallback immediately.
- Model deserialization checks for unknown fields, numeric coercion, absent
  optionals, and subtype-deduction coupling.
- Both positive and negative VPN classification so a dead VPN is not accepted as
  online merely because it uses `TRANSPORT_VPN`.
- Numeric semantic-version comparisons, HTTP 403 credential preservation, and a
  successful refresh response with an unreadable body.
- Download redirects, mid-transfer 401 responses, oversized resumes, and free
  space checks before chunked bodies.
- Stronger callback-cardinality checks and removal of the test tree's remaining
  compiler warnings.

The lifecycle audit also found the most serious source of suite flakiness:
`MediaManagerTest` shut down its MockWebServer but left the shared
`DeviceManager` singleton pointing at that dead server for whichever class ran
next. The new rules make teardown structural instead of relying on every class to
repeat it correctly. `reset()` also stops leaking roughly 800 temporary
directories per full run.

### Known host-JVM boundaries

The suite documents the Android surfaces it cannot honestly exercise. The most
important is `Handler(Looper.getMainLooper()).post {}`: on the host JVM the body
never runs, leaving 12 of `AbsAudioPlayer`'s 17 plugin methods unreachable.
`MediaProgressSyncer.start()` is also excluded because it creates a non-daemon,
15-second timer thread that its reset and pause paths do not reap, while its
handler-bound body still cannot execute.

These gaps are recorded rather than covered with assertions that would pass
without running production logic.

### Why this matters

Changes to progress sync, downloads, authentication, API handling, persistence,
and connectivity now have an executable regression net. Reviewers can see which
contracts a change affects and CI can catch a regression in the change that
introduces it rather than after a user report.

The red-spec workflow has already demonstrated that value: 50 defect specs went
green through separate production fixes without weakening their assertions, and
the final audit found three more live defects by making existing assertions more
honest and adding real server-shaped fixtures.

## How have you tested this?

From the repository root, with JDK 21, Node 20, Android SDK Platform 35, and
Build-Tools 35.0.0 installed:

```bash
export JAVA_HOME=/usr/lib/jvm/java-21-openjdk-amd64
export PATH="$JAVA_HOME/bin:$PATH"
export ANDROID_HOME="$HOME/Android/Sdk"
export ANDROID_SDK_ROOT="$ANDROID_HOME"
npm ci
./android/gradlew :app:testDebugUnitTest -p android --no-daemon --rerun
./android/gradlew :app:jacocoDebugUnitTestReport -p android --no-daemon
```

Verified results:

- `android-jvm-test-framework` at `b83aeee4`: **560 tests, 7 enabled failures**
- With `fix-model-guards-from-suite-audit` applied (the resulting `master`):
  **560 tests, 0 failures**

The seven branch-tip failures are exactly the three documented model guards
above. The coverage task now runs the test task itself, so it cannot silently
publish numbers from a stale `.exec` file.

## Screenshots

None. No frontend changes.


ckbaker10:
Opening a new PR that merges the fixed with a propper description

nichwall:
Hey, I have not read through this in detail but since this is a very large PR can we keep the test updates separate from code changes?

I think it is unlikely to be merged as it is since smaller fixes which are easier to review will be probably be looked at first, leading to merge conflicts. We reviewed a number of PRs recently to get a new release out before the 90-day Testflight build expired, but the focus is on finishing up the new web client right now.

 (Edit, whoops, posted on wrong PR, copying to the right one)

PR #1958, opened 2026-08-13
Fix: Prevent crashes when decoding corrupted cover images

ant1kvar:
App crashes on Android 9+ with `ImageDecoder$DecodeException` when trying to load corrupted or unsupported cover images.

FATAL EXCEPTION: main
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
Caused by: android.graphics.ImageDecoder$DecodeException: Failed to create image decoder with message 'unimplemented'Input contained an error.

Wrapped all `ImageDecoder.decodeBitmap()` and `MediaStore.Images.Media.getBitmap()` calls in try-catch blocks. Now returns `null` on decode failure and logs the error instead of crashing.

- `PlaybackSession.kt` - `resolveCoverBitmap()`
- `PlayerNotificationService.kt` - `updateMediaSession()`
- `AbMediaDescriptionAdapter.kt` - `getCurrentLargeIcon()`
- `LocalLibraryItem.kt` - `getMediaDescription()`

Tested with corrupted cover images - app no longer crashes, shows placeholder image instead.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/data/LocalLibraryItem.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/data/LocalLibraryItem.kt
@@ -142,11 +142,17 @@ class LocalLibraryItem(
 
     var bitmap:Bitmap? = null
     if (coverContentUrl != null) {
-      bitmap = if (Build.VERSION.SDK_INT < 28) {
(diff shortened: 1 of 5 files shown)


ant1kvar:
On item/player cover views, if a `raw` cover fails to load/decode, fall back to the resized/cached cover URL instead of showing "Invalid Cover"

Issue #1900, opened 2026-06-17
[Bug]: Android app login fails when username contains trailing whitespace


JasonLoCascio-Personal:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

# Android app login fails when username contains trailing whitespace

## Description

The Android app appears to submit usernames with trailing whitespace exactly as entered/autofilled rather than trimming leading/trailing whitespace before authentication.

This results in a generic "Unauthorized" error even when the username and password are otherwise correct.

## Steps to Reproduce

1. Open the Android app.
2. Enter a valid server URL.
3. Enter a valid username with a trailing space (e.g. `jason `).
4. Enter the correct password.
5. Tap Login.

## Actual Behaviour

The app displays:

`Unauthorized`

Server logs show:

```text
[LocalAuth] Failed login attempt for username "jane " (User not found)
```

Note the trailing space in the username.

## Expected Behaviour

The app should either:

1. Trim leading/trailing whitespace from usernames before authentication, or
2. Display a more specific error indicating that the username was not found.

Most login forms trim accidental whitespace automatically, particularly when usernames are entered via mobile keyboards, autofill systems, or password managers.

## Additional Context

In my case:

* The same username/password combination works correctly in the web interface.
* The web interface also works from the browser on the same Android device.
* The Android app only reports "Unauthorized".
* The actual cause was only discoverable by inspecting the Audiobookshelf server logs.

Relevant log entries:

```text
[[LocalAuth] Failed login attempt for username "jane " (User not found)

``

This suggests the authentication backend treats `jane` and `jane ` as different usernames and the Android login flow is not trimming whitespace before submission.


### Steps to Reproduce the Issue

Try to login with your username and an extra space on the end


### What was Expected?

A successful login

### Phone Model

Samsung A7

### Phone OS

Android 10

### Audiobookshelf App Version

Android App - 0.13.0

### Installation Source

Google Play Store

### Additional Notes

_No response_

ShamiaAkterShanta:
Hi, I’d like to work on this issue. I will investigate trimming leading/trailing whitespace from the username before authentication.

Issue #1852, opened 2026-04-19
[Bug]: Desync when Android connects to Server


ReaderGuy42:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

I've only recently started using the Android app for listening. Mostly love it, but I've had a couple instances of desyncing when the Android app connects to the server.

### Steps to Reproduce the Issue

Listen to audiobook offline in the Android app.
Connect back to server.
Occasional desync 

### What was Expected?

I would hope it takes the furthest position in the book to select as the new progress in the book

### Phone Model

Pixel 6a

### Phone OS

Android 16

### Audiobookshelf App Version

iOS App - 0.11.0, Android App - 0.11.0

### Installation Source

Other (List in "Additional Notes")

### Additional Notes

Installed via F-droid, App version 0.12.0-beta

Itadiki:
+1

chattah:
For anyone experiencing this there is a workaround we can use until this issue is resolved.  If you click on the three dots next to the green play button and select history you can click on the timestamp before the server resync which will take you back to where you were before the position was lost.

Itadiki:
+1

ReaderGuy42:
I haven't tried the workaround because I only just saw it and it doesn't work offline. But this bug still occurs in 0.13.0-beta.

C0rish:
I can also confirm this bug.

I’ve been using Audiobookshelf for quite some time now, but I’ve only noticed this issue in the last few weeks.

The error occurs regardless of the playback method I have used (website, app or Android Auto version of the App). Sometimes it works smoothly, while at other times the most recent timestamp seems to be simply ignored.

However, the server logs show the listening history accurately and without errors.

It’s strange that this error didn’t exist a while ago, but it is particularly annoying because switching between chapters (or viewing past timestamps) is more difficult - or even impossible - in Android Auto.

Issue #1847, opened 2026-04-14
[Bug]: Not saving location when paused from notification / lock screen


barelylit:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

I started playing a book when i cycled home yesterday after work, listened to it for nearly an hour, then on arrival clicked pause on the lock screen notification. 

When i started playing again this morning, it played again from where i had started yesterday as if i had not played anything on the way home at all.

### Steps to Reproduce the Issue

Play from the app
Let phone lock
wait a while
Pause from lock screen

<img width="960" height="2142" alt="Image" src="https://github.com/user-attachments/assets/b287e31a-27a7-431f-8730-8e9c2b6f3e8c" />

### What was Expected?

Save the actual play location not where it was prior to listening

### Phone Model

Pixel 9 Pro

### Phone OS

Android 16

### Audiobookshelf App Version

Android App - 0.11.0

### Installation Source

Google Play Store, Other (List in "Additional Notes")

### Additional Notes

Using 0.12.0-beta

PR #1961, opened 2026-08-15
Android host-JVM test suite and the fixes it produced

### LINKED ISSUE #1817: [Bug]: Invalid downloaded cover file causes crash

Yes

### What was the Problem?

I have an audiobook on my server (Doctor Who "The Black Hole" from Humble Bundle), and it shows up with a valid cover image on the server.

However, after downloading this to a local file onto my phone, I somehow ended up with a 0 byte .jpg file as the cover image. This causes the following problems with the audiobookshelf android app

- When trying to play this audio book from local files (not connected to server), the app immediately crashes
- When this file is present, the Android Auto UI for Audiobookshelf just sits there constantly loading, then finally gives a failure message.

I have managed to fix this issue on my device by deleting and downloading local files from server again, which resulted in a valid cover art file. However, the underlying issue of incorrectly handling invalid cover art files still remains.

I guess adding a 0 byte .jpg cover image to local files would reproduce, although I haven't figured out how to do this manually on my own phone.

### What was Expected?

Expected audiobookshelf to play the audio file.

### Phone Model

Pixel 9a

### Phone OS

Android 16 stock, running audiobookshelf  app version 0.12.0-beta

Android App - 0.11.0

### Installation Source

Google Play Store

### Additional Notes

As far as I can tell the issue is distinct from #1143

ckbaker10:
## Scope

Against `advplyr/audiobookshelf-app@master` (base `12025ab5`):

| Area | Files | Diff |
| --- | ---: | --- |
| Tests, fixtures, test docs | 66 | +12,572 / −18 |
| Production Kotlin | 17 | +336 / −92 |
| CI | 1 | +54 |

No frontend, no iOS, no new features. The production diff is 18 bug fixes, each with a test in the
same branch.

Still two things in one PR — the suite, and the fixes. They can be split at commit `297c85b2` if you
want them separate; that commit is tests only and is red by design (see below).

## Brief summary

Adds host-JVM unit tests for the Android app. Previously there were two: the Capacitor template's
`2 + 2 == 4` and a stale package-name assertion.

550 tests, no emulator, ~90s, green.

Working method for every fix here: write the defect as an enabled failing test against the intended
contract, fix it in a separate commit, check the test goes green without touching its assertion. The
last two commits show that directly — `297c85b2` is 550 tests / 7 failures, `00da0b2e` is 550 / 0,
and the assertions are identical in both.

## Which issues are fixed?

Fixes #1817 — invalid downloaded cover images no longer crash playback.

The other fixes address underlying causes across the clusters below. Each contract is kept as a
regression test:

- Progress conflict and reconnect: #1945, #1940, #1852, #1516, #1510, #1442, #1416, #1338,
  #1298, #739
- Download integrity and resume: #1838, #1827, #1709, #1479, #1428
- Auth and transient HTTP: #1908, #1900, #1901
- Pause and external control: #1847, #1828, #1491
- Connectivity state: #1702, #1560, #1802
- Offline local playback with an unusable network: #1068
- Local media lifecycle and stale download records: #1680, #1630, #1392, #902, #595
- Large and invalid media: #1684, #1650, #1731
- Finished-state persistence: #524
- Cover-image and shared-storage cover crashes: #1817, #1143

36 issues with fixes or regression coverage; 33 currently open upstream. The three that aren't are
#1838 and #1479 (completed) and #1901 (closed as not planned).

Three more had no issue filed and were found while writing the tests:

- Auto sleep timer never fires if the time string has no separator. `"0600".split(":")` is
  `["0600"]`, which parses fine as the integer 600, so the existing fallback never ran. Needed a
  range check, not a parse check. `SleepTimerManager` compares the current hour to 600 and never
  matches.
- Author renders as "Unknown" for valid server responses. `getAuthorDisplayName()` only read the
  flat `authorName`, which the server sends in its minified and expanded serializers but not in
  plain `toOldJSON()` — that one sends the `authors` collection. Hits Android Auto labels and media
  descriptions.
- A playback position before the first track selects the last track. `getCurrentTrackIndex()` fell
  through to `size - 1` unconditionally, which is right past the end and wrong before the start.
  `getNextTrackIndex()` returns 0 for the same input.

## Pull Request Type

Android only.

- Tests, harness, fixtures, docs — most of the diff.
- Production Kotlin, 17 files: progress reconciliation, downloads, auth, connectivity
  classification, Android Auto browsing, domain models. Bug fixes only.
- CI: `android-tests.yml`.

Test config additions, each load-bearing:

- JUnit 4, MockK, MockWebServer, `kotlinx-coroutines-test`.
- Real `org.json` on the test classpath. AGP's mockable `android.jar` strips the methods Capacitor's
  `JSObject` needs, so `JSONObject` silently returns nulls instead of parsing.
- Java 21 launcher for unit tests only. Capacitor classes are compiled for 21, the app toolchain is
  17. Scoped to test tasks, so `assembleDebug` is unaffected.
- Three `--add-opens`; PaperDb's Kryo serializer reflects into the JDK on 21+.
- JaCoCo report task that depends on a fresh test run.

### Covered

Progress reconciliation, playback sessions, track/chapter boundaries. Download integrity, resume,
queue state, incomplete-download cleanup. Auth failures, credential preservation, request
construction, the app/server contract. Connectivity classification including VPNs and captive
portals. PaperDb persistence and cleanup. Domain-model parsing and deserialization. Media controls,
external pause, browse tree, Cast helpers, and the Capacitor plugin methods that are reachable.

Contract tests include golden fixtures derived from the server's 2.36.0 serializers. They carry
every field the server emits, including ones the client doesn't model, so they catch response-shape
changes that a hand-written fixture can't — a hand-written fixture is written to match the model, so
it can't disagree with it. That's how the author bug above turned up.

### Harness

Host-JVM Android tests have known traps: framework methods return null, singletons leak between
classes, PaperDb caches statically, async callbacks let weak tests pass by accident.

- `AbsTestEnvironment` — isolated PaperDb dir per test, resets shared state, wraps the known gaps in
  the mockable jar.
- `AbsSingletonRule` — setup/teardown around every participating class, so one suite can't leave a
  dead MockWebServer or stale singleton for whatever class the shared Gradle JVM runs next.
- `MockServerRule` / `withMockServer` — class-scoped and test-scoped server lifecycles.
- `RecordingDownloadCallback` — atomics and latches rather than sleeps, so duplicate completion
  callbacks are actually detected.

`android/app/src/test/java/com/audiobookshelf/app/TESTING.md` has the conventions, harness notes,
host-JVM boundaries, issue map and maintenance rules. `android/TESTING.md` is setup and commands.

### Audit

All 503 tests that existed at the time were read against the production path they name. That added
47 cases and fixed assertions that could pass for the wrong reason:

- A 14-row playback-speed table where later rows were satisfied by calls recorded in earlier rows
  (one mock, no `clearMocks`, at-least-once `verify`).
- `verify` through a safe call, which records nothing and passes if the property is null.
- A track-index assertion accepting two answers under a name requiring one.
- A malformed-time assertion expecting the same value as the fallback, so it couldn't tell parsing
  from failure.
- Fixed sleeps and non-atomic counters for duplicate-callback detection.

Two of the three unreported defects came out of tightening those; the third came from the golden
fixtures. Both were sitting behind assertions that were passing.

### Not covered

`Handler(Looper.getMainLooper()).post {}` doesn't run its body on a host JVM, so 12 of
`AbsAudioPlayer`'s 17 plugin methods are unreachable. `MediaProgressSyncer.start()` spawns a
non-daemon timer thread and its body is handler-bound. `PlayerNotificationService` and `CastPlayer`
are out of reach entirely.

These are documented rather than covered with no-throw assertions that would pass without running
anything.

## How have you tested this?

JDK 21, Node 20, Android SDK Platform 35, Build-Tools 35.0.0:

```bash
export JAVA_HOME=/usr/lib/jvm/java-21-openjdk-amd64
export PATH="$JAVA_HOME/bin:$PATH"
export ANDROID_HOME="$HOME/Android/Sdk"
export ANDROID_SDK_ROOT="$ANDROID_HOME"
npm ci
./android/gradlew :app:testDebugUnitTest -p android --no-daemon --rerun
```

```text
BUILD SUCCESSFUL
550 tests completed
```

Any failure is a regression. A known defect is an enabled failing test here, not `@Ignore`, so if
the suite is red, `TESTING.md` §7.1 says what's outstanding.

Coverage:

```bash
./android/gradlew :app:jacocoDebugUnitTestReport -p android --no-daemon
```

Depends on the test task, so it won't report off a stale `.exec`.

## Screenshots

None. No UI changes.


nichwall:
Hey, I have not read through this in detail but since this is a very large PR can we keep the test updates separate from code changes?

I think it is unlikely to be merged as it is since smaller fixes which are easier to review will be probably be looked at first, leading to merge conflicts. We reviewed a number of PRs recently to get a new release out before the 90-day Testflight build expired, but the focus is on finishing up the new web client right now.

nichwall:
Also just commenting so you're aware since I saw two more large commits since my previous comment, we are also not reviewing AI generated PRs because they are so big and tend to be complicated. We may eventually get to it and having a better test harness would be great, but as of right now we are not reviewing them because the tend to be very large and have a lot of churn.

https://audiobookshelf.org/docs/contributing/general/#ai-tools

ckbaker10:
@nichwall 

AI is the perfect tool to add test coverage as it's a task that is mainly ignored by people which are driven by feature creep and not stability. As is proven by the state of testing in the apps repository.

The few fixes for the found issues actually already provide a great deal of stability to the app, it's the files under main/java/com/audiobookshelf/app the main code update is only test cases. 

I won't debate your stance, you should just get used to AI as a tool as it will not vanish from our code ecosystem anymore.

I'm aware of your critique and do assure you I actually proofread what was generated here.

nichwall:
Yes, I agree AI is a useful tool and we definitely need to improve test coverage. I'm just letting you know it will likely be a while before this is reviewed, especially if there are additional commits. I appreciate that you have read through it, this was more of an FYI because we have been getting an increase of low quality PRs (usually because of AI, but it applies to human generated too) so we are much more hesitant to review them since they are much better appearing good at a glance.

But yes, don't want to debate it here, thanks for working on the test coverage.

My biggest concern is to break up code changes from adding a test harness because they are both large changes.

PR #1964, opened 2026-08-19
Add miniplayer display settings

### LINKED ISSUE #1950: [Enhancement]: Allow disabling of book title and cover art in the miniplayer via settings

It would be nice to have a setting to disable the display of the book title and cover art in the scrolling display at the bottom of the application when playing back a book

Image

This would eliminate the duplication of information that is already displayed on the screen and allow for more of the chapter title to be displayed. 

I don't have a mock-up for the settings screen

Android App - 0.13.0

There is no setting currently in settings.

Dajiben666:
## Brief summary

Adds settings to show or hide the book cover and title/author in the miniplayer.

## Which issue is fixed?

Fixes #1950

## In-depth Description

This PR adds two playback settings:
- Show cover in miniplayer
- Show title in miniplayer

Both settings are enabled by default to preserve the current behavior.

The settings are stored in device settings and supported on both Android and iOS. The fullscreen player continues to show the cover and title regardless of the miniplayer settings.

## How have you tested this?

Tested locally on Android.

Steps:
1. Open Settings → Playback Settings.
2. Disable “Show cover in miniplayer” and verify the cover is hidden.
3. Disable “Show title in miniplayer” and verify the title/author is hidden.
4. Re-enable both settings and verify they are shown again.
5. Verify the fullscreen player still shows the cover and title.
6. Restart the app and verify the settings are preserved.

## Screenshots

Before:
Screenshot_20260819_172026

After:
Screenshot_20260819_201212
Screenshot_20260819_195803
Screenshot_20260819_195812
Screenshot_20260819_195836


kuci-JK:
Looking at the screenshots I have the following suggestion:
- when there is no cover image the title / chapter should be left aligned without the "space" as it would allow for larger portion of the title / chapter to be visible (and that is what the #1950 is about)

Can you add screenshots how it looks like when there is actually the chapter title and/or the author filled for the book?

In the last screenshot seems to be unnecessary space there with both options turned off. When there will be the chapter title and the author I think it will look fine. My first thought in this case was that it would be better to make the minimized player shorter to show more of the page content, but now I think that is unnecessary to introduce different layout for this specific case (both off, no author, no chapter info filled).

PS: I am not really the target user as I like to have both the cover and title there :)

PR #1968, opened 2026-08-23
Fix Android background playback stops caused by WebView bridge traffic

### LINKED ISSUE #1530: [Bug]: 0.9.79 will randomly stop playing content.

Yes

### What was the Problem?

When listening to audiobooks, the player will randomly stop playing. I have verified the app it set to never sleep on Android, I have cleared the cache, and I have uninstalled and reinstalled. 

I am currently running 0.9.77 because the issue does not happen with that version. 

Choose audiobook. 
Press play. 
Wait. 

Unable to discern any pattern. 

### What was Expected?

Played the audiobook without stopping at random. 

### Phone Model

Samsung S22+

### Phone OS

Android 14

Android App - 0.9.79

### Installation Source

Google Play Store

### Additional Notes

_No response_

gubenkoved:
## Brief summary

Reduce WebView/native bridge traffic while the Android app is backgrounded, preventing Android from killing the application during playback because of excessive cached-process Binder traffic.

## Which issue is fixed?

Fixes #1530

## Pull Request Type

Android only. This changes the frontend player lifecycle and Android native bridge event delivery.

## In-depth Description

Process exit diagnostics captured on a Pixel 10 Pro running Android 17 showed the application being killed with:

```text
reason=9 (EXCESSIVE RESOURCE USAGE)
description=excessive binder traffic during cached
```

The app already avoids sending several frequent native display events to a backgrounded WebView. Two recurring sources of bridge traffic remained:

1. `AudioPlayer.vue` continued calling native `getCurrentTime()` once per second while the WebView was hidden.
2. Every 15-second progress sync could still send sync-status events and a serialized, growing media-history object to the hidden WebView.

This change:

- Stops the one-second player UI polling interval when the app loses visibility.
- Restarts polling when the app becomes visible again and playback is active.
- Adds a visibility guard inside the interval to handle lifecycle races.
- Suppresses recurring progress-sync status and media-history display events while the native plugin is backgrounded.

Native playback and progress synchronization continue normally in the background. Only UI updates that cannot be displayed are skipped.

## How have you tested this?

Build verification following the repository's Android workflow:

- `npm run generate`
- `npx cap sync android`
- `gradlew.bat assembleDebug --no-daemon`

Device verification on a Pixel 10 Pro running Android 17:

1. Installed the debug APK and started streamed audiobook playback.
2. Confirmed that the visible player called `getCurrentTime()` once per second.
3. Sent the app to the background and monitored Logcat for 45 seconds.
4. Confirmed that no further `getCurrentTime()` bridge calls occurred after the visibility transition.
5. Confirmed that native progress synchronization continued at 15-second intervals without emitting the suppressed WebView events.
6. Confirmed that the media session remained in the `PLAYING` state and `PlayerNotificationService` remained a foreground service.
7. Returned to the app and confirmed that one-second player UI polling resumed immediately.
8. Paused playback after testing.

## Screenshots

Not applicable; there are no visual changes.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/player/PlayerNotificationService.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/player/PlayerNotificationService.kt
@@ -1069,6 +1069,8 @@ class PlayerNotificationService : MediaBrowserServiceCompat() {
     DeviceManager.widgetUpdater?.onPlayerClosed()
     stopForeground(Service.STOP_FOREGROUND_REMOVE)
     stopSelf()
+    PlayerNotificationListener.isForegroundService = false
+    isStarted = false
   }
 
   fun sendClientMetadata(playerState: PlayerState) {
--- a/android/app/src/main/java/com/audiobookshelf/app/plugins/AbsAudioPlayer.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/plugins/AbsAudioPlayer.kt
@@ -96,10 +96,12 @@ class AbsAudioPlayer : Plugin() {
         }
 
         override fun onProgressSyncFailing() {
+          if (!isInForeground) return
           emit("onProgressSyncFailing", "")
         }
 
         override fun onProgressSyncSuccess() {
+          if (!isInForeground) return
           emit("onProgressSyncSuccess", "")
         }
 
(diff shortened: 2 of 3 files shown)


nichwall:
Overall this looks good at a glance, but I think it is missing a few UI updates when returning from the background.
- Sync success: if the sync success state changes while in the background, it is not updated when returning to the foreground.
- History: The history UI would not update when returning from background. History events are still tracked by the native code, but to see the updates the History page would need to be closed and opened again.

I have not tested these changes, these are just my initial comments if you can test both of those cases as well to make sure they work.

gubenkoved:
Thank you @nichwall! I have updated the PR with another commit here.

As to testing these changes, I can only say that these changes fixes intermittent stops for me completely over the hours of use with experimental build on my Pixel 10 device.

PR #1972, opened 2026-08-24
Fix Stage Manager resize issue on iPadOS (#1563)

### LINKED ISSUE #1563: [Bug]: Unresponsive UI layout of mobile app on iPadOS

Yes

### What was the Problem?

In Audiobookshelf mobile app v0.9.81-beta, the UI is not adaptable to app resizing on my iPad. It works perfectly in fullscreen mode but loses its dynamic layout when the window is shrunk in Stage Manager mode.

1. Set iPad to Stage Manager mode
2. Open Audiobookshelf mobile app
3. Play a chapter in fullscreen mode
4. Resize the app window, the layout breaks

### What was Expected?

The UI should be responsive to window size, like the previous version.

### Phone Model

iPad Pro 12.9in Gen 3

### Phone OS

PadOS v18.4.1

iOS App - 0.9.79

### Installation Source

Testflight

### Additional Notes

Audiobookshelf iOS app v0.9.81-beta
Server on Docker - macOS

Michael-128:
## Brief summary

The app would not resize properly in iPadOS Stage Manager - content would get cut off. It turned out to be a `@capacitor/status-bar` related issue ([details](https://github.com/ionic-team/capacitor-plugins/issues/2337)). The fix was simply to update the `@capacitor/status-bar` package from `^7.0.0` to `^7.0.6`.

## Which issue is fixed?

Fixes #1563

## Pull Request Type

**Does this affect only Android, only iOS, or both?**
Both platforms get a dependency update, but the fix is aimed specifically at iPadOS.

**Does this change the frontend or the backend of the apps?**
Frontend (dependency update).

## In-depth Description

**How does it work?**
Upgrades `@capacitor/status-bar` to `^7.0.6`, which is a version that includes the resize fix.

**Why is this the best solution?**
It's the best solution, because the package is now at the latest version (for v7), and it doesn't require any code changes.

**Does it solve a problem that affects multiple users or is this an edge case for your setup?**
It fixes #1563 (and my own setup too).

## How have you tested this?

Yes, I've tested this on an iPadOS simulator and my own iPad. Further testing might be required for Android.
1. Compile this branch & Run the app.
2. Enter Stage Manager.
3. Resize window.
4. After some delay, it will resize (previously it wouldn't resize at all).

> **Note**: There also seems to be a delay when resizing. However, I feel like it's out-of-scope of this PR.

### Before
https://github.com/user-attachments/assets/e8442580-95b6-4bba-bd5f-7d2c00683ddd
### After
https://github.com/user-attachments/assets/fd394990-29ba-4031-9497-fb3135ca9b07

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
         "@capacitor/ios": "^7.0.0",
         "@capacitor/network": "^7.0.0",
         "@capacitor/preferences": "^7.0.0",
-        "@capacitor/status-bar": "^7.0.0",
+        "@capacitor/status-bar": "^7.0.6",
         "@nuxtjs/axios": "^5.13.6",
         "@teckel/vue-pdf": "^4.3.5",
         "@webnativellc/capacitor-filesharer": "7.0.4",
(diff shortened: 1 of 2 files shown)


advplyr:
Thanks!

PR #1975, opened 2026-08-25
Android TV (5/10): navigation engine kit (dormant library)

bilbospocketses:

Heads up: the rest of the series is now open. It was originally paced in waves, each opening after the previous merged, but that just kept the work invisible — so PRs 5-10 and two small companion PRs are all up now. Review and merge in whatever order suits you; the dependency graph is in the plan.

### Scope
The D-pad navigation engine as a **library only** — 17 modules under `plugins/tv/` that export functions and attach to nothing at runtime. Nine utility modules (focus memory, spatial navigation, scroll helpers, visibility, overlay focus, focus entry, focus color, selectors, shared context) and eight per-page handlers. Every file is under 235 LOC.

### Safety
Nothing registers a listener, installs a router hook, or touches the store. This is dead code on every device until the integration PR wires it up — merging it alone changes no behavior anywhere. The split exists so the function library can be reviewed separately from the wiring that turns it on.

### Testing
Inert on its own, so verified by build: `npm run generate` green. The modules are exercised end-to-end by the integration PR and verified on a Google TV Streamer 4K across a 13-batch manual smoke suite.

Depends on the `isAndroidTv` flag from #1892. Required by the integration PR.

### DIFF
--- a/docs/pr-05-engine-kit.md
+++ b/docs/pr-05-engine-kit.md
@@ -1,0 +1,42 @@
+# PR 05 — TV navigation engine kit (utility modules + page handlers)
+
+Part of a 10-PR series adding Android TV support to audiobookshelf-app,
+replacing the original PR #1843.
+
+## Full series plan
+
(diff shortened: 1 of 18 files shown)


bilbospocketses:

PR #1976, opened 2026-08-25
Android TV (6/10): activate the navigation engine

bilbospocketses:

### Scope
Activates the engine kit. Two new files under `plugins/tv/` plus small edits to two existing files:
- `listeners.js` — global keydown listener and the router/store/eventBus subscribers that drive focus.
- `index.js` — plugin entry point; returns immediately unless `isAndroidTv` is set.
- `nuxt.config.js` — registers the plugin (client mode).
- `store/index.js` — adds the `resetLastBookshelfScrollData` mutation that `listeners.js` commits on library switch. It lives here rather than in #1892 because its only caller is the engine.

This is where D-pad navigation actually starts working on a TV.

### Stacked PR
**This branch is stacked on #1975**, so until that merges the diff here also shows its 17 modules. Merge the kit first and this reduces to the four files above.

### Safety
`index.js` no-ops on anything that is not an Android TV, so phone and tablet behavior is byte-identical to upstream — nothing is installed and no listener is registered.

### Testing
GTV Streamer 4K: 13-batch manual smoke suite (library grid, item detail, podcast episodes, author pages, logs, stats, settings, player). Phone: plugin no-ops, navigation unchanged. `npm run generate` green standalone.

Depends on #1975 (stacked) and the `isAndroidTv` flag from #1892.

### DIFF
--- a/docs/pr-05-engine-kit.md
+++ b/docs/pr-05-engine-kit.md
@@ -1,0 +1,42 @@
+# PR 05 — TV navigation engine kit (utility modules + page handlers)
+
+Part of a 10-PR series adding Android TV support to audiobookshelf-app,
+replacing the original PR #1843.
+
+## Full series plan
+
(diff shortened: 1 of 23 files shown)


bilbospocketses:

PR #1977, opened 2026-08-25
Android TV (7/10): audio player behavior on TV

bilbospocketses:

### Scope
Four TV-gated behavior changes in `components/app/AudioPlayer.vue`:
- **Close instead of minimize** — on TV, leaving fullscreen closes the player rather than collapsing to the mini bar, which is very hard to get back to with a D-pad (no taskbar or tray to reach it from). A user setting to choose between the two is planned.
- **Keep-awake during playback** — holds a screen-wake lock while audio plays so Ambient Mode does not kick in, which otherwise kills playback on Chromecast with Google TV. Uses `@capacitor-community/keep-awake`, already in `package.json`.
- **Hide History in the player menu** — it routes to a page that collapses fullscreen into the mini-player state TV no longer uses. History stays reachable from book detail pages.
- The wake lock is released when playback closes.

### Safety
Every change is guarded on `isAndroidTv`; phone and tablet behavior is unchanged.

### Testing
GTV Streamer 4K: playback survives an idle period that previously triggered the screensaver; Back and the collapse control both stop playback and close the player; the menu no longer offers History. Phone: minimize, History, and sleep behavior unchanged.

Depends on the `isAndroidTv` flag from #1892 and on the live engine for focus restore after close. Also edits `AudioPlayer.vue`, which #1893 touches in different regions — whichever merges second may need a trivial rebase.

### DIFF
--- a/components/app/AudioPlayer.vue
+++ b/components/app/AudioPlayer.vue
@@ -133,6 +133,7 @@
 import { Capacitor } from '@capacitor/core'
 import { AbsAudioPlayer } from '@/plugins/capacitor'
 import { Dialog } from '@capacitor/dialog'
+import { KeepAwake } from '@capacitor-community/keep-awake'
 import { getAverageColorFromCoverUrl } from '@/utils/coverAverageColor'
 import WrappingMarquee from '@/assets/WrappingMarquee.js'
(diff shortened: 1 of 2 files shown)


bilbospocketses:

PR #1895, opened 2026-06-06
Android TV (4/10): CSS focus-ring foundation

bilbospocketses:
Part of the Android TV support work, replacing #1843 (split into smaller PRs per the review-volume feedback).

### Scope
The TV focus-ring CSS foundation: `assets/css/tv-focus.css` + the `--tv-focus-color` custom property, a `tvFocusColor` user setting (default `#1ad691`), nuxt.config registration, and a `tv-focus-dots` hook on the loading indicator. All scoped under `.android-tv`.

### Safety
`.android-tv`-scoped → completely inert on phone/tablet. Nothing drives the focus rings yet, so there's no visible change.

### Testing
Phone: no visual change. GTV Streamer 4K: `--tv-focus-color` present on `:root.android-tv`.

Independent and reviewable on its own. A later settings PR adds a color picker that writes `--tv-focus-color`.

PR #1978, opened 2026-08-25
Android TV (8/10): TV settings section + focus-ring color picker

bilbospocketses:

### Scope
A "TV Settings" section at the top of `pages/settings.vue`, shown only on Android TV, containing one control: a focus-ring color picker.
- `components/ui/TvFocusColorPicker.vue` — new D-pad-navigable swatch row with a default marker and a white-outline selection state.
- `pages/settings.vue` — the TV-only section plus `isAndroidTv` / `tvFocusColor` computeds and a `setTvFocusColor` method dispatching `user/updateUserSettings`.

The chosen color feeds the `--tv-focus-color` CSS variable that `tv-focus.css` (#1895) consumes for every focus ring. Focus rings are the only way to tell where you are on a TV, and the default green does not read well against every cover or theme — hence making it configurable.

### On i18n
The section header and row label are hardcoded English rather than routed through `$strings`, to keep the translation catalog out of a TV PR. Happy to move them into `strings/en-us.json` before merge if you would prefer that.

### Safety
The section is behind `v-if="isAndroidTv"`, and the `mt-10` spacer on the next header is TV-gated, so the phone layout is byte-identical.

### Testing
GTV Streamer 4K: section appears only on TV, every swatch reachable by D-pad, selection applies immediately and survives a restart. Phone: settings page unchanged.

Depends on the `tvFocusColor` key and `tv-focus.css` from #1895, and on the live engine. Also edits `pages/settings.vue`, which #1893 touches in different regions — whichever merges second may need a trivial rebase.

### DIFF
--- a/components/ui/TvFocusColorPicker.vue
+++ b/components/ui/TvFocusColorPicker.vue
@@ -1,0 +1,84 @@
+<template>
+  <div class="tv-focus-color-picker flex items-center gap-3">
+    <button
+      v-for="preset in PRESETS"
+      :key="preset.hex"
+      type="button"
+      tabindex="0"
+      class="swatch"
+      :style="{ backgroundColor: preset.hex }"
+      :aria-label="`Focus ring color: ${preset.name}${preset.hex === value ? ' (selected)' : ''}`"
+      @click="select(preset.hex)"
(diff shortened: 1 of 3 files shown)


bilbospocketses:

PR #1979, opened 2026-08-25
Android TV (9/10): server connect form D-pad navigation

bilbospocketses:

### Scope
D-pad navigation for `components/connection/ServerConnectForm.vue`. This is the first screen a new user sees, and it is unusually demanding on a TV: it mixes text inputs, a saved-server list, and action buttons, and it is reached before any library UI exists, so the generic page handlers have nothing to work with.

One cohesive feature rather than separate tweaks:
- Directional movement between the URL/username/password inputs, saved server entries, and submit/cancel.
- Auto-focus with the on-screen keyboard on every form transition and on error recovery, so focus is never left nowhere.
- Username labels above each saved server entry, for fast user switching on a shared TV.
- Server list below the logo in a scrollable column.

### Safety
TV-gated; phone and tablet rendering and behavior are unchanged.

### Testing
GTV Streamer 4K: fresh install through to a connected library using only the remote, including a deliberate bad-URL error and recovery, and switching between two saved users. Phone: connect form unchanged.

Depends on the keyboard hygiene in #1893 and on the live engine.

### DIFF
--- a/components/connection/ServerConnectForm.vue
+++ b/components/connection/ServerConnectForm.vue
@@ -1,17 +1,18 @@
 <template>
   <div class="w-full max-w-md mx-auto px-2 sm:px-4 lg:px-8 z-10">
-    <div v-show="!loggedIn" class="mt-8 bg-primary overflow-hidden shadow rounded-lg px-4 py-6 w-full">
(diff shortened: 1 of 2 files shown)


bilbospocketses:

PR #1980, opened 2026-08-25
Android TV (10/10): author detail page

bilbospocketses:

### Scope
Adds `pages/author/_id.vue`, an author detail page listing that author's books, and makes author cards navigate to it. Also a `tabindex` on `AuthorCard.vue` so a D-pad can reach it, a one-line layout tweak in `AuthorImage.vue`, and routing in `pages/bookshelf/authors.vue`.

### A note on scope — your call
**This one is arguably not TV-specific.** The app has no author detail page on any platform today: tapping an author on a phone does not open one either. It was built for TV because browsing by author with a remote is common, but mobile users would get the same feature, and it is not gated on `isAndroidTv`.

Flagging it explicitly rather than slipping it in: happy for this to ship as-is, be reworked as a general non-TV feature, or be dropped from the series entirely. Nothing in PRs 1-9 depends on it.

### Testing
GTV Streamer 4K: author cards reachable, open the detail page, item grid navigable, Back returns to the author bookshelf with focus restored to the originating card. Phone: the new page renders correctly when reached by tap.

Depends on the keyboard hygiene in #1893 for the shared card components.

### DIFF
--- a/components/cards/AuthorCard.vue
+++ b/components/cards/AuthorCard.vue
@@ -1,6 +1,6 @@
 <template>
-  <div>
-    <div :style="{ width: width + 'px', height: height + 'px' }" class="bg-primary box-shadow-book rounded-md relative overflow-hidden">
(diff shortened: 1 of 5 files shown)


bilbospocketses:

PR #1981, opened 2026-08-25
Android: back-button handling (drawer dismiss + TV exit when logged out)

bilbospocketses:
A small companion PR alongside the Android TV series replacing #1843 — independent of the series except where noted.

### Scope
Android back-button handling in `plugins/init.client.js`, around 12 LOC:
- **All Android:** pressing Back while the side drawer is open dismisses the drawer instead of navigating away. This is standard Android behavior and applies to phones and tablets too, not just TV.
- **Android TV only:** pressing Back on the Home screen while logged out exits the app, rather than leaving the user on a dead screen with nowhere to go.

### Safety
The drawer-dismiss half is a general Android improvement and is not TV-gated. The exit-when-logged-out half is gated on TV.

### Testing
GTV Streamer 4K and an Android phone: the drawer dismisses on Back on both; TV exits cleanly from logged-out Home; normal back-navigation elsewhere is unaffected.

The TV-exit branch depends on the `android-tv` class from #1892. The drawer-dismiss part is fully independent and can merge on its own.

### DIFF
--- a/docs/pr-back-button.md
+++ b/docs/pr-back-button.md
@@ -1,0 +1,28 @@
+# PR — Android back-button handling (drawer dismiss + TV exit)
+
+A small related PR alongside the 10-PR Android TV series (see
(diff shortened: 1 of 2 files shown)


bilbospocketses:

Issue #1970, opened 2026-08-23
[Bug]: Downloads are broken for 0.14.0-beta


kuci-JK:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

The download of the media is broken as of mobile app version 0.14.0-beta.

The download starts, downloads all files, the UI shows download complete processing, then the message disappears but the download button is still there and nothing happens. Clicking on the button again causes red message with download already started message to show up. The app occupies quite a bit of storage after this failed download. Server logs shows invalid range requests.

### Steps to Reproduce the Issue

1. Go to the book detail page
2. Click download button
3. Notification shows up, wait (kept the screen awake to make sure its not issue with android stopping the app)
4. After the download completes the UI shows "Download complete. Processing" (dont remember the exact wording)
5. The Processing message box goes away and nothing happens (the download button does not disappear)
6. Clicking on download again causes the red "Download already started for this media entity" to show up and that is it.

MP3: In the downloading UI (the queue) the items do not disappear as they reach 100%, they seem to get stuck at 100% (not sure how it was before as I usually download m4b audiobooks). The behavior looks the same to me for both mp3 and m4b.

When I then kill the app and open it again there is the circle with items in queue, but no new notification shows up, the items are all at 100%.

No green check mark showed next to the completed items in the queue.

### What was Expected?

I expected the app do download and correctly process the file as in older version.

### Phone Model

Redmi Note 11

### Phone OS

Android 13, Xiaomi HyperOS 1.0.1.0.TGKEUXM

### Audiobookshelf App Version

Android App - 0.14.0-beta

### Installation Source

Google Play Store

### Additional Notes

Might be related #1963 as I encountered similar logs

Downloading to internal app storage.

After upgrading to the new android app version I had issue that the download wouldn't event start, I would just click download, confirm the dialog and nothing would happen, no notification would show. Clicking again triggered download already started. Fixed that by clearing all app data. But the download now does not complete.

Tried reinstalling the app, did not help.

Downloading fails for both m4b and mp3 audiobooks.

Downloaded data stays in the app storage as the app now takes 0.91 GB after my test.

In the server log what looks interesting to me that there is info about finished download followed by the invalid range requests.
<details><summary>Server Log</summary>
<code>

[...]

Issue #1831, opened 2026-03-27
[Bug]: Custom headers/tokens for Cloudflare Zero Trust are commented out. Patch with fix attached.


PandaUnit-TWL:
# Patch: Wire custom headers through all three HTTP stacks (fixes Cloudflare Zero Trust / reverse proxy auth)

**Re: #254** — Custom header support for Cloudflare tunnels

## Summary

The custom headers feature is already implemented in the ABS app codebase — `ServerConnectionConfig` has a `customHeaders` field (`DeviceClasses.kt` line 36–47), `CustomHeadersModal.vue` exists and works, and the data persists correctly. However, the feature was never fully wired up:

1. The UI button to access Custom Headers was never exposed on the server connect form
2. The native Android HTTP layer (`ApiHandler.kt`) never read `customHeaders` from the config
3. The audio player (`PlayerNotificationService.kt`) never sent custom headers on streaming requests
4. Several JS-layer requests (`/status`, `/ping`, OAuth, token refresh) didn't pass headers through

This meant that even if you manually set custom headers, they were silently dropped on most requests — making the app unusable behind header-authenticated reverse proxies like Cloudflare Zero Trust.

## What the patch does

**4 files changed, 33 insertions, 15 deletions.**

### 1. `components/connection/ServerConnectForm.vue`
- Adds a **"Custom Headers"** link on the server address entry screen (calls the existing `addCustomHeaders()` method → opens existing `CustomHeadersModal.vue`)
- `getServerAddressStatus()` now passes `this.serverConfig.customHeaders` to `getRequest()`
- `connectToServer()` now passes `config.customHeaders` to `pingServerAddress()`
- `oauthRequest()` now includes `this.serverConfig.customHeaders` in the `CapacitorHttp.get()` call

### 2. `plugins/nativeHttp.js`
- `request()` now spreads `serverConnectionConfig.customHeaders` into headers for all post-login API calls
- `refreshAccessToken()` signature changed from `(refreshToken, serverAddress)` to `(refreshToken, serverConnectionConfig)` so custom headers are included on token refresh
- `handleTokenRefresh()` updated to pass the full `serverConnectionConfig` instead of just `.address`

### 3. `android/app/src/main/java/com/audiobookshelf/app/server/ApiHandler.kt`
- `getRequest()` — reads `customHeaders` from config or `DeviceManager.serverConnectionConfig`, applies via `builder.addHeader()` loop
[...]

PR #1998, opened 2026-08-26
Fix ~1 minute Android startup block with many downloaded files

### LINKED ISSUE #1997: [Bug]: v0.14.0-beta startup blocked for ~1 minute with many downloaded files

Yes

### What was the Problem?

Since updating to v0.14.0-beta the Android app takes about a minute to start when there are many downloaded files. The main thread is blocked during `Activity.onCreate`, so the system logs `Activity pause timeout` / `Activity stop timeout` and `am start -W` times out. Nothing is rendered until the block ends.

Measured on a library with 2851 downloaded files (12 GB, per-chapter mp3s in a SAF folder on `primary:Audiobooks`):

```
20:43:23.642  Capacitor: Starting BridgeActivity
20:43:23.792  DeviceManager: Device Manager Singleton invoked
20:43:33.936  GC freed 37MB
20:44:18.234  AbsLogger: load: AbsLogger plugin initialized
20:44:20.018  mounted: initializing first load
```

54.4 s in one synchronous block, ~59 s until the UI is usable, on every cold start.

**Root cause.** Two Capacitor plugins do heavy I/O in `load()`, which Capacitor calls synchronously on the main thread from `BridgeActivity.onCreate`. Both code paths come from the download rewrite in `eb2483d0` (merged via #1937, first released in v0.14.0-beta).

The numbers below are from a debug build on a second device with 1557 downloaded files, where Capacitor's `Registering plugin instance:` lines allow attributing the block to individual plugins. Absolute values are higher than in a release build; the ratios are the point.

| Phase | v0.14.0-beta |
| --- | --- |
| `AbsDownloader.load()` | 16.6 s |
| `AbsDatabase.load()` | 58.1 s |
| **total block** | **75.4 s** |

*1. `AbsDatabase.load()` — one storage-provider round-trip per file.* `cleanLocalLibraryItems` checks every downloaded file. Up to v0.13.0-beta that was `File(localFile.absolutePath).exists()`. v0.14.0-beta replaced it with `LocalFile.exists(ctx)`, which for `content://` urls opens a real file descriptor:

```kotlin
ctx.contentResolver.openFileDescriptor(Uri.parse(contentUrl), "r")?.use { true }
```

That is one ContentResolver/Binder round-trip per file. During the block `com.android.externalstorage` sits at ~19% CPU (0 ticks when the app is idle) while the app's main thread waits on binder. For 1557 files these checks alone take 53.8 s. The motivation for the change is sound — `File.exists()` is unreliable under scoped storage — but the per-file cost went from microseconds to tens of milliseconds, on the main thread, at every start.

*2. `AbsDownloader.load()` — queue restore on the main thread.* `DownloadServiceHost.ensure()` deserializes the download db and probes shared storage with `DocumentFile.findFile()` (which lists the whole directory per lookup). The persisted items are read twice: once in `DownloadItemManager`'s constructor via `IncompleteDownloadCleanup.cleanupExpired()`, once in `restoreQueue()` — visible as two GC spikes (40 MB, 19 MB). The db read alone measured 1.5–3.4 s. While this runs, the main thread also cannot attach the bridge, because `setEventEmitter`/`hasWork` wait on the same lock as `restoreQueue`.

*3. Pre-existing, not a regression.* Once the storage access is fixed, the cleanup is dominated by `cleanLocalLibraryItems` scanning the whole local file list for every track, twice — quadratic in the number of files an item has. For a book with 370 chapter files that is ~274k comparisons and 4.2 s; across the library 15.0 s.

1. Download a library with a few thousand files into a folder on shared storage (SAF), e.g. 10 books of per-chapter mp3s.
2. Force-stop the app.
3. Start it and time until the UI appears: `adb shell am start -W -n com.audiobookshelf.app/.MainActivity`.
4. The app is unresponsive for roughly a minute; `logcat` shows the gap between `Starting BridgeActivity` and `load: AbsLogger plugin initialized`.

Happens offline as well, so the server is not involved. Scales with the number of downloaded files, not their size.

### What was Expected?

The app starts in about 2 s, as it did in v0.13.0-beta with the same library.

### Phone Model

Fairphone 4

### Phone OS

Android 15

Android App - 0.14.0-beta

### Installation Source

Other (List in "Additional Notes")

### Additional Notes

Installed from the GitHub release apk.

A fix is ready and tested on the affected device; PR follows and links back to this issue. It moves both cleanups off the main thread, establishes folder access once per item instead of once per file (keeping the full per-document check for when access is not confirmed), and replaces the quadratic track matching with a set lookup. Cold start goes from 75.4 s to 0.3–0.8 s, the cleanup itself from ~69 s to 1.6 s.

While investigating I also hit `PluginLoadException: Unable to load plugin instance` for `AbsDownloader` on v0.14.0-beta: `attachBridge` calls `startForegroundService` when the restored queue has work, and when the app is not allowed to start a foreground service at that moment (e.g. launched while the screen is locked) the resulting `ForegroundServiceStartNotAllowedException` propagates out of `load()`, so Capacitor drops the entire download plugin for that app session. That would leave the download button doing nothing until the app is restarted, which sounds a lot like what is reported in #1970 — I have not confirmed they are the same, but the PR guards that call.

FelixMau:
## Brief summary

Cold start on Android took about a minute with many downloaded files, because two Capacitor plugins do heavy storage I/O in `load()`, which runs synchronously on the main thread. This moves that work off the main thread and makes it cheap: **75.4 s blocked → 0.3–0.8 s**, and the library cleanup itself from ~69 s → 1.6 s.

## Which issue is fixed?

Fixes #1997

## Pull Request Type

## In-depth Description

Four commits, each doing one thing.

**1. `Fix slow app startup with many downloaded files`** — `AbsDatabase.load()` ran `cleanLocalMediaProgress`, `cleanLocalLibraryItems` and `cleanLogs` synchronously. Capacitor calls `load()` from `BridgeActivity.onCreate`, so this blocked the first frame. Wrapped in `GlobalScope.launch(Dispatchers.IO)`, matching how the other methods of the same plugin already work.

**2. `Restore the download queue off the main thread`** — `AbsDownloader.load()` called `DownloadServiceHost.ensure()`, which deserializes the download db and probes shared storage. `DownloadServiceHost` now creates the manager synchronously, so `ensure()` still returns a usable instance to every caller, but runs `restoreQueue()` in a coroutine and defers the bridge attach until that job completes, which preserves the existing event and `bridgeReady` ordering. The persisted items were also read twice, once in `DownloadItemManager`'s constructor via `IncompleteDownloadCleanup.cleanupExpired()` and once in `restoreQueue()`; the loaded list is now passed to the cleanup so the db is read once.

This commit also guards `startForegroundService`. With the queue restored in the background it can be called while the app is in the background, where it throws `ForegroundServiceStartNotAllowedException`. That exception already escapes `load()` on v0.14.0-beta today — it is reproducible by launching the app while the screen is locked with work in the queue — and Capacitor then drops the whole plugin with `PluginLoadException: Unable to load plugin instance`, leaving downloads dead until the app is restarted. That may be what #1970 is seeing, though I have not confirmed it.

**3. `Establish folder access once instead of once per file`** — the expensive part of `cleanLocalLibraryItems`. Checking each file with `openFileDescriptor` costs a round-trip per file; listing the parent folder instead only helps by about 3x, because the provider stats every entry it returns.

Read access, however, is granted per document tree, not per document, so `contentResolver.persistedUriPermissions` answers it for a whole item at once. While that grant holds, a file that is still on disk is intact, and checking that costs a `stat`.

The full per-document check is kept and is used whenever the grant is not confirmed, so nothing is dropped from the local library on the basis of the cheap check alone. In that case every file is verified through the provider as before, except that the item's folder is listed once via `buildChildDocumentsUriUsingTree`, falling back to the per-file check when the folder cannot be listed. The three checks are now separate, named methods on `LocalFile` (`exists(ctx)`, `exists(ctx) { siblingIds }`, `existsOnDisk()`) with the constraint documented on the cheap one.

Worth flagging for review: a plain `File.exists()` is *not* a substitute for the provider check. Measured on Android 15, the app can `stat` files it holds no grant for, and cannot open even its own downloaded files through the `File` api (`exists=true`, `canRead=false`). Presence and accessibility are separate questions, which is why access is established separately rather than inferred from a path.

**4. `Match audio tracks to local files with a set lookup`** — not a regression, but it dominates the cleanup once the storage access is fixed. Every track scanned the whole local file list, twice. Collecting the ids into a set once per item is ~150x faster for a book with 370 files.

## How have you tested this?

On a Fairphone 4 / Android 15 with a library of 1557 downloaded files in a SAF folder on internal storage, using a debug build side by side with unmodified v0.14.0-beta. Cold start measured from logcat: `Starting BridgeActivity` → `load: AbsLogger plugin initialized`, and the plugin attribution comes from the `Registering plugin instance:` lines in between. Absolute values are higher than a release build would give; the ratios are what matters.

| Phase | v0.14.0-beta | this PR |
| --- | --- | --- |
| `AbsDownloader.load()` | 16.6 s | 0.06 s |
| `AbsDatabase.load()` | 58.1 s | 0.02 s |
| **main thread blocked** | **75.4 s** | **0.3–0.8 s** |
| until `mounted` | 77.7 s | 2.3–4.7 s |

The cleanup itself, instrumented, now on a background thread:

| | existence checks | track matching | whole cleanup |
| --- | --- | --- | --- |
| v0.14.0-beta | 53.8 s | 15.0 s | ~69 s |
| one folder listing per item | 18.2 s | 15.0 s | 33.5 s |
| + access established per folder | 1.3 s | 15.0 s | 16.3 s |
| + set lookup for tracks | 1.3 s | 0.11 s | **1.6 s** |

The 53.8 s vs 18.2 s comparison is an A/B on the same device and the same files, with only the checking strategy swapped.

Behaviour checks on the same device: all queued download items are still restored and forwarded to the frontend (visible as `DownloadProgressIndicator onDownloadItem` in the web console); no local library item lost files or tracks during the cleanup across every run; `persistedUriPermissions` correctly resolved to the granted tree for all items, including books in nested subfolders, so the cheap path is the one actually taken; and `AbsDownloader` now loads without the `PluginLoadException` described above.

Both intermediate commits build on their own.

## Screenshots

No frontend changes.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/data/DeviceClasses.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/data/DeviceClasses.kt
@@ -73,6 +73,13 @@ data class LocalFile(
         var mimeType: String?,
         var size: Long
 ) {
+  /**
+   * Full check: does the file exist and can this app actually open it?
+   *
+   * Costs a ContentResolver round-trip per file, so callers holding many files should prefer
+   * [exists] with a folder listing, or [existsOnDisk] when access to the folder is already known
+   * to be intact.
+   */
   @JsonIgnore
   fun exists(ctx: Context): Boolean {
     if (contentUrl.startsWith("content:")) {
@@ -87,26 +94,34 @@ data class LocalFile(
   }
 
   /**
-   * Existence check backed by a pre-fetched set of sibling document ids.
-   *
-   * Opening a file descriptor per file (see [exists]) costs a ContentResolver IPC round-trip
-   * (~15ms each), which adds up to minutes on libraries with thousands of downloaded files.
-   * When the caller already listed the parent folder once, this only does a set lookup.
-   * Falls back to the per-file check when no sibling set is available.
+   * Same check as [exists], but backed by the document ids of the parent folder, so a whole
+   * folder costs one query instead of a round-trip per file. Falls back to the per-file check
+   * when the folder cannot be listed (permission lost, non-tree uri).
    */
   @JsonIgnore
-  fun exists(ctx: Context, siblingDocumentIds: Set<String>?): Boolean {
-    if (siblingDocumentIds != null && contentUrl.startsWith("content:")) {
-      val documentId =
-              try {
-                DocumentsContract.getDocumentId(Uri.parse(contentUrl))
-              } catch (e: Exception) {
-                null
-              }
-      if (documentId != null) return siblingDocumentIds.contains(documentId)
+  fun exists(ctx: Context, siblingDocumentIds: () -> Set<String>?): Boolean {
+    if (contentUrl.startsWith("content:")) {
+      val documentIds = siblingDocumentIds()
+      if (documentIds != null) {
+        val documentId =
+                try {
+                  DocumentsContract.getDocumentId(Uri.parse(contentUrl))
+                } catch (e: Exception) {
+                  null
+                }
+        if (documentId != null) return documentIds.contains(documentId)
+      }
     }
     return exists(ctx)
   }
+
+  /**
+   * Is the file still on disk? Costs a stat, so it is orders of magnitude cheaper than the SAF
+   * checks above, but it only answers presence: under scoped storage a path can be visible while
+   * the file is not openable. Only use it when access to the containing folder has been
+   * established separately.
+   */
+  @JsonIgnore fun existsOnDisk(): Boolean = absolutePath.isNotEmpty() && File(absolutePath).exists()
 
   @JsonIgnore
   fun isAudioFile(): Boolean {
--- a/android/app/src/main/java/com/audiobookshelf/app/data/DeviceClasses.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/data/DeviceClasses.kt
@@ -2,6 +2,7 @@ package com.audiobookshelf.app.data
 
 import android.content.Context
 import android.net.Uri
+import android.provider.DocumentsContract
 import android.support.v4.media.MediaDescriptionCompat
 import android.util.Log
 import com.fasterxml.jackson.annotation.JsonIgnore
@@ -83,6 +84,28 @@ data class LocalFile(
       }
     }
     return File(absolutePath).exists()
+  }
+
+  /**
+   * Existence check backed by a pre-fetched set of sibling document ids.
+   *
+   * Opening a file descriptor per file (see [exists]) costs a ContentResolver IPC round-trip
+   * (~15ms each), which adds up to minutes on libraries with thousands of downloaded files.
+   * When the caller already listed the parent folder once, this only does a set lookup.
+   * Falls back to the per-file check when no sibling set is available.
+   */
+  @JsonIgnore
+  fun exists(ctx: Context, siblingDocumentIds: Set<String>?): Boolean {
+    if (siblingDocumentIds != null && contentUrl.startsWith("content:")) {
+      val documentId =
+              try {
+                DocumentsContract.getDocumentId(Uri.parse(contentUrl))
+              } catch (e: Exception) {
+                null
+              }
+      if (documentId != null) return siblingDocumentIds.contains(documentId)
+    }
+    return exists(ctx)
   }
 
   @JsonIgnore
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/DbManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/DbManager.kt
@@ -1,6 +1,8 @@
 package com.audiobookshelf.app.managers
 
 import android.content.Context
+import android.net.Uri
+import android.provider.DocumentsContract
 import android.util.Log
 import com.audiobookshelf.app.data.*
 import com.audiobookshelf.app.models.DownloadItem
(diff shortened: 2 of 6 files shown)


FelixMau:
**Disclosure: this PR was written with AI assistance (Claude Code).**

I just had the issue and knew i could fix it for my own usage with the AI and thought you might want to use what it found for me as well. I dont know Kotlin and cant review or rate the work of it myself. Please just close it if annoying. I will convert my files to m4a in future and wont have the issue anyway then.

I only found your note in #1831 — that AI generated PRs and issues are not being reviewed right now — after opening this and #1997. Sorry for that ordering. Please close both without further explanation if they are unwanted; no reply needed and none expected. If you would rather have the diagnosis without the patch, the measurements in #1997 stand on their own and the branch stays available in my fork.

To make triage cheap, here is what was actually exercised on device (Fairphone 4, Android 15, debug build, 1557 downloaded files, alongside unmodified v0.14.0-beta):

- startup timings and per-plugin attribution from logcat, and the A/B of the three checking strategies on identical files
- restored download items still reach the frontend (`DownloadProgressIndicator onDownloadItem`), and no local library item lost files or tracks in any run
- `persistedUriPermissions` resolved to the granted tree for every item, including books in nested subfolders, so the cheap path is the one actually taken
- both intermediate commits build on their own

What is **not** exercised, and where I would look first if you do review it:

- the branch where the tree grant is *missing*. It always resolved to granted on my device, so the selection of the full per-document path was never taken in a live run. The two checks it delegates to were each measured separately, but that branch itself is unproven.
- the podcast half of the track-matching change — my library is books only, so only the book branch ran.
- release builds and any other device. Absolute numbers here are debug-build numbers; the ratios are the claim, not the milliseconds.
- iOS is untouched.

I am the reporter, not a Kotlin reviewer — I noticed the app took a minute to open and provided the device and library. Please weigh it accordingly.

Issue #1510, opened 2025-03-10
[Bug]: When device connects to server network progress overwrites


totopesce:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

When listening to audiobooks localy on my phone and I connect to my local netwotk, where the server is, the progress on the book gets overwritten, it seems to be with whats on the server. 

### Steps to Reproduce the Issue

1. Listen to any local book disconnected from the server
2. Pause and close the app
3. Connect to the server network and open the app

### What was Expected?

Progress from my phone should update the servers progress. 

### Phone Model

Poco x6 pro

### Phone OS

Android 14

### Audiobookshelf App Version

Android App - 0.9.79

### Installation Source

Google Play Store

nichwall:
Related to
https://github.com/advplyr/audiobookshelf-app/issues/1456
https://github.com/advplyr/audiobookshelf-app/issues/1355
https://github.com/advplyr/audiobookshelf-app/issues/1120
https://github.com/advplyr/audiobookshelf-app/issues/739

advplyr:
I'm not able to reproduce this but in v0.9.80 a logs page was added so if anyone is able to reproduce this we should be able to find out exactly what is happening.

Here are the steps I took to try to reproduce this issue (Android 14, Google Pixel 6):
1. Downloaded a book in the app
2. Put my phone on airplane mode
3. Started playing the audiobook and jumped to around 66% then let it play at least 30 seconds (important).
4. Opened the web client on my desktop to confirm that the progress is still showing 33% there.
5. Closed out of the app completely
6. Took my phone out of airplane mode
7. Opened the app & watched it successfully connect
8. Observed the progress is correctly updated on the server to 66%

If anyone is able to reproduce the issue following those steps that would be most helpful because the logs will show every step.

The mobile app logs show that the local playback session gets sent to the server.
![Image](https://github.com/user-attachments/assets/061d1ee0-6a9e-4e84-a023-76e9313b028c)

In the server logs I can see the local playback session come in and the progress getting updated.
```bash
[2025-05-22 16:40:17.200] INFO: [PlaybackSessionManager] Syncing local session "The Adventures of Tom Sawyer" (033a0381-a71e-466c-ab7b-09b0949f2f5a)
[...]

Issue #1442, opened 2025-01-19
[Bug]: After closing and opening the App it timeskips to a earlier mark


WickedSquirrel:
### What was the Problem?

I'm using the newest version of the App with the newest Version of the Server and the newest version of Android on a Pixel 8.

The Problem I often, but not always, encounter is that after closing the app and opening it at a later time the Audiobook continues from an earlier time mark even though it remembers where I left off.

![Image](https://github.com/user-attachments/assets/6035b451-60df-43e7-82bf-ea864ee7f6e0)

I can then click on the lower time mark that's farther into the audiobook and it will play from there.
But if I don't immediately recognize that it's an earlier time mark and then skip back a bit because I don't recognise where I am the list of history marks gets longer and it gets harder to find the actual point where I stopped and thus want to continue.

### Steps to Reproduce the Issue

Stop an Audiobook.
Close the App.
Open the App.
Start the Audiobook.. -> wrong time mark (sometimes)

### What was Expected?

The right Time mark.

### Phone Model

Pixel 8

### Phone OS

Android 15

### Audiobookshelf App Version

Android App - 0.9.78

### Installation Source

Google Play Store

### Additional Notes

The Problem is not new.. I think I have had problems like this for a few months.

It does not only occur when I leave my home network as I previously thought but also when I just stay in it and the App all the time had the chance to connect to the server.

nichwall:
Related to https://github.com/advplyr/audiobookshelf-app/issues/1416, https://github.com/advplyr/audiobookshelf-app/issues/1298, https://github.com/advplyr/audiobookshelf-app/issues/1355,

PR #2004, opened 2026-08-28
Push isFinished/ebookLocation to the server on reconnect when offline (#1397)

### LINKED ISSUE #1397: [Bug]: finishing audiobook in offline-modus will not be sync to mediaserver when device is online next time
### What was the Problem?

when I finish a audiobook offline :
image

and connect to the Mediaserver next time 
neither the history-Entries nor the "finished at" will be sync from device to mediaserver:
image

1. download audiobook
2. disconnect from Mediaserver
3. liste and finish audiobook offline
4. connect to MedioServer
5. goto Audiobook n WebGui
There is no "finished at" in WebGUI

### What was Expected?

when finish a audiobook offline add it to some kind of queue and next time when device is online (connected to mediaserver)
processing the queue -> History and status "finished" will be sync to medaiserver

### Phone Model

Huawei P30

### Phone OS

Android 12

Android App - 0.9.77

### Installation Source

Google Play Store

### Additional Notes

_No response_

ljlongwing:
## Brief summary

`syncLocalMediaProgressForUser` reconciles local and server media progress on every server connection, but its "local is newer than server" branch only ever patched ebook fields. `isFinished` toggled while offline (including "mark finished" on a downloaded book) is never sent up if the immediate push in `updateLocalMediaProgressFinished` failed because the device was offline - so the server stays behind and can overwrite the local state on a later sync.

This broadens that branch to also PATCH `isFinished` when local `lastUpdate` is newer and it actually differs from the server - mirroring the existing ebookLocation handling right next to it.

## Which issue is fixed?

Fixes #1397. Should also help the "server overwrites my newer progress" reports where the item was marked finished offline: #1338, #1442, #1510, #1852, #1945 (to the extent those are about `isFinished`, not currentTime drift).

## Pull Request Type

Android, native (Kotlin). Touches `ApiHandler.syncLocalMediaProgressForUser` only.

## In-depth Description

Before, the reconciliation loop over `user.mediaProgress` was:

- `server.lastUpdate > local.lastUpdate` → overwrite local from server (all fields)
- `local.lastUpdate > server.lastUpdate` **and `local.ebookLocation != null`** → patch `ebookLocation`/`ebookProgress` to the server
- otherwise → count as up-to-date

**Note this PR does *not* touch `currentTime`/`progress`.** Those are already synced to the server through a different path: `AbsDatabase.syncLocalSessionsWithServer` calls this function first, then calls `sendSyncLocalSessions`, which replays saved local `PlaybackSession`s to `/api/session/local-all` and is what actually advances server `currentTime`/`progress`. An earlier version of this PR also pushed `currentTime`/`progress` directly from `syncLocalMediaProgressForUser`, which would have double-synced that data through both paths - thanks @advplyr for flagging that.

`isFinished` (like `ebookLocation`) has no `PlaybackSession` behind it - toggling finished doesn't record one. `updateLocalMediaProgressFinished` already tries to PATCH `isFinished` to the server immediately when the user toggles it, but if the device is offline at that moment the push is just lost, and nothing retries it later. This change makes `syncLocalMediaProgressForUser`'s "local is newer" branch catch that up on the next connect, sending only `isFinished` (plus `lastUpdate`) - the same shape `updateLocalMediaProgressFinished` already sends.

Also fixed a copy/paste bug in the `ebookProgress` change-log line (it logged `isFinished` values), and made the PATCH endpoint episode-aware (`/api/me/progress/:libraryItemId/:episodeId`) so this works for podcast episodes, not just books.

## How have you tested this?

- `./android/gradlew -p android assembleDebug` builds cleanly.
- Not yet exercised on a device against a real offline scenario - see note below.

## Notes / open questions for a maintainer

- **This runs on every server connection and writes to the server**, so I'd still like real device testing before merge (mark finished while offline → reconnect) and a second look at the scope now that it's narrowed to just `isFinished`.
- The webview-liveness gap @advplyr flagged (this whole sync only runs while `default.vue` is mounted, so it doesn't help if the app is never reopened between connections) is real but out of scope here - happy to discuss moving the trigger into native separately.
- **Known remaining gap, not addressed here:** local progress for an item the server has *no* record of at all (downloaded, then played entirely offline) is still skipped, because the loop only iterates `user.mediaProgress`. iOS has an analogous path that this PR does not touch.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/server/ApiHandler.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/server/ApiHandler.kt
@@ -849,26 +849,24 @@ class ApiHandler(var ctx:Context) {
               DeviceManager.dbManager.saveLocalMediaProgress(localMediaProgress)
               numLocalMediaProgressUpdated++
             } else if (localMediaProgress.lastUpdate > mediaProgress.lastUpdate) {
-              // Local progress is more recent than the server's - e.g. the user listened, or marked
-              // the item finished, while offline. Push the local values up so the server (the source
-              // of truth that other devices read, and that would otherwise overwrite this local
-              // progress on the next sync) is not left behind.
+              // Local progress is more recent than the server's. currentTime/progress from actual
+              // playback is already covered: AbsDatabase.syncLocalSessionsWithServer calls this
+              // function first, then replays saved local PlaybackSessions to the server via
+              // sendSyncLocalSessions, which is what advances server currentTime/progress. Pushing
+              // those fields here too would sync the same activity twice through two paths.
               //
-              // Local lastUpdate is only advanced to "now" when something actually happened locally
-              // (playback save, mark finished, ebook progress); updateFromServerMediaProgress copies
-              // the server timestamp. So a newer local timestamp reliably means real local activity,
-              // and the position values are correct even if the device clock is skewed.
+              // isFinished (and ebookLocation) have no PlaybackSession behind them - toggling
+              // finished, or moving an ebook location, doesn't record a session. updateLocalMediaProgressFinished
+              // / updateLocalEbookProgress already try to push those to the server immediately when
+              // they happen, but that push is lost if the device was offline at the time, and nothing
+              // else retries it. This is the gap this branch closes on reconnect, mirroring exactly
+              // what those immediate pushes send.
               val pushLogs = mutableListOf<String>()
-              if (mediaProgress.progress != localMediaProgress.progress) {
-                pushLogs.add("progress ${mediaProgress.progress} -> ${localMediaProgress.progress}")
-              }
-              if (mediaProgress.currentTime != localMediaProgress.currentTime) {
-                pushLogs.add("currentTime ${mediaProgress.currentTime} -> ${localMediaProgress.currentTime}")
-              }
               if (mediaProgress.isFinished != localMediaProgress.isFinished) {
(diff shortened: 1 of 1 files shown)


advplyr:
The issue with this approach is local progress is synced to the server from `default.vue`. This would double sync local progress.

Users may be having sync issues with local progress if they never open the app between connections. Since the sync happens in webview, if webview is not alive it won't sync. If we close this gap it requires moving that logic out webview and into native kotlin/swift

ljlongwing:
Good catch, thanks. Pushed a revision that narrows this to just `isFinished` and `ebookLocation` - dropped `currentTime`/`progress`/`duration` from the local->server push entirely, since you're right that those are already covered by the session-based sync (`sendSyncLocalSessions` replaying saved `PlaybackSession`s right after this function runs).

`isFinished`/`ebookLocation` are different: they aren't backed by a `PlaybackSession` at all (toggling finished or moving an ebook location doesn't record one), and `updateLocalMediaProgressFinished`/`updateLocalEbookProgress` already try to push them to the server immediately when they happen - but that immediate push is silently lost if the device is offline at the time, with nothing to retry it. That's the actual gap for #1397 (mark-finished-offline never syncing), and this PR is now scoped to just closing it on reconnect, mirroring exactly what those immediate-push call sites already send.

Re: the webview-not-alive gap you flagged - agreed that's real and separate (this only ever runs from `default.vue`'s `mounted`/reconnect hooks), and out of scope for this PR. Happy to look at moving the trigger into native if that's wanted, but figured it's worth discussing separately/with maintainer input on approach given the bigger surface area.

Issue #1963, opened 2026-08-17
[Bug]: cover art fails to download after update to 0.14.0-beta


GrumbleBumpkin:
### I have verified that the [bug is not already awaiting release](https://github.com/advplyr/audiobookshelf-app/issues?q=is%3Aissue%20label%3A%22awaiting%20release%22)

Yes

### What was the Problem?

I recently updated to 0.14.0-beta and now many books - including books I have previously managed to download to local in 0.13.0-beta - no longer complete the download and seem to be stuck attempting to download the cover art, but instead of a green check by the download item in the download progress list like the audio files, the cover is stuck at 100%.

This ultimately causes the local item to never get created. This means that the local item can't be played, only streamed (which does still work). Similarly, the item doesn't appear when accessing the Local Media management menu, so the item can't be deleted. The UI for the book in question can be in one of two states:

- **It thinks the item is not downloaded:** In this case, there is a button to stream the item, and the button to download the book to local is visible and clickable, but clicking it leads to a warning at the bottom of the screen saying "Download already started for this media entity". This is the state entered when first downloading a book that then fails.
- **It thinks the item is downloading:** This is the case after closing and reopening the app. There is a number of downloads pending in the upper right of the screen as usual, but no progress is made and the number shown is the number of cover art files stuck at 100%. The stream option is still available for the book, but this time the Download button is orange as though there is a download in progress. There's a "Downloading... (0%)" dialog under the Stream button, but it never progresses. This state also survives a reboot of the phone.

I've looked in the Logs menu, but there are no error logs, but there is a traceback in the server logs when attempting to download the item. I'm not sure how it's related to this issue, but I can verify that it doesn't occur when I download one of my books that doesn't have this download issue:

RangeNotSatisfiableError: Range Not Satisfiable
at createHttpError (/app/node_modules/send/index.js:984:7)
at SendStream.error (/app/node_modules/send/index.js:270:31)
at SendStream.send (/app/node_modules/send/index.js:666:19)
at onstat (/app/node_modules/send/index.js:725:10)
at FSReqCallback.oncomplete (node:fs:198:5)

I last updated my server installation on 8/3 and have downloaded books since then, so I don't think this is a server side issue despite the errors appearing there. I updated my client to 0.14.0-beta yesterday and only now have encountered this issue.

### Steps to Reproduce the Issue

1. Go to an item in the ABS library.
2. Click the download button.
3. Wait for download to complete. Local item is not created and cannot be played locally.

[...]

PR #2005, opened 2026-08-28
Refresh the auth token on a 401 in the download path (follow-up to #1953)

ljlongwing:
## Brief summary

Follow-up to #1953, as discussed there. That PR recovers 416 / interrupted transfers; this one handles a **401** on the file download request, which wedges a download into the same "stuck at 100% / Download already started" state by a different route.

## Which issue is fixed?

Follow-up to #1953. Also relevant to the non-Range-bug reports in #1970 / #1963 where "kill the app and reopen" is the only recovery.

## Pull Request Type

Android, native (Kotlin).

## In-depth Description

`InternalDownloadManager.download()` sends `Authorization: Bearer <token>` using the access token read once in `DownloadItemManager.startDownload()`. Every other network path in the app refreshes on 401 (`ApiHandler.makeRequest` → `handleTokenRefresh`, and the JS `nativeHttp`), but the file downloader has no 401 handling — a 401 falls through to the generic failure path. When the access token expires while items sit in the queue, or mid-transfer on a large book, each part fails, burns its transfer retries against the dead token, and latches `terminalFailureAt`. It only clears once something else in the app refreshes the token.

Changes:

- **`ApiHandler`**: extract the `/auth/refresh` core out of the private `handleTokenRefresh` into a reusable `refreshAuthTokens(httpClient?, onResult: (String?) -> Unit)`. `handleTokenRefresh` now delegates to it; behavior for the existing API 401 path (including `handleRefreshFailure` session-clear semantics) is unchanged.
- **`InternalDownloadManager`**: detect HTTP 401 and report it distinctly via a new `InternalProgressCallback.onAuthError()` instead of a generic failure.
- **`DownloadItemManager`**: on `onAuthError`, park the part (`downloadId = null`, not failed — **no transfer retry consumed**), bump a separate `authRetryCount` (bound `MAX_AUTH_RETRIES = 2`), and fire a single de-duplicated `refreshAuthTokens()`. On success, `checkUpdateDownloadQueue()` restarts the parked part(s) with the freshly stored token. On failure, the parked parts go to terminal failure (and the session is cleared by the refresh path, as elsewhere). Also refactored the terminal-failure block out of `failOrRetry` into a shared `markTerminalFailure`, and added a `terminalFailureAt != null` guard in `handlePartUpdate` so a stale watcher iteration can't un-latch a terminal auth failure.
- **`DownloadItemPart`**: add `@JsonIgnore var authRetryCount: Int = 0`.

Layered so it sits alongside #1953's 416 branch (my `onAuthError` path vs their `416` path; my `handleAuthError` vs their `retryDownloadItem`). If #1953 lands first this will need a small rebase in the same three files, and vice versa — happy to do whichever.

## How have you tested this?

- `./android/gradlew -p android assembleDebug` builds cleanly.
- Not yet exercised on a device against a live token expiry — the logic is a mirror of the app's existing `handleTokenRefresh` retry, scoped to the download path.

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/managers/DownloadItemManager.kt
@@ -9,12 +9,14 @@ import com.audiobookshelf.app.device.DeviceManager
 import com.audiobookshelf.app.device.FolderScanner
 import com.audiobookshelf.app.models.DownloadItem
 import com.audiobookshelf.app.models.DownloadItemPart
+import com.audiobookshelf.app.server.ApiHandler
 import com.fasterxml.jackson.core.json.JsonReadFeature
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.getcapacitor.JSObject
 import java.io.File
 import java.io.FileInputStream
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicBoolean
 import kotlin.math.max
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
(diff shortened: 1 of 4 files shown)


ljlongwing:
Closing as superseded - the same 401-during-download bug got fixed independently and merged into master this week:

- 4b8abe7d "Refresh auth token on download 401"
- e06ce4a8 "Only clear refresh token and logout when refresh request explicitly rejects..."
- 71bd354b "Refactor token refresh to better handle types of failures and retries during download"

That version is more robust than this branch (per-server-connection-config token refresh, and a distinction between a rejected refresh token vs. a transient network failure so a blip doesn't log the user out), so nothing left to add here. Thanks for the fix!

PR #2010, opened 2026-09-04
Offer the download again when the local copy is behind the server

davidj4tech:
## Problem

A downloaded book never learns that its server item gained tracks.

`showDownload()` returns `false` whenever `hasLocal` is true, so once a local copy exists the download button is gone and there is no other action that fetches the missing files. Meanwhile the item page keeps listing the **server** item's tracks with a LOCAL badge, and playback prefers the local copy — so the new track is visible in the UI and silently absent from the audio. The only way out is to delete the local copy and download the whole item again, which is not discoverable and re-fetches everything.

Two ways to hit this:

- **Tracks added to an existing item.** Anything appended to a book folder after it was downloaded (a missing part that arrives later, a re-encode split into more files, or — my case — a library item that is written to incrementally).
- **A download that only partly completed.** The local copy holds what it got, plays it, and offers no way to fetch the rest.

Related: #525 is the same shape for chapters (server updates, local copy keeps the old ones), and #1362 asks for the local folder to be reconciled with the server.

## What this PR does

Client-side only, `pages/item/_id/index.vue`, no changes to the downloader or the scanner:

- `localCopyIsOutdated` — true when the local copy has fewer audio tracks than the server item (books only; podcasts already download per episode and are excluded).
- `showDownload()` offers the button again in exactly that case, with a `sync` icon instead of `download` so it does not read as a duplicate download.
- The confirm dialog says what will happen: *"X" has N more audio tracks on the server than the copy in <folder>. Update the download?*
- The update goes back to the folder the existing copy is in (`localLibraryItem.folderId`) rather than prompting again, so it cannot land a second copy in a different folder.

The download path itself is unchanged and already does the right thing: `FolderScanner.scanDownloadItem` looks up `local_<libraryItemId>` and reuses it, so this updates the existing local item rather than creating a duplicate, and `setAudioTracks` renumbers and re-offsets the merged result.

## What it deliberately does not do

It re-downloads every track, not just the missing ones. Skipping files already on disk is exactly what #1898 implements in `DownloadItemManager`, and the two compose: with that PR merged, this button becomes an incremental update for free, since parts already on disk complete immediately and the scan still sees the full set. I kept this one to the UI so the two do not overlap.

## Testing

**Builds clean** — `build-apk.yml` on this branch: `npm ci`, `npm run generate`, `npx cap sync`, `gradlew assembleDebug`, APK uploaded, all green.

**Run on a device** (see the comment below): a two-track item downloaded to Internal App Storage, a third track added on the server, and the `sync` button appeared and completed the update. Done on **v0.13.0-beta**, because a master build cannot download anything at all (#1970) and so never reaches this code.

What I did verify beyond the build:

- The server-side behaviour it responds to, measured against Audiobookshelf 2.35.1: appending a file to a book folder keeps the same item id, appends the track at the right offset, leaves existing files' `ino`, index and mtime untouched, and preserves the listener's position (stored in seconds). So "the local copy is behind" is a real state a server can be in, not a hypothetical.
- The code paths by reading: `showDownload`, `download()`'s folder resolution, `AbsDownloader.downloadLibraryItem` → `startLibraryItemDownload`, and `FolderScanner.scanDownloadItem`/`scanParts`.
- The branch table, by running the four computed properties as plain functions over nine cases: not downloaded, downloaded and current, item grew, partial download, local copy somehow ahead, ebook-only with no tracks, podcast, viewing the local item page, and a user without download permission.

If you would rather this arrived as an issue with the analysis and no patch, say so and I will close it — the finding is worth more than the diff.

### DIFF
--- a/pages/item/_id/index.vue
+++ b/pages/item/_id/index.vue
@@ -60,7 +60,7 @@
               <span v-if="!showPlay" class="px-2 text-base">{{ $strings.ButtonRead }} {{ ebookFormat }}</span>
             </ui-btn>
             <ui-btn v-if="showDownload" :color="downloadItem ? 'warning' : 'primary'" class="flex items-center justify-center mx-1" :padding-x="2" @click="downloadClick">
-              <span class="material-symbols text-2xl" :class="downloadItem || startingDownload ? 'animate-pulse' : ''">{{ downloadItem || startingDownload ? 'downloading' : 'download' }}</span>
+              <span class="material-symbols text-2xl" :class="downloadItem || startingDownload ? 'animate-pulse' : ''">{{ downloadItem || startingDownload ? 'downloading' : localCopyIsOutdated ? 'sync' : 'download' }}</span>
             </ui-btn>
             <ui-btn color="primary" class="flex items-center justify-center mx-1" :padding-x="2" @click="moreButtonPress">
               <span class="material-symbols text-2xl">more_vert</span>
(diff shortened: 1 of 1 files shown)


davidj4tech:
Tested on a device now, and it does what it says.

**Build:** this branch's change applied to **v0.13.0-beta**, not master. On a master build (0.14.0-beta) the item never downloads in the first place — #1970 — so there is no local copy to be behind, and the patch cannot be reached. Everything below is therefore the same diff on the last release where downloading works.

**Setup:** a book folder with two 20-second tracks, downloaded to Internal App Storage on Android (Pixel 8a, Audiobookshelf server 2.35.1).

**Then:** dropped a third file into the folder, scanned. The item went to 3 tracks / 60s, same item id, first two files keeping their inodes and offsets (0, 20, new one at 40).

**Result:** back on the item page, the button reappeared with the `sync` icon, the confirm named the one missing track, and the update completed — the local copy now has all three.

So the two states this addresses are both real, and the fix reaches them:

- an item that gains tracks after it was downloaded (measured above),
- a download that only partly completed, which lands in the same place with no way out.

Still worth pairing with #1898 so the update fetches only what is missing rather than everything; this PR deliberately leaves the downloader alone.

PR #2012, opened 2026-09-09
Fix username whitespace on login

### LINKED ISSUE #1900: [Bug]: Android app login fails when username contains trailing whitespace

ShamiaAkterShanta:
## Brief summary

Trims leading and trailing whitespace from the username before sending the login request. This prevents login failures when a username is entered with accidental spaces.

## Which issue is fixed?

Fixes #1900

## Pull Request
Android frontend bug fix.

## In-depth Description

The login request previously used the username exactly as entered by the user. If the username contained a leading or trailing space, authentication could fail.

This change applies `trim()` to the username before sending the login request. The password is not modified.

## How have you tested this?

Verified that the login request now uses the trimmed username and confirmed that no other files were changed.

## Screenshots

Not applicable. This change does not modify the user interface.

### DIFF
--- a/components/connection/ServerConnectForm.vue
+++ b/components/connection/ServerConnectForm.vue
@@ -639,7 +639,7 @@ export default {
         'x-return-tokens': 'true',
         ...(this.serverConfig.customHeaders || {})
       }
-      return this.postRequest(`${this.serverConfig.address}/login`, { username: this.serverConfig.username, password: this.password || '' }, headers, 20000)
+      return this.postRequest(`${this.serverConfig.address}/login`, { username: this.serverConfig.username.trim(), password: this.password || '' }, headers, 20000)
         .then((data) => {
           if (!data.user) {
             console.error(data.error)


ShamiaAkterShanta:
I also checked the backend behavior as suggested.

I tested this on a local Audiobookshelf v2.36.0 server. I created a user with trailing whitespace in the username and was still able to log in using the username without the trailing space.

I also updated the same user's username with trailing whitespace and confirmed that login without the trailing space still works.

I also tested the web login with trailing whitespace, and it logged in successfully.

This confirms that the backend/web flow handles the trailing whitespace, while this PR fixes the issue in the Android app login flow.

PR #2016, opened 2026-09-10
[Feat] mTLS client cert authentication

### LINKED ISSUE #353: Support SSL certificates

I wasn't sure whether to open a separate request or tag on to #254 , but I'm using SSL certificates to secure access to self-hosted apps. Works very well for the web version using apache2 to proxy, but presumably the app (Android 12) doesn't know to prompt for a certificate to present, and produces "Failed to ping server" as an error.

App to connect using a TLS certificate.
I'm not familiar with Capacitor development but this looks promising: https://github.com/cordova-ccafix/cordova-plugin-client-certificate-support

Avoiding app and using webapp only, reducing security by removing requirement for authentication.

### LINKED ISSUE #1419: [Enhancement]: Add mTLS support

Jnewbon:
## Brief summary

Implementation of Mutual TLS support on the Android platform (Re attempt from #1692)

## Which issue is fixed?

Resolves #1419 
Resolves #353 

## Pull Request Type

Affects Android, Both front and back end

## In-depth Description

My plans as it stands is,
- Build UI Client cert selector for each server
  - [x] Select a cert per server
  - [x] Edit a cert for existing servers
- Update Kotlin API to make use of the selected cert
  - [x] Separate the client builder and encapsulate mTLS usage
  - [x] Redirect API calls into this new builder
  - [x] Ensure frontend is routed through mTLS handler
  - [x] Tested that this actually works
- Update the VUE components to make use of the cert
  - [x] Ensure frontend is routed through mTLS handler

### 1. UI Selection
There seems to be 2 ways a website handles thing my use case Cloudflare mTLS handshake succeds weather or not a client certificate is provided, you only find out you failed via a 403 error, where as if the mTLS check is handled at the during the handshake then the app specifically knows it needs a client cert, so in both instances, i have to reporting a client cert is required during new server process

Sadly documentation for mTLS heavily favors hard-coded cert and keys where the app developer provide their own server and only want verified apps to be able to communicate with their server. i have been unable to find a way of getting the process to be automated thus have fallen to the backup of getting the user to manually select a cert when setting up the server connection.
<details>
<summary>ScreenShot</summary>
image

image

image

image

</details>

### 2. Kotlin API
Updating the Kotlin code to use mTLS based on a pre-selected certificate is well documented, so implementing this wasn't overly difficult, the main issue is ensuring that all calls in the Kotlin code based is routed via a client that has the mTLS certificate set. While this code is mostly done i haven't been able to test it properly due to the next section.

### 3. Frontend UI
Since my PR #1692 it seems that intercepting TLS handshake calls made by the front end app has become a lot simpler thus almost all functional code is kotlin based with the only changed front end wise is UI based for cert selection

Comment and suggestion are welcome, I'm not a native app developer, so I'm bundling through this a lot with trial and error.

## How have you tested this?

Yes Android, and Cloudflare mTLS, this also works if the server suddenly doesn't need the client cert, (eg locally hosted) in CF case its ignored, i haven't test with other privders

<details>
<summary>ScreenShot</summary>
image
image
image

</details>

Disclaimer: Claude was used to help with this work

### DIFF
--- a/android/app/src/main/java/com/audiobookshelf/app/MainActivity.kt
+++ b/android/app/src/main/java/com/audiobookshelf/app/MainActivity.kt
@@ -12,6 +12,7 @@ import android.os.IBinder
 import android.util.Log
 import android.view.ViewGroup
 import android.view.WindowInsets
+import android.webkit.ClientCertRequest
 import android.webkit.WebView
 import androidx.core.app.ActivityCompat
 import androidx.core.view.updateLayoutParams
@@ -20,11 +21,17 @@ import com.anggrayudi.storage.SimpleStorageHelper
 import com.audiobookshelf.app.managers.DbManager
 import com.audiobookshelf.app.player.PlayerNotificationService
 import com.audiobookshelf.app.plugins.AbsAudioPlayer
+import com.audiobookshelf.app.plugins.AbsCertificate
 import com.audiobookshelf.app.plugins.AbsDatabase
 import com.audiobookshelf.app.plugins.AbsDownloader
 import com.audiobookshelf.app.plugins.AbsFileSystem
 import com.audiobookshelf.app.plugins.AbsLogger
+import com.audiobookshelf.app.server.MtlsManager
 import com.getcapacitor.BridgeActivity
+import com.getcapacitor.BridgeWebViewClient
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 
 
 class MainActivity : BridgeActivity() {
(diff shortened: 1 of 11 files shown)


Jnewbon:
I'm installing this version to my mobile to do a longer in use test with full mTLS enabled before i mark this PR as ready

Jnewbon:
Been running this for a week or no issues with my configuration

PR #2033, opened 2026-09-19
Stop startup from reading every log record

huggenknubbel:
Opening the app took 45s on a device I listen on every night, against 5s on a
device I barely use. 21s of that passed before the first frame, with the app
logging nothing at all.

`AbsDatabase.load()` runs `cleanLogs()` on the main thread, and pruning has to
deserialise every log record to learn its age - the record name is a random id,
so the age is only inside the value. A progress sync writes three records every
15s, so a night of listening leaves several thousand of them, and 48h of
retention keeps several thousand more.

Measured on a Pixel 10 Pro XL, cold page cache, same build and same network:

| records in `log` | first frame | fully initialized |
|---|---|---|
| 5539 | 21.4s | 45.6s |
| 0 | 2.4s | 18.6s |
| 5568 | 12.9s | 34.8s |
| 0 | 1.9s | 18.6s |

Two changes:

- Records are now named `<timestamp>-<id>`, so pruning is a key scan with
  nothing deserialised.
- Closed hours are folded into one record per hour in a second book, named after
  that hour. The record count now follows the retention window rather than how
  much the app is used - 48 records instead of thousands. Folding runs off the
  main thread; records written by the old layout cost one read each, once.

Retention is unchanged at 48 hours, and `getAllLogs` still returns the same
entries - it reads the folded records plus the open hour.

Verified on device: new records carry the timestamp in the name, the leftover
old-layout records were folded into a single record rather than dropped, and
the first frame now arrives in 3.6s with the log book populated.


huggenknubbel:
Closing this - a single append-only file is a better fit for a log than key-value records, and makes the pruning free instead of merely cheap. I will open a fresh PR with that approach.

